PR #2, opened 2019-03-21
First batch of NF scripts

jchoude:
There you go!


scilpy/utils/bvec_bval_tools.py
```diff
@@ -0,0 +1,6 @@
+import numpy as np
```
GuillaumeTh: Please add # `-*- coding: utf-8 -*-`.

scripts/scil_extract_dwi_shell.py
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env python
```
GuillaumeTh: please add `# -*- coding: utf-8 -*-`.

PR #5, opened 2019-03-23
Add requirements and setup

jchoude:


GuillaumeTh:
Looks good to me. We can merge if nothing else.

PR #11, opened 2019-06-20
BF: Fix shell indices

GuillaumeTh:
Add a unique function on the shell indices otherwise the script crash on a case like bvals_to_extract="0 1500 1499".


mdesco:
I approve

arnaudbore:
Perfect, merging

PR #9, opened 2019-06-17
Futurize transform scripts

frheault:
First batch of scripts needed for the BST to be in the open Scilpy.
Scripts related to transformation of images/tractograms often used together (and in the BST pipeline)

Updated and tested for Python3
Simple data for testing
https://drive.google.com/file/d/1zMG3YOSpXEQw0ta04XGLAN4DyII1013Z/view?usp=sharing

PR #12, opened 2019-06-20
Futurize local tracking

frheault:
Transfert of the local_tracking script

The argparser of the script was heavily modify to be as similar as possible to the pft tracking already merged.
Modifications in both tracking scripts were done to increase consistency and ease the usage (variable name, order of parameters, comments)

The argparser and behavior of the pft tracking were not change.

Testing data
https://drive.google.com/file/d/18GeWgZbwbHUWSilsnK3_ZVfWQAzPKwgH/view?usp=sharing

PS : Sorry for the conversion of double quote to single quote for string. Probably a bad Find and Replace, but I was not able to easily revert without loosing changes. So I left it that way.


scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,269 @@
...
+                            if scaled_min_length <= length(s) <= scaled_max_length)
+    if args.compress:
+        filtered_streamlines = (
+            compress_streamlines(s, args.compress)
+            for s in filtered_streamlines)
+
+    tractogram = LazyTractogram(lambda: filtered_streamlines,
+                                affine_to_rasmm=seed_img.affine)
+
+    # TODO replace with create_header_from_image after PR #9
```
arnaudbore: #9 has been merged :+1:

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
+#! /usr/bin/env python
```
jchoude: While we are here, you could rename this to simply `scil_compute_local_tracking.py`. I don't think we should port the other local tracking script.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+# -*- coding: utf-8 -*-
+
+"""
+Local streamline HARDI tractography.
```
jchoude: switch `HARDI` and `streamline` please.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+
+DETERMINISTIC = 'det'
+PROBABILISTIC = 'prob'
+EUDX = 'eudx'
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
jchoude: Change class to `RawTextHelpFormatter`, else the formatting of the help is broken.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+
+    p._optionals.title = 'Generic options'
+    p.add_argument('sh_file',
+                   help='Spherical harmonic file. \n'
+                   '(isotropic resolution, nifti, see --basis).')
+    p.add_argument('seed_file',
+                   help='Seeding mask (isotropic resolution, nifti).')
+    p.add_argument('mask_file',
+                   help='Seeding mask(isotropic resolution, nifti).\n' +
```
jchoude: missing a space before `(`

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+    seed_sub_exclusive.add_argument(
+        '--npv', type=int,
+        help='Number of seeds per voxel.')
+    seed_sub_exclusive.add_argument(
+        '--nt', type=int,
+        help='Total number of seeds to use.')
+
+    p.add_argument(
+        '--sphere', choices=sorted(SPHERE_FILES.keys()),
+        default='symmetric724',
```
jchoude: Default should probably change. It is `repulsion` in the PFT code, and Dipy is also changing.

scripts/scil_compute_tracking_dipy.py
frheault: The other script has "_dipy" at the end, do we care to put it on both, remove one ?

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+from dipy.data import SPHERE_FILES, get_sphere
+from dipy.direction import (DeterministicMaximumDirectionGetter,
+                            ProbabilisticDirectionGetter)
+from dipy.direction.peaks import PeaksAndMetrics
+from dipy.tracking.local import BinaryTissueClassifier, LocalTracking
+from dipy.tracking.streamlinespeed import length, compress_streamlines
+from dipy.tracking import utils as track_utils
+
+import nibabel as nib
+from nibabel.streamlines import Field
```
jchoude: Unused.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+from nibabel.orientations import aff2axcodes
```
jchoude: Unused.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+    add_overwrite_arg(out_g)
+
+    log_g = p.add_argument_group('Logging options')
+    add_verbose(log_g)
+
+    return p
+
+
+def _get_direction_getter(args, mask_data):
+    sh_data = nib.load(args.sh_file).get_data().astype('float64')
```
jchoude: If you instead used `nib.get_fdata()` wouldn't you already get your `float64` data?

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+        help='Random number generator seed.')
+    add_overwrite_arg(out_g)
+
+    log_g = p.add_argument_group('Logging options')
+    add_verbose(log_g)
+
+    return p
+
+
+def _get_direction_getter(args, mask_data):
```
jchoude: This could go in a module.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+
+    assert_inputs_exist(parser, [args.sh_file, args.seed_file, args.mask_file])
+    assert_outputs_exists(parser, args, [args.output_file])
+
+    if not nib.streamlines.is_supported(args.output_file):
+        parser.error('Invalid output streamline file format (must be trk or ' +
+                     'tck): {0}'.format(args.output_file))
+
+    if not args.min_length > 0:
+        parser.error('minL must be > 0, {}mm was provided.'
```
jchoude: Name of the argument changed.

scripts/scil_compute_tracking_dipy.py
```diff
@@ -0,0 +1,265 @@
...
+                     .format(args.min_length))
+    if args.max_length < args.min_length:
+        parser.error('maxL must be > than minL, (minL={}mm, maxL={}mm).'
```
jchoude: Name of arg changed.

scilpy/tracking/tools.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+DETERMINISTIC = 'det'
```
jchoude: Since those are defined here, you should import the definitions in your other script instead of repeating it.

scripts/scil_compute_pft_dipy.py
```diff
@@ -102,12 +104,12 @@ def _build_args_parser():
...
         'When no option is provided, uses --npv 1.')
-    seed_group.add_argument(
+    seed_sub_exclusive = seed_group.add_mutually_exclusive_group()
+    seed_sub_exclusive.add_argument(
         '--npv', type=int,
         help='Number of seeds per voxel.')
-    seed_group.add_argument(
-        '--nt', type=int,
-        help='Total number of seeds to use.')
+    seed_sub_exclusive.add_argument('--nt', type=int,
```
jchoude: Your alignment here is not the same as the others (changing line after the `(`)

scripts/scil_compute_pft_dipy.py
```diff
@@ -176,36 +175,27 @@ def main():
...
     if args.npv and args.npv <= 0:
-        parser.error("Number of seeds per voxel must be > 0.")
+        parser.error('Number of seeds per voxel must be > 0.')
 
     if args.nt and args.nt <= 0:
-        parser.error("Total number of seeds must be > 0.")
+        parser.error('Total number of seeds must be > 0.')
 
     fodf_sh_img = nib.load(args.sh_file)
+    fodf_sh_img = nib.load(args.sh_file)
```
jchoude: Repetition?

scripts/scil_compute_tracking_dipy.py
jchoude: We should probably remove it in for the other as well.

However, this would need to be coordinated with @GuillaumeTh and @arnaudbore for Tractoflow.

jchoude:
nvm the old comments. Previous version.

jchoude:
To me this would be good to go.

PR #14, opened 2019-06-25
Futurize streamlines compression

frheault:
Simple transfert of scripts, remove dependency from TractConverter while keeping the lazy_loading/saving feature and tested for python 3

### DIFF
--- a/scilpy/io/streamlines.py
+++ b/scilpy/io/streamlines.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+
+import nibabel as nib
+
+
+def check_tracts_same_format(parser, tractogram_1, tractogram_2):
+    """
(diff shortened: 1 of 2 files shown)


scilpy/io/streamlines.py
```diff
@@ -0,0 +1,25 @@
...
+    if not nib.streamlines.is_supported(tractogram_1):
+        parser.error('Format of "{}" not supported.'.format(tractogram_1))
+    if not nib.streamlines.is_supported(tractogram_2):
+        parser.error('Format of "{}" not supported.'.format(tractogram_2))
+
+    ext_1 = os.path.splitext(tractogram_1)[1]
+    ext_2 = os.path.splitext(tractogram_2)[1]
+    if not ext_1 == ext_2:
+        parser.error(
+            'Input and output tractogram file must use the same format.')
```
jchoude: `file` -> `files`

PR #16, opened 2019-06-27
Futurize streamlines math

frheault:
Transfert from bitbucket
Tested with Python3

### DIFF
--- a/scilpy/utils/streamlines.py
+++ b/scilpy/utils/streamlines.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from functools import reduce
+import itertools
+
+import numpy as np
+
+
+MIN_NB_POINTS = 10
+KEY_INDEX = np.concatenate((range(5), range(-1, -6, -1)))
+
+
+def get_streamline_key(streamline, precision=None):
+    # Use just a few data points as hash key. I could use all the data of
+    # the streamlines, but then the complexity grows with the number of
+    # points.
+    if len(streamline) < MIN_NB_POINTS:
+        key = streamline.copy()
+    else:
+        key = streamline[KEY_INDEX].copy()
+
+    if precision is not None:
+        key = np.round(key, precision)
+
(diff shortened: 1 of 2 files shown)


scilpy/utils/streamlines.py
```diff
@@ -1,9 +1,117 @@
...
+
+def subtraction(left, right):
+    """Subtraction of two streamlines dict (see hash_streamlines)"""
+    return {k: v for k, v in left.items() if k not in right}
+
+
+def union(left, right):
+    """Union of two streamlines dict (see hash_streamlines)"""
+
+    # In python 3 : return {**left, **right}
```
jchoude: If we decide to just support Python 3, we could use that syntax

jchoude:
Apart from @GuillaumeTh comments this is good to go for me.

scripts/scil_streamlines_math.py
```diff
@@ -0,0 +1,176 @@
...
+    p.add_argument('--save_metadata_indices', '-m', action='store_true',
+                   help='Save streamline indices to metadata. Has no '
+                   'effect if --no-data\nis present. Will '
+                   'overwrite \'ids\' metadata if already present.')
+
+    p.add_argument('--save_indices', '-s', metavar='OUTPUT_INDEX_FILE',
+                   help='Save the streamline indices to the supplied '
+                   'json file.')
+
+    p.add_argument('--verbose', '-v', action='store_true', dest='verbose',
```
GuillaumeTh: Please use function

scripts/scil_streamlines_math.py
```diff
@@ -0,0 +1,176 @@
...
+    args = p.parse_args()
+
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+
+    assert_inputs_exist(p, args.inputs)
+    assert_outputs_exists(p, args, [args.output])
+
+    # Load all input streamlines.
+    data = [load_data(f) for f in args.inputs]
```
GuillaumeTh: Please update with dipy class

PR #17, opened 2019-06-28
Minor correction for transform scripts

frheault:
Adress JC's comments post-PR about the transform scripts


scilpy/io/image.py
```diff
@@ -5,12 +5,12 @@
```
arnaudbore: get_data is not useful anymore and import nibabel is missing.

PR #33, opened 2019-08-05
Add script scil_count_Streamlines.py

arnaudbore:
One reviewer would be enough...

### DIFF
--- a/scripts/scil_count_streamlines.py
+++ b/scripts/scil_count_streamlines.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+
+import argparse
+import json
+import os
+
+import nibabel as nib
+
+from scilpy.io.utils import assert_inputs_exist
+
+
+def _build_arg_parser():
+    parser = argparse.ArgumentParser(
+        description='Returns the number of streamlines in a bundle',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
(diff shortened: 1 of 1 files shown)


scilpy/image/utils.py
```diff
@@ -9,7 +9,6 @@
 def count_non_zero_voxels(image):
     """
     Count number of non zero voxels
-
```
GuillaumeTh: Why do you remove this enter ?

scripts/scil_count_streamlines.py
```diff
@@ -0,0 +1,42 @@
...
+
+    bundle_name, _ = os.path.splitext(os.path.basename(args.bundle))
+    bundle_tractogram_file = nib.streamlines.load(args.bundle, lazy_load=True)
+    stats = {
+        bundle_name: {
+            'tract_count': int(bundle_tractogram_file.header['nb_streamlines'])
+        }
+    }
+
+    print(json.dumps(stats, indent=args.indent))
```
GuillaumeTh: Maybe we can add a option to save the information in a file too ?

scripts/scil_count_streamlines.py
```diff
@@ -0,0 +1,42 @@
...
+import os
+
+import nibabel as nib
+
+from scilpy.io.utils import assert_inputs_exist
+
+
+def _build_arg_parser():
+    parser = argparse.ArgumentParser(
+        description='Returns the number of streamlines in a bundle',
```
GuillaumeTh: Returns -> Return 
bundle -> tractogram

PR #38, opened 2019-08-09
NF count_non_zero_voxels

arnaudbore:
Just need one review/approval for this PR

### DIFF
--- a/scilpy/image/utils.py
+++ b/scilpy/image/utils.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import nibabel as nib
+import numpy as np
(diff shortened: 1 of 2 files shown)


scilpy/image/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    if isinstance(image, six.string_types):
+        nb_object = nib.load(image)
+    else:
+        nb_object = image
+
+    data = nb_object.get_data(caching='unchanged')
+
+    # Count the number of non-zero voxels.
+    if len(data.shape) >= 4:
```
frheault: Just wondering, should there be an option to have a true voxel count ?
Does anyone ever need that ?

PR #50, opened 2019-09-18
Futurize registration tractogram

arnaudbore:
New script:
- scil_register_tractogram.py

You can look at the doc/tractogram_registration.md to test it.


doc/tractogram_registration.md
```diff
@@ -31,4 +31,3 @@ scil_apply_warp_to_tractogram.py MOVING_FILE REFERENCE_FILE DEFORMATION_FILE OUT
 * The DEFORMATION_FILE needs to be the InverseWarp.nii.gz (very important)
 
 *The file outputted by this script is a tractogram*
-
```
GuillaumeTh: Why ? Please revert.

scripts/scil_register_tractogram.py
```diff
@@ -0,0 +1,118 @@
...
+(http://www.sciencedirect.com/science/article/pii/S1053811915003961)
+"""
+
+
+def register_tractogram(moving_filename, static_filename,
+                        only_rigid, amount_to_load, matrix_filename,
+                        verbose):
+
+    amount_to_load = max(250000, amount_to_load)
+    moving_tractogram = nib.streamlines.load(moving_filename, lazy_load=True)
```
GuillaumeTh: @frheault Can we use the stateful_tractogram here ?

scripts/scil_register_tractogram.py
```diff
@@ -0,0 +1,123 @@
...
+                   help='Path of the moving tractogram (*.trk).')
+
+    p.add_argument('static_file',
+                   help='Path of the target tractogram (*.trk).')
+
+    p.add_argument('--out_name',
+                   default='transformation.npy',
+                   help='Filename of the transformation matrix, \n'
+                        'the registration type will be appended as a suffix,\n'
+                        '[transformation_affine/rigid.npy]')
```
boiteaclou: should be "transformation_<affine/rigid>.npy" or "<out_name>_<affine/rigid>.npy"

scripts/scil_register_tractogram.py
arnaudbore: Done!
@frheault can you validate ?

scripts/scil_register_tractogram.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _build_args_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=DESCRIPTION, epilog=EPILOG)
+
+    p.add_argument('moving_tractogram',
+                   help='Path of the moving tractogram (*.trk).')
+
+    add_reference(p, 'moving_tractogram')
```
GuillaumeTh: I don't like that the reference argument is displayed as an option. I think for this specific use case you can copy paste twice the reference argument and don't use the function. Moreover I'm not a fan of the reference function but it is just a personal opinion :smile:

scripts/scil_register_tractogram.py
```diff
@@ -0,0 +1,128 @@
...
+                          verbose=verbose)
+    _, transfo, _, _ = ret
+    np.savetxt(matrix_filename, transfo)
+
+
+def _build_args_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=DESCRIPTION, epilog=EPILOG)
+
+    p.add_argument('moving_tractogram',
```
GuillaumeTh: If it is a trk do we must to require an reference ?

frheault: I think this is just an old help message, there is no limitation on the filetype

scripts/scil_register_tractogram.py
frheault: What do you mean copy paste twice ? Having two --reference with the same name ? or using the same reference twice (not valid for both tractogram)

Having a input specific reference is required. Either only trk are supported, or all the verification are done in the scripts and we do not use a function, or the way @arnaudbore did it. But there is not much possibilities

scilpy/io/streamlines.py
```diff
@@ -91,18 +91,54 @@ def save_from_voxel_space(streamlines, anat, ref_tracts, out_name):
...
+                                   bbox_check=True, argName=None):
+
     _, ext = os.path.splitext(filepath)
     if ext == '.trk':
         sft = load_tractogram(filepath, 'same',
                               bbox_valid_check=bbox_check)
     elif ext in ['.tck', '.fib', '.vtk', '.dpy']:
-        if args.reference is None:
+        if argName:
+            if args.__getattribute__(argName + '_ref'):
```
frheault: I would put that in a variable and re-use it (also a bit easier to understand if the variable name is clear)

scilpy/io/streamlines.py
```diff
@@ -91,18 +91,54 @@ def save_from_voxel_space(streamlines, anat, ref_tracts, out_name):
...
+    sequence = iter(sequence)
+    chunk = list(islice(sequence, n))
+    while len(chunk) > 0:
+        yield chunk
+        chunk = list(islice(sequence, n))
+
+
 def load_tractogram_with_reference(parser, args, filepath,
-                                   bbox_check=True):
+                                   bbox_check=True, argName=None):
```
frheault: underscore, no camel case

scripts/scil_register_tractogram.py
```diff
@@ -0,0 +1,128 @@
...
+                   help='Path of the moving tractogram (*.trk).')
+
+    add_reference(p, 'moving_tractogram')
+
+    p.add_argument('static_tractogram',
+                   help='Path of the target tractogram (*.trk).')
+
+    add_reference(p, 'static_tractogram')
```
frheault: I would put them after the positional argument, not interleave

scripts/scil_register_tractogram.py
GuillaumeTh: Ok I understand nvm

PR #51, opened 2019-09-18
Move tractogram function to io streamlines and typo

arnaudbore:
Following JC comment PR42 
load_tractogram_with_reference is now in `scilpy/io/streamlines.py`


jchoude:
Doesn't work. You're missing the `load_tractogram` import in `scilpy.io.streamlines`, and it is probably unused now in `scilpy.io.utils.`.

Did you test? ;)

PR #56, opened 2019-10-16
STYLE: Make `utils` docstrings' style be consistent

jhlegarreta:
Make `utils` docstrings' style be consistent: stick to the NumPy/SciPy
style.


jchoude:
This is a simple PR, I'll go ahead and merge it. Thanks @jhlegarreta

PR #57, opened 2019-10-23
Improving font of mosaic

EmmaRenauld:
The font was really small on my screen for the mosaic of the bundles.

As is, the script does not consider the arg "font size" if no font name was given, but it is not explained in the -h help.

I added a try: exception that tries to find a truetype font on the computer. This way you don't have to remember where the fonts are everytime you want to use the script and change the size. Should work on most linux versions. Else, explains that it can't and uses default, like before.

### DIFF
--- a/scripts/scil_visualize_bundles_mosaic.py
+++ b/scripts/scil_visualize_bundles_mosaic.py
@@ -59,6 +59,20 @@ def get_font(args):
             print('Font {} was not found. '
                   'Default font will be used.'.format(args.ttf))
             font = ImageFont.load_default()
+    elif args.ttf_size is not None:
+        # default font is not a truetype font, so size can't be changed.
+        # to allow users to change the size without having to know where fonts
+        # are in their computer, we could try to find a truetype font ourselves.
+        # They are often present in /usr/share/fonts/
+        font_path = '/usr/share/fonts/truetype/freefont/FreeSans.ttf'
+        try:
+            font = ImageFont.truetype(font_path, args.ttf_size)
+        except Exception:
+            print('You did not specify a font. It is difficult'
+                  'for us to adjust size. We tried on font {} '
+                  'but it was not found.'
+                  'Default font will be used, for which font cannot be changed.'.format(font_path))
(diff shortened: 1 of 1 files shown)


scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -59,6 +59,20 @@ def get_font(args):
...
                   'Default font will be used.'.format(args.ttf))
             font = ImageFont.load_default()
+    elif args.ttf_size is not None:
+        # default font is not a truetype font, so size can't be changed.
+        # to allow users to change the size without having to know where fonts
+        # are in their computer, we could try to find a truetype font ourselves.
+        # They are often present in /usr/share/fonts/
+        font_path = '/usr/share/fonts/truetype/freefont/FreeSans.ttf'
+        try:
+            font = ImageFont.truetype(font_path, args.ttf_size)
```
GuillaumeTh: I'm ok with this fix but I'm curious. Did you try to set the font to None and set the size ?

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -59,6 +59,20 @@ def get_font(args):
...
+        except Exception:
+            print('You did not specify a font. It is difficult'
```
GuillaumeTh: Please use logging.error instead of print,

scripts/scil_visualize_bundles_mosaic.py
arnaudbore: It will return an error from `ImageFont.py`

arnaudbore:
> I think we should use the font if it's provided by the user if it doesn't work switch to FreeSans.ttf and finally if it doesn't find it print (with logging) an explicit message.

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -59,6 +59,20 @@ def get_font(args):
             print('Font {} was not found. '
```
arnaudbore: Use logging

PR #62, opened 2019-11-13
Futurize tractometry 1

arnaudbore:
First of many PRs for tractometry.

### DIFF
--- a/scilpy/tracking/tools.py
+++ b/scilpy/tracking/tools.py
@@ -33,9 +33,9 @@ def filter_streamlines_by_length(sft, min_length=0., max_length=np.inf):
         List of filtered streamlines by length.
 
     filtered_data_per_point: dict
-        dict of data per point for filtered streamlines
+        dict of data per point for filtered streamlines.
     filtered_data_per_streamline: dict
-        dict of data per streamline for filtered streamlines
+        dict of data per streamline for filtered streamlines.
     """
 
     lengths = []
--- a/scilpy/tracking/tools.py
+++ b/scilpy/tracking/tools.py
@@ -8,10 +8,7 @@ from dipy.tracking.metrics import length, downsample
 from dipy.tracking.streamline import set_number_of_points
 
 
-def filter_streamlines_by_length(streamlines,
-                                 data_per_point,
-                                 data_per_streamline,
-                                 min_length=0., max_length=np.inf):
+def filter_streamlines_by_length(sft, min_length=0., max_length=np.inf):
     """
     Filter streamlines using minimum and max length
 
@@ -42,7 +39,7 @@ def filter_streamlines_by_length(streamlines,
     """
 
     lengths = []
-    for streamline in streamlines:
+    for streamline in sft.streamlines:
         lengths.append(length(streamline))
 
     lengths = np.asarray(lengths)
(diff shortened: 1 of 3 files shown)


scilpy/tracking/tools.py
```diff
@@ -36,23 +33,23 @@ def filter_streamlines_by_length(streamlines,
...
 
     lengths = np.asarray(lengths)
 
     filter_stream = np.logical_and(lengths >= min_length,
                                    lengths <= max_length)
 
-    filtered_streamlines = list(np.asarray(streamlines)[filter_stream])
-    filtered_data_per_point = data_per_point[filter_stream]
-    filtered_data_per_streamline = data_per_streamline[filter_stream]
+    filtered_streamlines = list(np.asarray(sft.streamlines)[filter_stream])
```
frheault: Casting streamlines as a numpy array ? This will crash if streamline do not have the same number of point, right ?

scripts/scil_filter_streamlines_by_length.py
```diff
@@ -16,14 +22,20 @@ def _build_args_parser():
         description='Filter streamlines by length.')
     p.add_argument('in_tractogram', type=str,
                    help='Streamlines input file name.')
+
+    add_reference(p)
```
frheault: This is technically shown as optional in the help (with a --). Should this be lower with the other ?

scripts/scil_filter_streamlines_by_length.py
```diff
@@ -16,14 +22,20 @@ def _build_args_parser():
         description='Filter streamlines by length.')
     p.add_argument('in_tractogram', type=str,
```
frheault: Already a string by default

scripts/scil_filter_streamlines_by_length.py
```diff
@@ -16,14 +22,20 @@ def _build_args_parser():
...
                    help='Streamlines input file name.')
+
+    add_reference(p)
+
     p.add_argument('out_tractogram', type=str,
```
frheault: Already a string by default

scilpy/tracking/tools.py
```diff
@@ -36,23 +33,23 @@ def filter_streamlines_by_length(streamlines,
...
-        dict of data per point for filtered streamlines
+        dict of data per point for filtered streamlines.
     filtered_data_per_streamline: dict
-        dict of data per streamline for filtered streamlines
+        dict of data per streamline for filtered streamlines.
     """
 
     lengths = []
-    for streamline in streamlines:
+    for streamline in sft.streamlines:
```
boiteaclou: Could be replaced by a np.apply_along_axis (gives you a numpy array for free as output also)

scripts/scil_compute_streamlines_length_stats.py
```diff
@@ -0,0 +1,54 @@
...
+        description='Compute streamlines min, mean and max length, as well as '
+                    'standard deviation of length in mm.',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file.')
+
+    add_reference(p)
+
+    p.add_argument('--indent',
```
boiteaclou: Following arguments can be placed in a group "Json output format"

scripts/scil_filter_streamlines_by_length.py
```diff
@@ -1,29 +1,39 @@
...
+    p.add_argument('in_tractogram',
                    help='Streamlines input file name.')
-    p.add_argument('out_tractogram', type=str,
+    p.add_argument('out_tractogram',
                    help='Streamlines output file name.')
     p.add_argument('--minL', default=0., type=float,
                    help='Minimum length of streamlines. [%(default)s]')
     p.add_argument('--maxL', default=np.inf, type=float,
                    help='Maximum length of streamlines. [%(default)s]')
+    p.add_argument('--no_empty', action='store_true',
```
boiteaclou: Is there a use case where we would allow an empty output file ? If not, I would remove the argument and just return the debug message (or an exception)

arnaudbore: Some of our pipeline or collaborators still need this option.

PR #63, opened 2019-11-13
Futurize assign color to trk

arnaudbore:


scripts/scil_assign_color_to_trk.py
```diff
@@ -0,0 +1,66 @@
...
+                    'formatted as 0xRRGGBB',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help='Tractogram.')
+
+    p.add_argument('out_tractogram',
+                   help='Colored TRK tractogram.')
+    p.add_argument('color',
+                   help='Hexadecimal RGB color (ie. 0xRRGGBB).')
```
frheault: On wiki, Google converter and first website, hexadecimal color are always represented as such "#51f542". The # instead of 0x. My question is, why is this different ?

scripts/scil_assign_color_to_trk.py
```diff
@@ -0,0 +1,69 @@
...
+from scilpy.io.utils import (assert_inputs_exist,
+                             assert_outputs_exist,
+                             add_overwrite_arg)
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description='Assign an hexadecimal RGB color to a Trackvis TRK '
+                    'tractogram. The hexadecimal RGB color should be '
+                    'formatted as 0xRRGGBB',
```
GuillaumeTh: or #RRGGBB

scripts/scil_assign_color_to_trk.py
```diff
@@ -0,0 +1,69 @@
...
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help='Tractogram.')
```
GuillaumeTh: Please add : `Must be in trk format` or something like this.

scripts/scil_assign_color_to_trk.py
```diff
@@ -0,0 +1,69 @@
...
+
+    p.add_argument('out_tractogram',
+                   help='Colored TRK tractogram.')
+    p.add_argument('color',
+                   help='Can be either hexadecimal (ie. #RRGGBB or 0xRRGGBB).')
```
GuillaumeTh: Doesn't work properly with # format. The parser don't take the argument. We must to write the color as "#RRGGBB".

scripts/scil_assign_color_to_trk.py
arnaudbore: Input does not have to be TRK

GuillaumeTh: yes my bad. sorry

PR #69, opened 2019-11-27
Residuals

EmmaRenauld:
Note that previous definition for "outlier volumes" was wrong. It was checking outlier voxels.
Corrected. The definition of "outlier volume" is still not defined.

+ added new figure to replace the temporary one.

### DIFF
--- a/scripts/scil_compute_dti_metrics.py
+++ b/scripts/scil_compute_dti_metrics.py
@@ -405,10 +405,10 @@ def main():
         medianprops = dict(linestyle='-', linewidth=2.5, color='firebrick')
         meanprops = dict(linestyle='-', linewidth=2.5, color='green')
         axe[0, 0].bxp(stats, showmeans=True, meanline=True, showfliers=False,
-                     medianprops=medianprops, meanprops=meanprops)
+                      medianprops=medianprops, meanprops=meanprops)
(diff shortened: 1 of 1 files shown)


scripts/scil_compute_dti_metrics.py
```diff
@@ -341,26 +339,52 @@ def main():
...
         nib.save(R_img, args.residual)
 
-        R_k = np.zeros(data.shape[-1])  # mean residual per DWI
+        # Each volume's residual statistics
+        if args.mask is None:
+            logger.info("Outlier detection will not be performed, since no "
+                        "mask was provided.")
+        stats = [dict.fromkeys(['label', 'mean', 'iqr', 'cilo', 'cihi', 'whishi',
+                                'whislo', 'fliers', 'q1', 'med', 'q3'], [])
+                 for i in range(data.shape[-1])]  # stats with format for boxplots
```
GuillaumeTh: I think something is wrong here ?

EmmaRenauld: Can you explain? It's producing a list of dict for i in ...

GuillaumeTh: Ok my bad it's ok. I didn't see that it was for stats.

PR #72, opened 2019-12-04
Futurize bundle screenshot

frheault:
Transfert of the scripts, adapted to dipy 1.0.

Switched the template to any mni152 for simplicity for dti and bundle screenshot
The visualisation is now all in rasmm, meaning the mni152 template can be cropped or resampled and it will work.

Added the local_coloring option and the reference coloring now support any colormap instead of the default JET.

data for testing:
https://drive.google.com/open?id=1imRdCPI0KTymLw7TEKSEjkfVqt9s_5lk

### DIFF
--- a/scripts/scil_screenshot_bundle.py
+++ b/scripts/scil_screenshot_bundle.py
@@ -60,7 +60,6 @@ def _build_args_parser():
 def prepare_data_for_actors(bundle_filename, reference_filename,
                             target_template_filename):
     sft = load_tractogram(bundle_filename, reference_filename)
-    # sft.to_vox()
     streamlines = sft.streamlines
 
     # Load and prepare the data
@@ -71,7 +70,6 @@ def prepare_data_for_actors(bundle_filename, reference_filename,
     target_template_img = nib.load(target_template_filename)
     target_template_data = target_template_img.get_data()
     target_template_affine = target_template_img.affine
-    zooms = 1 / float(target_template_img.header.get_zooms()[0])
 
     # Register the DWI data to the template
     transformed_reference, transformation = register_image(target_template_data,
@@ -81,7 +79,6 @@ def prepare_data_for_actors(bundle_filename, reference_filename,
     # transformation = np.eye(4)
     streamlines = transform_streamlines(streamlines,
                                         np.linalg.inv(transformation))
-    # streamlines = transform_streamlines(streamlines, zooms * np.eye(4))
 
     return streamlines, transformed_reference
 
@@ -118,7 +115,6 @@ def main():
 
     # Get the relevant slices from the template
     target_template_img = nib.load(args.target_template)
-    zooms = 1
 
     x_slice = int(target_template_img.shape[0] / 2)
     y_slice = int(target_template_img.shape[1] / 2)
@@ -140,7 +136,7 @@ def main():
                   args.uniform_coloring[1] / 255.0,
                   args.uniform_coloring[2] / 255.0)
     elif args.reference_coloring:
-        colors = reference
+        colors = reference_data
     else:
         colors = None
 
--- a/scripts/scil_screenshot_bundle.py
+++ b/scripts/scil_screenshot_bundle.py
@@ -3,11 +3,14 @@
 import argparse
 import os
 
+from dipy.io.stateful_tractogram import Space, StatefulTractogram
 from dipy.io.streamline import load_tractogram
 from dipy.tracking.streamline import transform_streamlines
 from fury import actor
+import matplotlib.pyplot as plt
 import nibabel as nib
 import numpy as np
+from scipy.ndimage import map_coordinates
 
 from scilpy.io.utils import (add_overwrite_arg,
                              assert_inputs_exist,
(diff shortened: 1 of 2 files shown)


scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+   Sagittal can be capture from the left (default) or the right.
+   """
+
+
+def _build_args_parser():
+    p = argparse.ArgumentParser(
+        description=DESCRIPTION,
+        formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_bundle',
+                   help='Path of the input bundle')
```
arnaudbore: We should expand this to all format.

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+    p.add_argument('in_anat',
```
boiteaclou: Shouldn't this be added via add_reference ?

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+        else:
+            output_filenames.append(os.path.join(args.output_dir,
+                                                 '{0}.png'.format(axis_name)))
+
+    assert_outputs_exist(parser, args, output_filenames)
+
+    if args.output_dir and not os.path.isdir(args.output_dir):
+        os.mkdir(args.output_dir)
+
+        if args.anat_opacity < 0.0 or args.anat_opacity > 1.0:
```
boiteaclou: Bad indentation

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+    p.add_argument('--right', action='store_true',
+                   help='Take screenshot from the right instead of the left \n'
+                        'for the sagittal plane')
+    p.add_argument('--anat_opacity', type=float, default=0.3,
+                   help='Set the opacity for the anatomy, use 0 for complete '
+                   'transparency, 1 for opaque')
+    p.add_argument('--output_suffix',
+                   help='Add a suffix to the output, else the axis name is used')
+    p.add_argument('--output_dir', default='',
+                   help='Put all images in a specific directory, will overwrite')
```
boiteaclou: The overwrite param makes it sure that it will not overwrite, update help message.

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+                  args.uniform_coloring[1] / 255.0,
+                  args.uniform_coloring[2] / 255.0)
+    elif args.reference_coloring:
+        sft.to_vox()
+        streamlines_vox = sft.get_streamlines_copy()
+        sft.to_rasmm()
+        colors = []
+        normalized_data = reference_data / np.max(reference_data)
+        cmap = plt.get_cmap(args.reference_coloring)
+        for points in streamlines_vox:
```
boiteaclou: This could be replaced by an apply_along_axis since streamlines_vox is an ArraySequence

scripts/scil_screenshot_bundle.py
frheault: This one is mandatory for the script, and serve as a reference for registration (first purpose)

scripts/scil_screenshot_bundle.py
frheault: Not sure how to call it, I can't make it work.
Will ask for help IRL

frheault: I don't know how to call it, can't make it work.
Will ask for help IRL (if you know how to use it)

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+
+from scilpy.io.utils import (add_overwrite_arg,
+                             assert_inputs_exist,
+                             assert_outputs_exist)
+from scilpy.utils.image import register_image
+from scilpy.viz.screenshot import display_slices
+
+DESCRIPTION = """
+   Register bundle to a template for screenshots using a reference.
+   The template are in /mnt/braindata/Other/simple_template_viz/
```
arnaudbore: Should not use something from our own repository

arnaudbore: And don't forget to add no skull.

scripts/scil_screenshot_bundle.py
```diff
@@ -0,0 +1,195 @@
...
+    p = argparse.ArgumentParser(
+        description=DESCRIPTION,
+        formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_bundle',
+                   help='Path of the input bundle')
+    p.add_argument('in_anat',
+                   help='Path of the reference file (.nii or nii.gz)')
+    p.add_argument('target_template',
+                   help='Path to the target MNI152template for registration, '
+                        'Any choice of modality works')
```
arnaudbore: no skull here also or if skull in anat -> skull on template as well.

scripts/scil_screenshot_bundle.py
frheault: We could put an MNI152 (with an wmparc) it is always useful for 3MB of space

frheault:
@arnaudbore I would say it is ready, what do you want to do for the mni template ? (storage)

scripts/scil_screenshot_bundle.py
arnaudbore: It will be in the singularity not in the scilpy repo. Please update DESCRIPTION

frheault:
For the MNI template, like I said since it is any MNI152 in any resolution, no need to store them.

PR #78, opened 2019-12-12
Add installation instructions for developpers

EmmaRenauld:


arnaudbore:
Doesn't fit with the new PR#87

PR #79, opened 2019-12-12
Futurize tractometry 3

arnaudbore:
New script: scil_compute_centroid.py


scripts/scil_compute_centroid.py
```diff
@@ -0,0 +1,82 @@
...
+    p = argparse.ArgumentParser(
+        description='Compute bundle centroid',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file.')
+    p.add_argument('out_centroid',
+                   help='Output centroid streamline filename.')
+
+    p.add_argument('--distance_thres',
```
frheault: If the goal is to obtain a single centroid and nothing else. Why is this even a parameters, let's just use 1000000 ?
The goal is to average ALL streamlines into a single centroid. anything too low will crash, it doesn't support negative. Simply using infinity could work 100% of the time, no ?

frheault: Or to save a ton of line of code. Manually resample to NBR_POINTS and then simply use np.average on the appropriate axis.

arnaudbore: I agree

PR #80, opened 2019-12-12
Futurize tractometry 4

arnaudbore:
Two scripts to compute map using endpoints. 
I am not really happy with these names. Don't hesitate if you have some suggestions.


scripts/scil_endpoints_map.py
```diff
@@ -0,0 +1,111 @@
...
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=DESCRIPTION,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_bundle',
+                   help='Fiber bundle filename.')
+    p.add_argument('reference',
+                   help='Reference anatomic filename.')
+    p.add_argument('endpoints_map_head',
```
boiteaclou: It's not really clear the next two arguments are outputs (use groups ?)

scripts/scil_endpoints_metric.py
```diff
@@ -0,0 +1,127 @@
...
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description='Projects metrics onto the endpoints of streamlines. The '
+                    'idea is to visualize the cortical areas affected by '
+                    'metrics (assuming streamlines start/end in the cortex).',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file.')
+
```
boiteaclou: Unneeded space

scripts/scil_endpoints_map.py
arnaudbore: I decided to improve the help.

scripts/scil_endpoints_map.py
```diff
@@ -0,0 +1,107 @@
...
+                             assert_inputs_exist,
+                             assert_outputs_exist)
+
+DESCRIPTION = '''
+Computes the endpoint map of a bundle. The endpoint map
+is simply a count of the number of streamlines that
+start or end in each voxel. The idea is to estimate the
+cortical areas affected by the bundle (assuming
+streamlines start/end in the cortex)
+'''
```
frheault: I know the concept of head and tail was confusing to philippe, maybe saying explicitely that if the streamlines are not aligned in X, Y or Z the head/tail are basically random and not really two coherent group.

arnaudbore: Copy paste :smile:

PR #83, opened 2019-12-17
remove not used vtk dependancy

arnaudbore:


jchoude:
Just to make sure, you remove this function because it's never used, right?

Are we sure it won't be used by any incoming PR?

PR #85, opened 2019-12-31
Added missing numpy include line creating error when building extensions

gauvinalexandre:


jchoude:
Thanks for that.

However, just to make sure, in which case did you get this error? 

I just tried building using Python 3.5, Cython 0.29.12, and didn't get any issue. Do you are a specific "how-to"?

gauvinalexandre:
I was simply building from a fresh new VM with the latest Linux Mint. The building error was indicating it couldn't find the *.h headers. I added the line and it did the trick. I was wondering why the other lines around had the "include" and not this specific one...

Would you like to have more details?

PR #87, opened 2020-01-17
Add basic test fix PIP and add CI travis

GuillaumeTh:
Add a test for each script that test --help option.

We can now use pip install.

A travis CI test multiple python version and run the tests.


arnaudbore:
Please modify README as well

GuillaumeTh:
@arnaudbore yes good catch. I will fix the conflict too

PR #67, opened 2019-11-19
Futurize - similarity summary script

frheault:
Transfert of the first summarize script (1/3)
- Support stateful_tractogram

This script comes with useful function for various similarity measures. The script looks very complex, it is due to the fact that it can support tens (or hundreds) of input and performs comparisons without busting memory with temporary files to accelerate redundant computation.

it was originally designed for Tractostorm, then used for Max&Laurent atlas comparison, RBX test-retest evaluation with Imeka and my RBX chapter for my thesis, could be useful for TractoInferno.

Here so test data and example commands:
https://drive.google.com/open?id=1nfSLIJpdzFDnFQ-OxyRVlBRbbPx061qe

PR #88, opened 2020-01-17
Add scil_compute_streamlines_dice.py script

cbedetti:
Implementation of `scil_compute_streamlines_weighted_dice.py` from bitbucket scilpy.
I removed weighted from the name since it's an option.

It works with two trks of mine.

Is there any testing data?

### DIFF
--- a/scripts/scil_compute_streamlines_dice.py
+++ b/scripts/scil_compute_streamlines_dice.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+
+"""
+Compute the dice coefficient between two streamline files.
+
+Option to weight the dice coefficient based on the number of tracts passing
+through a voxel
+"""
+
+
+import argparse
+import json
+
+import numpy as np
+
+from scilpy.io.streamlines import load_tractogram_with_reference
+from scilpy.io.utils import add_reference_arg, assert_inputs_exist
+from scilpy.tractanalysis.streamlines_metrics import compute_tract_counts_map
+
+
+def _build_arg_parser():
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+
+    parser.add_argument("in_bundle1",
+                        help="Tractogram filename. Format must be one of \n"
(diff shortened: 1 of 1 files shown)


scripts/scil_compute_streamlines_dice.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(
+        parser, args.in_bundle1, optional=args.reference)
+    assert_inputs_exist(
+        parser, args.in_bundle2, optional=args.reference)
```
arnaudbore: These four lines should be merged into one assert_inputs_exist call

arnaudbore:
You can provide your own data, but you're right we should provide a dataset to run some tests.

frheault:
There is code in PR #67 computing the Dice and WDice, I think it should be the same function in a module.
You could steal some of that code and I will deal with the conflict when PR #67 is ready to merge

scripts/scil_compute_streamlines_dice.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+
+"""
+Compute the dice coefficient between two streamline files.
```
boiteaclou: Would it be better to change streamline here by tractogram ?

scripts/scil_compute_streamlines_dice.py
```diff
@@ -0,0 +1,102 @@
...
+                        help="Tractogram filename. Format must be one of \n"
+                        "trk, tck, vtk, fib, dpy.")
+
+    add_reference_arg(parser)
+
+    parser.add_argument("--weighted", action="store_true",
+                        help="Weight the dice coefficient based on the \n"
+                        "number of tracts passing through a voxel")
+
+    parser.add_argument("--indent", type=int, default=2,
```
boiteaclou: By default, do we pretty print with 2 or 4 indent ?

cbedetti:
Reading PR #67 of @frheault , there's already all it needs.
I think this PR should be closed.

PR #92, opened 2020-01-20
print json tract count before/after filtering

arnaudbore:
Nothing to say...

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -22,6 +22,16 @@ def check_tracts_same_format(parser, filename_list):
             parser.error('All tracts file must use the same format.')
 
 
+def add_json_arg(parser):
+    g1 = parser.add_argument_group(title='Json options')
+    g1.add_argument('--indent',
+                    type=int, default=2,
+                    help='Indent for json pretty print.')
+    g1.add_argument('--sort_keys',
+                    action='store_true',
+                    help='Sort keys in output json.')
+
+
 def add_reference_arg(parser, arg_name=None):
     if arg_name:
         parser.add_argument('--'+arg_name+'_ref',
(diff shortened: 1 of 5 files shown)


scilpy/io/utils.py
```diff
@@ -22,6 +22,16 @@ def check_tracts_same_format(parser, filename_list):
             parser.error('All tracts file must use the same format.')
 
 
+def add_json_arg(parser):
```
jchoude: Should probably be `add_json_args` with an `s`, since you have more than 1 arg ;)

scripts/scil_filter_streamlines_by_length.py
```diff
@@ -72,6 +79,12 @@ def main():
 
     save_tractogram(new_sft, args.out_tractogram)
 
+    tc_bf = len(sft.streamlines)
+    tc_af = len(new_streamlines)
+    print(json.dumps({'tract_count_before_filtering': int(tc_bf),
```
jchoude: You added an argument (`--json`) to know if the user wants the values printed or not, but here you don't check if it is set.

scilpy/io/utils.py
```diff
@@ -22,6 +22,16 @@ def check_tracts_same_format(parser, filename_list):
...
+    g1 = parser.add_argument_group(title='Json options')
+    g1.add_argument('--indent',
+                    type=int, default=2,
+                    help='Indent for json pretty print.')
+    g1.add_argument('--sort_keys',
```
jchoude: Never used. I guess it's for the future in tractometry?

arnaudbore: exactly

scripts/scil_filter_streamlines_by_length.py
arnaudbore: fix

scilpy/io/utils.py
arnaudbore: fix

PR #94, opened 2020-01-27
Change flip tracts

gabe1496:
Work with .trk and .tck, .tck with a ref.


scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Flip tracts around specific axes.
```
jchoude: Since this script is now called `flip_streamlines`, please change `tracts` to `streamlines` or `tractogram`.

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
...
+
+IMPORTANT: this script should only be used in case of absolute necessity. It's
+better to fix the real tools than to force flipping tracts to have them fit in
```
jchoude: Same here for `tracts`

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
...
+def _build_args_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help='Path of the input tractogram file.')
+    p.add_argument('out_tractogram',
+                   help='Path of the output tractogram file.')
+
+    p.add_argument('axes', metavar='dimension',
```
jchoude: I think we could just remove the metavar and just have `axes` directly.

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
...
+    parser = _build_args_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, args.in_tractogram)
+    assert_outputs_exist(parser, args, args.out_tractogram)
+
+    sft = load_tractogram_with_reference(parser, args, args.in_tractogram)
+    sft.to_vox()
+    sft.to_corner()
+    if len(args.axes) < 1:
```
jchoude: Since the `axes` argument is set to `nargs='+'`, the argparser will automatically check if it is set. This check shouldn't be needed.

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
...
+        flip_vector[0] = -1.0
+    if 'y' in flip_axes:
+        flip_vector[1] = -1.0
+    if 'z' in flip_axes:
+        flip_vector[2] = -1.0
+
+    return flip_vector
+
+
+def get_tracts_bounding_box(tracts):
```
jchoude: This could be moved to a module, it will probably be useful to other scripts.

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,113 @@
...
+
+    global_min = np.min(mins, axis=0)
+    global_max = np.max(maxs, axis=0)
+
+    return global_min, global_max
+
+
+def get_shift_vector(sft):
+    dims = sft.dimensions
+    voxel_dim = sft.voxel_sizes
```
jchoude: I feel like `voxel_dim` should not be used in the new version of the script. When the streamlines are loaded, a `to_vox()` is applied. In that case, the shift vector shouldn't be scaled by the voxel sizes. Else, they are not displaced correctly.

I have an issue that I think is caused by this when trying to flip streamlines coming from a 1.25x1.25x1.25 dataset. Saving at the end fails because they are out of bounds.

EmmaRenauld:
There might be some vocabulary to clarify in the lab. In dwi_ml, (and maybe in tractography??) we use flip to mean that we inverse the beginning and end of the streamline.

Maybe we should change to reverse, and you could change to (?) reflexion? symmetry?

frheault:
@gabe1496 Did you fix JC comments ? Kurt would need that script on a larger scale soon.

scripts/scil_flip_streamlines.py
gabe1496: This function has been deleted and it is now included in StatefulTractogram.

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,99 @@
...
+
+    streamlines = sft.streamlines
+
+    for streamline in streamlines:
+        mod_streamline = streamline + shift_vector
+        mod_streamline *= flip_vector
+        mod_streamline -= shift_vector
+        flipped_streamlines.append(mod_streamline)
+
+    new_sft = StatefulTractogram(flipped_streamlines, sft, sft.space,
```
GuillaumeTh: Please use from_sft()

scripts/scil_flip_streamlines.py
```diff
@@ -0,0 +1,99 @@
...
+    if 'y' in flip_axes:
+        flip_vector[1] = -1.0
+    if 'z' in flip_axes:
+        flip_vector[2] = -1.0
+
+    return flip_vector
+
+
+def get_shift_vector(sft):
+    dims = sft.space_attribute[1]
```
GuillaumeTh: Change with space_attributes

jchoude:
```
There might be some vocabulary to clarify in the lab. In dwi_ml, (and maybe in tractography??) we use flip to mean that we inverse the beginning and end of the streamline.

Maybe we should change to reverse, and you could change to (?) reflexion? symmetry?
```

We normally call that uniformizing the streamlines endpoints.

PR #95, opened 2020-01-28
Change split volume labels

gabe1496:
-Add the directory ..data/LUT with the look up table
-Add the arguments output_dir and output_prefix
-Add the utilisation of assert_inputs_exist and assert_outputs_exist


scripts/scil_split_volume_labels.py
```diff
@@ -0,0 +1,178 @@
...
+    add_overwrite_arg(p)
+
+    return p
+
+
+def main():
+    # Get the valid LUT choices.
+    path = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)),
+        '../data/LUT')
```
arnaudbore: This path doesn't work for me and make everything crash.

gabe1496:
Voici deux liens donnant accès à des données pour le PR:
https://drive.google.com/file/d/1nlyu-Cc1UgpOVCFtoNQ7uRLkS5KS2Mh4/view?usp=sharing
https://drive.google.com/file/d/1GgSOhOMQ8Yi52V8LyzRPPNVwlWI8JHz1/view?usp=sharing

scripts/scil_split_volume_labels.py
frheault: Ok this is much more complex than I thought. Before everyone was git clone the repo.
Now we will have pip install scilpy (soon), setup develop, pip install -e ./ (or something).

These different way to install it give rise to many case. I personally dont know how to make it clean.

I suggest the input argument json, that we will still put in the repo, and the user will have to link it. This is likely safer.

frheault:
@gabe1496 we decided that the LUT should be an mandatory argument.
However in the description you could give a tip on where to find and/or a tip on how to write it.

scripts/scil_split_volume_labels.py
ppoulin91: What is the error? I don't see why we can't make this work with a relative path?

ppoulin91: I would suggest using pathlib to make this easier (python 3.4+):
`pathlib.Path(__file__).parent / "../data/LUT"`

scripts/scil_split_volume_labels.py
```diff
@@ -0,0 +1,178 @@
...
+    assert_inputs_exist(parser, required)
+
+    label_image = nib.load(args.label_image)
+
+    if not issubclass(label_image.get_data_dtype().type, np.integer):
+        parser.error('The label image does not contain integers. ' +
+                     'Will not process.\nConvert your image to integers ' +
+                     'before calling.')
+
+    label_image_data = label_image.get_data()
```
ppoulin91: `get_data()` is deprecated now, isn't it? You would need to call get_fdata(), but it supports only float data, so nibabel recommends using the following: `np.asanyarray(img.dataobj)`

scripts/scil_split_volume_labels.py
```diff
@@ -0,0 +1,178 @@
...
+
+    if not m:
+        raise argparse.ArgumentTypeError("'" + str_to_parse + "' is not a " +
+                                         "range of numbers. Expected forms " +
+                                         "like '0-5' or '2'.")
+
+    start = m.group(1)
+    end = m.group(2) or start
+
+    start = int(start, 10)
```
ppoulin91: Add a comment to clarify that `10` is the base used when casting a `str`

scripts/scil_split_volume_labels.py
```diff
@@ -0,0 +1,178 @@
...
+are extracted.
+
+New lookup tables can be added by adding new files to the
+scilpy/data/LUT/ directory.
+"""
+
+
+# Taken from http://stackoverflow.com/a/6512463
+def parseNumList(str_to_parse):
+    m = re.match(r'(\d+)(?:-(\d+))?$', str_to_parse)
```
ppoulin91: Add comment explaining the pattern being parsed, with examples please :)

scripts/scil_split_volume_labels.py
ppoulin91: Ok, just to make this clear for everyone, here is the problem:
Using relative paths in a script doesn't work, because the scripts may be installed in a separate directory as binaries (so it's impossible to find the module's location from the location of the script).
Thankfully, we can use the `inspect` module to find the location of the `scilpy` module:

`import inspect
import pathlib
import scilpy

module_path = inspect.getfile(scilpy)  # /home/USERNAME/git/scilpy/scilpy/__init__.py
lut_dir = str(pathlib.Path(module_path).parent.parent / "data/LUT")
`

ppoulin91:
@gabe1496 I updated th PR with the solution for the relative path problem

ppoulin91:
The PR was tested, everything works well for now, just waiting for the relative path modifs

scripts/scil_split_volume_labels.py
jchoude: Thanks @ppoulin91 that was indeed the original reason for using the hardcoded path.

jchoude:
Just a quick note that LUT should be a mandatory argument only for the "label" mode, right?

scripts/scil_split_volume_by_labels.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+
+import nibabel as nib
+import numpy as np
+
+import scilpy
+from scilpy.io.utils import (add_overwrite_arg, assert_inputs_exist,
+                             assert_outputs_exist)
+
+DESCRIPTION = """
```
frheault: use the _ _doc_ _ standard

scripts/scil_split_volume_by_labels.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import division
```
frheault: Delete this line

scripts/scil_split_volume_by_ids.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import division
```
frheault: Delete this lne

scripts/scil_split_volume_by_ids.py
```diff
@@ -0,0 +1,130 @@
...
+import os
+import re
+
+import nibabel as nib
+import numpy as np
+
+from scilpy.io.utils import (add_overwrite_arg, assert_inputs_exist,
+                             assert_outputs_exist)
+
+DESCRIPTION = """
```
frheault: Use the _ _doc_ _ standard

frheault:
A bit sad for the 50 lines of duplicated code. BUT a lot easier to use!
I tested it, everything works like before.

arnaudbore:
Please add tests for each new script.

scripts/scil_split_volume_by_labels.py
```diff
@@ -0,0 +1,122 @@
...
+    label_image = nib.load(args.label_image)
+
+    if not issubclass(label_image.get_data_dtype().type, np.integer):
+        parser.error('The label image does not contain integers. ' +
+                     'Will not process.\nConvert your image to integers ' +
+                     'before calling.')
+
+    label_image_data = label_image.get_fdata().astype(int)
+
+    if args.scilpy_lut is not None:
```
arnaudbore: if args.scilpy_lut:

scripts/scil_split_volume_by_labels.py
```diff
@@ -0,0 +1,122 @@
...
+    assert_inputs_exist(parser, required)
+
+    label_image = nib.load(args.label_image)
+
+    if not issubclass(label_image.get_data_dtype().type, np.integer):
+        parser.error('The label image does not contain integers. ' +
+                     'Will not process.\nConvert your image to integers ' +
+                     'before calling.')
+
+    label_image_data = label_image.get_fdata().astype(int)
```
arnaudbore: np.int instead of int

scripts/scil_split_volume_by_labels.py
```diff
@@ -0,0 +1,122 @@
...
+
+import nibabel as nib
+import numpy as np
+
+import scilpy
+from scilpy.io.utils import (add_overwrite_arg, assert_inputs_exist,
+                             assert_outputs_exist)
+
+
+def _build_args_parser(luts):
```
arnaudbore: _build_arg_parser instead of _build_args_parser

scripts/scil_split_volume_by_ids.py
```diff
@@ -0,0 +1,127 @@
...
+    end = int(end)
+
+    if end < start:
+        raise argparse.ArgumentTypeError("Range elements incorrectly " +
+                                         "ordered in '" + str_to_parse + "'.")
+
+    return list(range(start, end+1))
+
+
+def _build_args_parser():
```
arnaudbore: _build_arg_parser()

PR #96, opened 2020-01-28
Encapsulate compute_ssst_frf() in a separate function

ppoulin91:
Some questions:
- Should we move check_b0_threshold from the script to the function?
- Should we leave parameter checks/warnings in the script? Both?


ppoulin91:
For context: this PR is the first of many where we try to separate scripts from functions that we can import in the new scil_vital/dwi_ml repository.

EmmaRenauld:
You miss this at the beginning

#!/usr/bin/env python
# -*- coding: utf-8 -*-

boiteaclou:
This is my opinion, but I think that if the function should not be executed when the check don't pass, then the checks should be in the function and raise exceptions. Those exceptions should then be caught in the script for the logging messages. Else, if it is just on the format of inputs, then it should be only checked in the script.

ppoulin91:
@EmmaRenauld I added the UTF-8 tag, but not the shebang since it's not a script

AntoineTheb:
@boiteaclou Seems about right. imo the scripts should have as little "logic" in them as possible and only validate file inputs that would be otherwise already validated by other scripts also using the "logic" behind

arnaudbore:
I'm not convinced it should end into a folder named `dwi` but everything else LGTM.

ppoulin91:
Any other ideas for the folder name? It doesn't really fit with existing ones...

AntoineTheb:
Could it fit in the `reconst` folder ?

GuillaumeTh:
I agree with @AntoineTheb, it could fit in recont and maybe you can create a file 'frf.py'. Normally in the future we will have the multi-tissue response function.

jchoude:
Yes, `reconst` would be a good module for that.

I also agree with @boiteaclou about the checks on files vs checks on arguments.

ppoulin91:
Nothing else should change, anyone wants to review the last modifications before merging?

arnaudbore:
If you change check_b0_threshold you have to do the modification everywhere.

ppoulin91:
@arnaudbore I did, the only other usages were in compute_dti_metrics and compute_fodf

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
...
+from dipy.segment.mask import applymask
+
+from scilpy.utils.bvec_bval_tools import (is_normalized_bvecs, normalize_bvecs,
+                                          check_b0_threshold)
+
+
+def compute_ssft_frf(data, bvals, bvecs, mask=None, mask_wm=None,
+                     fa_thresh=0.7, min_fa_thresh=0.5, min_nvox=300,
+                     roi_radius=10, roi_center=None, force_b0_threshold=False):
+    """Compute a single Fiber Response Function from a DWI.
```
jchoude: Maybe specify that this is for single shell data (or, at least, under b=1500). That the reason it's `ssst` response.

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
...
+import numpy as np
+from dipy.core.gradients import gradient_table
+from dipy.reconst.csdeconv import auto_response
+from dipy.segment.mask import applymask
+
+from scilpy.utils.bvec_bval_tools import (is_normalized_bvecs, normalize_bvecs,
+                                          check_b0_threshold)
+
+
+def compute_ssft_frf(data, bvals, bvecs, mask=None, mask_wm=None,
```
jchoude: In fact, should be `ssst`: single shell single tissue.

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
...
+    if mask_wm is not None:
+        data = applymask(data, mask_wm)
+    else:
+        logging.warning(
+            'No white matter mask specified! Only mask will be used '
+            '(if it has been supplied). \nBe *VERY* careful about the '
+            'estimation of the fiber response function to ensure no invalid '
+            'voxel was used.')
+
+    # Iteratively trying to fit at least 300 voxels. Lower the FA threshold
```
jchoude: 300 -> `min_nvox`

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
...
+    # when it doesn't work. Fail if the fa threshold is smaller than
+    # the min_threshold.
+    # We use an epsilon since the -= 0.05 might incurs numerical imprecision.
```
jchoude: `incurs` -> `incur`

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
+import logging
+import numpy as np
```
jchoude: Theoretically, there should be a blank line between system and third party imports.

scilpy/reconst/frf.py
```diff
@@ -0,0 +1,126 @@
...
+    gtab = gradient_table(bvals, bvecs, b0_threshold=bvals.min())
+
+    if mask is not None:
+        data = applymask(data, mask)
+
+    if mask_wm is not None:
+        data = applymask(data, mask_wm)
+    else:
+        logging.warning(
+            'No white matter mask specified! Only mask will be used '
```
jchoude: There is a mix of `' '` and `" "` in the strings through the file. Should we uniformize?

scilpy/reconst/frf.py
ppoulin91: Woops, fixed

scilpy/reconst/frf.py
ppoulin91: Fixed

scilpy/reconst/frf.py
ppoulin91: Done

scilpy/reconst/frf.py
ppoulin91: Done

scilpy/reconst/frf.py
ppoulin91: I'll convert everything to double quotes

ppoulin91:
JC's comments were addressed

PR #98, opened 2020-01-30
Fix count_streamlines to support tck file

frheault:
Count streamlines now handle tck file correctly and crash for unknown file format.
Only support trk and tck for lazy loading purpose and crash otherwise.

Uniformize the argparser at the same time (args.tractogram -> args.in_tractogram)

### DIFF
--- a/scripts/scil_count_streamlines.py
+++ b/scripts/scil_count_streamlines.py
@@ -1,5 +1,11 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+
+"""
+Return the number of streamlines in a tractogram. Only support trk and tck in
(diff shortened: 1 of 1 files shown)


arnaudbore:
> Seems good to me. Tested.
> 
> Maybe you could add explanation that it only accepts .tck and .trk? Either in the description or in the help of in_tractogram.
> 
> Idea for later? Maybe add possibility to just ouput the counts, not json-formatted. This way if you loop on a bunch of subject/bundles you can copy the outputs directly from terminal to an excel sheet.

I agree, in the help you should mentioned `tck or trk only`   
Excel is the devil :smiling_imp: , will keep using json :+1:

PR #100, opened 2020-02-01
Added new simple scripts to facilitate header/space check

frheault:
With the new tools in Dipy being quite robust and seeing how complex it is for non-experts to play with header/space, these new simple scripts could help everyone.

First script can print the raw header from trk, tck, nii, nii.gz and mgz (pretty print of the dictionary)
And the second verify if all inputs have the same spatial attribute, support trk, nii, nii.gz and check if the affine, dimensions, voxel_size and voxel_order are the same.

**100th PR !**


scripts/scil_summarize_bundles_similarity_measures.py
```diff
@@ -319,7 +319,7 @@ def main():
                     measure_dict[measure_name])
 
     with open(args.out_json, 'w') as outfile:
-        json.dump(output_measures_dict, outfile)
+        json.dump(output_measures_dict, outfile, indent=1)
```
ppoulin91: Sneaky Francois

scripts/scil_verify_space_attributes_compatibility.py
```diff
@@ -0,0 +1,46 @@
...
+    p = argparse.ArgumentParser(description=__doc__,
+                                formatter_class=argparse.RawTextHelpFormatter)
+
+    p.add_argument('in_files', nargs='+',
+                   help='List of file to compare (trk and nii).')
+
+    return p
+
+
+def main():
```
ppoulin91: Maybe add a success message...?

scripts/scil_print_header.py
```diff
@@ -0,0 +1,57 @@
...
+    assert_inputs_exist(parser, args.in_file)
+
+    _, in_extension = split_name_with_nii(args.in_file)
+
+    if in_extension in ['.tck', '.trk']:
+        header = nib.streamlines.load(args.in_file, lazy_load=True).header
+    elif in_extension in ['.nii', '.nii.gz', '.mgz']:
+        header = dict(nib.load(args.in_file).header)
+    else:
+        parser.error('{} is not a supported extension.', in_extension)
```
AntoineTheb: Missing `format` to string

scripts/scil_summarize_bundles_similarity_measures.py
arnaudbore: Use add_json_args and use it here.

scripts/scil_print_header.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if args.keys:
+        for key in args.keys:
+            if key not in header:
+                parser.error('Key {} is not in the header of {}.'.format(key,
+                             args.in_file))
+            print(" '{}': {}".format(key, header[key]))
+    else:
+        pp = pprint.PrettyPrinter(indent=1)
+        pp.pprint(header)
```
arnaudbore: Why use pprint and not json stuff here ?

scripts/scil_summarize_bundles_similarity_measures.py
frheault: C'est actually un erreur, je vais l'enlever

scripts/scil_print_header.py
EmmaRenauld: I tried and it's actually a lot more complicated than the print.
The header is composed of bytes, not strings. I would have to loop on all keys and values to convert to strings.
I recommand to keep as is.

PR #106, opened 2020-02-14
Add seed saving to tracking

AntoineTheb:
- Can save seeds using Local tracking
- Can save seeds using PFT
- Fixed seed viz tool to properly display seeds


scripts/scil_visualize_seeds.py
```diff
@@ -40,9 +41,11 @@ def main():
         raise ValueError("Invalid input streamline file format " +
                          "(must be trk): {0}".format(args.tractogram_filename))
 
-    # Load files and data
-    trk = TrkFile.load(args.tractogram)
-    tractogram = trk.tractogram
+    # Load files and data. TRKs can have 'same' as reference
+    tractogram = load_tractogram(args.tractogram, 'same')
+    # Streamlines are saved in RASMM but seeds are saved in VOX (?)
```
ppoulin91: If the streamlines are saved in RASmm, maybe we should also save the seeds in RASmm? Also, if the space is non-iso, this changes the behavior of the script, right?

AntoineTheb: > If the streamlines are saved in RASmm, maybe we should also save the seeds in RASmm?
Ideally, yes. But, dipy tracking is done is voxel space and then the tractogram is saved in RASmm. However, the data_per_streamline (or per_point, for that matter) is not modified for obvious reasons. I'm not sure how that should be handled.
> Also, if the space is non-iso, this changes the behavior of the script, right ?
I don't think so, unless the transformation from RASmm to vox doesn't handle non-iso.

frheault: I think there is great argument for both space, and no matter which one we choose it is going to be weird for some use later.

I think the seed should be saved in vox, since in scilpy we are doing the tracking in vox space.

scripts/scil_compute_local_tracking.py
```diff
@@ -106,6 +106,11 @@ def _build_arg_parser():
         help='Random number generator seed.')
     add_overwrite_arg(out_g)
 
+    out_g.add_argument(
+        '--save_seeds', action='store_true',
+        help='If set, save the seeds of the streamlines alongside the '
+             'streamlines themselves')
```
mdesco: Not sure i like this expression. "alongside the" 
Why not say something like save the seeds in the data_per_point of the trk of the streamlines
or something like that

scripts/scil_compute_pft.py
```diff
@@ -139,6 +139,11 @@ def _build_args_parser():
         help='Random number generator seed.')
     add_overwrite_arg(out_g)
 
+    out_g.add_argument(
+        '--save_seeds', action='store_true',
+        help='If set, save the seeds of the streamlines alongside the '
+             'streamlines themselves')
```
mdesco: Same as above

mdesco:
@AntoineTheb if you re-test the scripts and fix the nibabel deprecation.

PR #103, opened 2020-02-12
Encapsulate scil_extract_dwi_shell.py

ppoulin91:
- Encapsulated `scil_extract_dwi_shell.py` in `utils/bvals_bvecs_tools.py`
  - I was unsure where to put this, maybe this could be moved elsewhere?
- Moved a volume block iterator to `image/utils.py`
- Tested on HCP data

PR #108, opened 2020-02-18
Encapsulating extract_dwi_shell

EmmaRenauld:
Note:

- added if,else to run it without the loop on volumes if no block_size is used.
- changed get_data to get_fdata because former is deprecated.
- allowed optional 'dtype' arg. I don't know if you usually accept it but it would be usefull for us. Default is float64, but we like to save memory with float32.
- I had to change the logging message. Is that ok?

### DIFF
--- a/scilpy/utils/bvec_bval_tools.py
+++ b/scilpy/utils/bvec_bval_tools.py
@@ -120,6 +120,66 @@ def get_shell_indices(bvals, shell, tol=10):
         np.logical_and(bvals < shell + tol, bvals > shell - tol))[0]
 
 
+def volumes_iterator(img, size, dtype=None):
+    """Generator that iterates on gradient volumes of data"""
+
+    nb_volumes = img.shape[-1]
+
+    if size == nb_volumes:
+        yield list(range(nb_volumes)), img.get_data(dtype=dtype)
+    else:
+        for i in range(0, nb_volumes - size, size):
+            logging.info(
+                'Loading volumes {} to {}.'.format(i, i + size - 1))
+            yield list(range(i, i + size)), img.dataobj[..., i:i + size]
+        if i + size < nb_volumes:
+            logging.info(
+                'Loading volumes {} to {}.'.format(i + size, nb_volumes - 1))
+            yield list(range(i + size, nb_volumes)), img.dataobj[..., i + size:]
+
+
+def extract_dwi_shell(dwi_img, bvals, bvecs, bvals_to_extract, tol=10,
+                      block_size=None, dtype=None):
+    """
+    Parameters
+    ----------
+    dwi_img: nib.Nifti1image of the DWI image.
+    bvals: loaded bvals
+    bvecs: loaded bvecs
+    tol:
+    block_size:
+        Note. Was previously always set to img.shape[-1] if None.
+    """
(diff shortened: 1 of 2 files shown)


scilpy/utils/bvec_bval_tools.py
```diff
@@ -120,6 +120,70 @@ def get_shell_indices(bvals, shell, tol=10):
         np.logical_and(bvals < shell + tol, bvals > shell - tol))[0]
 
 
+def volumes_iterator(img, size, dtype=None):
+    """Generator that iterates on gradient volumes of data"""
+
+    nb_volumes = img.shape[-1]
+
+    if size == nb_volumes:
+        yield list(range(nb_volumes)), img.get_data(dtype=dtype)
```
AntoineTheb: `get_data` -> `get_fdata`

Also, this line is redundant with line 169, I would remove either one of them and handle the block size accordingly.

scilpy/utils/bvec_bval_tools.py
```diff
@@ -120,6 +120,70 @@ def get_shell_indices(bvals, shell, tol=10):
...
+            logging.info(
+                'Loading volumes {} to {}.'.format(i, i + size - 1))
+            yield list(range(i, i + size)), img.dataobj[..., i:i + size]
+        if i + size < nb_volumes:
+            logging.info(
+                'Loading volumes {} to {}.'.format(i + size, nb_volumes - 1))
+            yield list(range(i + size, nb_volumes)), img.dataobj[..., i + size:]
+
+
+def extract_dwi_shell(dwi_img, bvals, bvecs, bvals_to_extract, tol=10,
```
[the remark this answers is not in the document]
EmmaRenauld: Consistent with what?

scilpy/utils/bvec_bval_tools.py
EmmaRenauld: I agree that's bad. I didn't want to change too much Philippe's work so I copied and paste. I'll try to change that.

You say "as it was before". Where?

AntoineTheb: > You say "as it was before". Where? 
I just meant that before the split this line wasn't there and the block size was just set to `img.shape[-1]` if it wasn't specified in args

scilpy/utils/bvec_bval_tools.py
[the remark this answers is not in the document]
AntoineTheb: The default in `scripts/scil_extract_dwi_shell.py` for tolerance is 20

AntoineTheb:
I just realized this is the same as #103, what's up with that ? @ppoulin91 @EmmaRenauld

EmmaRenauld:
> I just realized this is the same as #103, what's up with that ? @ppoulin91 @EmmaRenauld
> Also there are still pep8 errors

Oups.... Our trello failed ;) Closing this.

AntoineTheb:
@EmmaRenauld Can you change #108 to #103 on the board so it doesn't get reviewed twice? Thanks !

EmmaRenauld:
> @EmmaRenauld Can you change #108 to #103 on the board so it doesn't get reviewed twice? Thanks !

Already done!

PR #111, opened 2020-02-21
Fix wrong argName

frheault:


EmmaRenauld:
I had noticed this and included it sneakily in my PR 109. I can remove it from my PR if you wish

frheault:
I think this will be merge very quickly, it is likely safer to fix this problem quickly.
And you will have a conflict when you will have to rebase ;)

PR #113, opened 2020-02-25
Compute sh from raw signal

EmmaRenauld:
Computing SH directly from raw DWI signal.
On bitbucket, was scil_fit_dwi_to_sh, using reconst.spherical_signal_to_sh_fit.
Simplified (no need for a class. One simple function).
Also, now uses Dipy's sf_to_sh.


EmmaRenauld:
To discuss. Results are not scaled. Ex, with Fibercup, mrview, I had to scale to 0.005 for a nice view.

Should we divide by something? Max(sh)? median(b0)? mean(raw_signal)? etc.

scilpy/reconst/raw_signal.py
```diff
@@ -0,0 +1,117 @@
...
+    # Checks
+    if not is_normalized_bvecs(bvecs):
+        logging.warning("Your b-vectors do not seem normalized...")
+        bvecs = normalize_bvecs(bvecs)
+    check_b0_threshold(force_b0_threshold, bvals.min())
+
+    # Ensure that this is on a single shell.
+    shell_values, _ = identify_shells(bvals)
+    shell_values.sort()
+    if shell_values.shape[0] != 2 or shell_values[0] > 20.:
```
ppoulin91: `shell_values[0] > 20.` We should use the threshold in check_b0_threshold, or use `bvals.min()` if --force_b0_threshold was given

scilpy/reconst/raw_signal.py
```diff
@@ -0,0 +1,117 @@
...
+    erroneous_voxels = np.any(dwi_weights > b0, axis=3)
+    nb_erroneous_voxels = np.sum(erroneous_voxels)
+    if nb_erroneous_voxels != 0:
+        logging.info("# of voxels where `dwi_signal > b0` in any direction: "
+                     "{}".format(nb_erroneous_voxels))
+        dwi_weights = np.minimum(dwi_weights, b0)
+
+    # Compute attenuation
+    dwi_attenuation = dwi_weights / b0
+    dwi_attenuation[np.logical_not(np.isfinite(dwi_attenuation))] = 0.
```
ppoulin91: We could add a small comment that this is done just in case we had some b-values==0

scilpy/utils/bvec_bval_tools.py
```diff
@@ -199,3 +199,26 @@ def mrtrix2fsl(mrtrix_filename, fsl_bval_filename=None,
                             filename_bval=fsl_bval_filename,
                             filename_bvec=fsl_bvec_filename,
                             verbose=1)
+
+
+def identify_shells(bvals, threshold=40.0):
+    """See DIPY's PR, will be added."""
```
ppoulin91: We should have a real docstring until the DIPY version is merged, to make sure we understand what is going on (and in case it isn't actually merged in DIPY)

scilpy/utils/bvec_bval_tools.py
```diff
@@ -199,3 +199,26 @@ def mrtrix2fsl(mrtrix_filename, fsl_bval_filename=None,
...
+    if not len(bvals):
```
ppoulin91: Why not `if len(bvals) == 0:`or just `if bvals:` ? I find it more confusing this way

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Script to compute the SH coefficient directly on the raw DWI signal.
+"""
+
+from __future__ import division, print_function
```
ppoulin91: I don't think we need future imports anymore

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,79 @@
...
+import argparse
+
+from dipy.core.gradients import gradient_table
+from dipy.io.gradients import read_bvals_bvecs
+import nibabel as nib
+import numpy as np
+
+from scilpy.io.utils import (add_overwrite_arg, add_sh_basis_args,
```
ppoulin91: check imports alphabetical ordering, add_force_b0 should be first

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,79 @@
...
+                             assert_inputs_exist, assert_outputs_exist,
+                             add_force_b0_arg)
+from scilpy.reconst.raw_signal import compute_sh_coefficients
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('dwi', metavar='dwi',
```
ppoulin91: metavar is already the same by default, no need to specify

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,79 @@
...
+                   help='Path of the dwi volume.')
+    p.add_argument('bvals',
+                   help='Path of the bvals file, in FSL format.')
+    p.add_argument('bvecs',
+                   help='Path of the bvecs file, in FSL format.')
+    p.add_argument('output', metavar='output',
```
ppoulin91: Same metavar comment

scilpy/reconst/raw_signal.py
EmmaRenauld: changed

scilpy/reconst/raw_signal.py
```diff
@@ -61,7 +61,11 @@ def compute_sh_coefficients(dwi, gradient_table, sh_order=8,
     # Ensure that this is on a single shell.
     shell_values, _ = identify_shells(bvals)
     shell_values.sort()
-    if shell_values.shape[0] != 2 or shell_values[0] > 20.:
+    if force_b0_threshold:
```
ppoulin91: It's actually the opposite (I know it's counter-intuitive...): if `force_b0_threshold` is given, we should use `bvals.min()` as the threshold (see function `check_b0_threshold`: https://github.com/scilus/scilpy/blob/master/scilpy/utils/bvec_bval_tools.py#L61)

scilpy/reconst/raw_signal.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
+
+import logging
+
+import numpy as np
+
```
frheault: Remove line

scilpy/reconst/raw_signal.py
```diff
@@ -0,0 +1,126 @@
...
+from dipy.core.sphere import Sphere
+from dipy.reconst.shm import sf_to_sh
+
+from scilpy.utils.bvec_bval_tools import (check_b0_threshold,
+                                          is_normalized_bvecs, normalize_bvecs,
```
frheault: identify_shells comes before normalize_bvecs, not sure for the last one since it is a variable

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,78 @@
...
+    p.add_argument('dwi',
+                   help='Path of the dwi volume.')
+    p.add_argument('bvals',
+                   help='Path of the bvals file, in FSL format.')
+    p.add_argument('bvecs',
+                   help='Path of the bvecs file, in FSL format.')
+    p.add_argument('output',
+                   help='Name of the output SH file to save.')
+
+    p.add_argument('--sh_order', type=int, default=8,
```
frheault: Use [%(default)s] in the help to indicate the value

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,78 @@
...
+                   help='SH order to fit (int).')
+    add_sh_basis_args(p)
+    p.add_argument('--smooth', type=float, default=0.006,
```
frheault: [%(default)s]

scripts/scil_compute_sh_from_signal.py
```diff
@@ -0,0 +1,78 @@
...
+    assert_inputs_exist(parser, [args.dwi, args.bvals, args.bvecs])
+    assert_outputs_exist(parser, args, args.output)
+
+    vol = nib.load(args.dwi)
+    dwi = vol.get_fdata()
+
+    bvals, bvecs = read_bvals_bvecs(args.bvals, args.bvecs)
+    gtab = gradient_table(args.bvals, args.bvecs, b0_threshold=bvals.min())
+
+    if args.mask is not None:
```
frheault:
```
mask = None
if args.mask:
   mask = nib.load(args.mask).get_fdata().astype(np.bool)
```
would be shorter

scilpy/reconst/raw_signal.py
EmmaRenauld: I changed the identify, but I couldn't find the info if variables are treated differently. Keeping as is.

scilpy/utils/bvec_bval_tools.py
EmmaRenauld: Is that clear?

scilpy/reconst/raw_signal.py
```diff
@@ -0,0 +1,125 @@
...
+
+from scilpy.utils.bvec_bval_tools import (check_b0_threshold, identify_shells,
+                                          is_normalized_bvecs, normalize_bvecs,
+                                          DEFAULT_B0_THRESHOLD)
+
+
+def compute_sh_coefficients(dwi, gradient_table, sh_order=8,
+                            basis_type='descoteaux07', smooth=0.006,
+                            use_attenuation=False, force_b0_threshold=False,
+                            mask=None, sphere=None):
```
jchoude: Why is there a `sphere` argument if the calling script cannot change it? Is it because other scripts use this internally?

EmmaRenauld: Yes, our things in machine learning.

jchoude: Ok

jchoude:
For the class: you can leave it like that for this PR. We would really like for you to create a PR for the class in Dipy, and then use the class here when merged.

EmmaRenauld:
> For the class: you can leave it like that for this PR. We would really like for you to create a PR for the class in Dipy, and then use the class here when merged.

Ok. I will put this on my to do list.

PR #115, opened 2020-02-26
Add more outputs to filter tractogram

arnaudbore:
This PR will help speedup pipelines/scripts that use a lot of scil_filter_tractogram.py to extract streamline count.

@mdesco can you check if the help (section display_count) is clear?


mdesco:
Did you manage to deal with filtering sequence order? 
You can merge

arnaudbore:
There is no order, except if you use filltering_list option. This is something I tried to explained when you read the help.

PR #119, opened 2020-03-03
Update sft filter tractogram

frheault:
Switch main function of scil_filter_tractogram.py to SFT (required dipy 1.1)
Cleaning up the code when needed, full testing by me. Same behaviour as before.

Functions are now safer to use outside of the script


EmmaRenauld:
@frheault I will review this next. Can you solve conflicts?

scilpy/segment/streamlines.py
```diff
@@ -2,56 +2,48 @@
...
+    sft.to_corner()
     # Copy-Paste from Dipy to get indices
+    # Should always be in vox/center space
     target_mask = np.array(target_mask, dtype=np.uint8, copy=True)
-    lin_T, offset = _mapping_to_voxel(affine)
 
-    streamline_index = _streamlines_in_mask(
-        streamlines, target_mask, lin_T, offset)
+    streamline_index = _streamlines_in_mask(list(sft.streamlines), target_mask,
+                                            np.eye(3), [-0.5, -0.5, -0.5])
```
EmmaRenauld: I don't really like the name... First, shouldn't it be plural streamlines_indices? It's not really an index, nor a list of indices. It's 0,1 or 2. streamlines_labels? in_mask_bools? I don't know.

scilpy/segment/streamlines.py
```diff
@@ -2,56 +2,48 @@
...
-from dipy.tracking._utils import _mapping_to_voxel
+from dipy.tracking.streamline import set_number_of_points
 from dipy.tracking.vox2track import _streamlines_in_mask
 from nibabel.affines import apply_affine
 import numpy as np
 
 
-def target_line_based(streamlines, target_mask, affine=None, include=True):
+def target_line_based(sft, target_mask):
+    sft.to_vox()
```
EmmaRenauld: Could you add a small description? Ex, explain why you don't need to return data_per_point and data_per_streamline?
Reason: only used in filter_grid_roi, which does the dealing with data_per_point.

Maybe then this function should be _target_line_based?

scilpy/segment/streamlines.py
```diff
@@ -2,56 +2,48 @@
...
     if filter_type == 'any':
-        _, line_based_indices = target_line_based(
-            streamlines, mask, transfo)
+        _, line_based_indices = target_line_based(sft, mask)
     else:
+        sft.to_vox()
+        sft.to_corner()
+        streamline_vox = sft.streamlines
         # For endpoint filtering, we need to keep 2 informations
         # Could be faster for either end, but the code look cleaner like this
```
EmmaRenauld: informations (with an s) does not exist. Two pieces of information. I know. Damn, english!

the code looks

scilpy/segment/streamlines.py
```diff
@@ -2,56 +2,48 @@
...
+        target_streamlines.append(sft.streamlines[idx])
 
     return target_streamlines, target_indices
 
 
 def filter_grid_roi(sft, mask, filter_type, is_not):
-    streamlines = list(sft.streamlines)
-    transfo, _, _, _ = sft.space_attributes
-
     line_based_indices = []
```
EmmaRenauld: Again, is it possible to add a small description?

scilpy/segment/streamlines.py
```diff
@@ -66,19 +58,22 @@ def filter_grid_roi(sft, mask, filter_type, is_not):
             line_based_indices = np.union1d(line_based_indices_1,
                                             line_based_indices_2)
```
EmmaRenauld: if:
elif:

There is no else? normal?

scilpy/segment/streamlines.py
frheault: It is a function from dipy, the dipy function is pretty crappy I agree. I
I will remove the return of streamlines. We only need the indices, which will also make this script easier to understand.
The indices are needed because of the sequential filtering. I will update the description and change the function name (and the variable)

scilpy/segment/streamlines.py
frheault: I will put an else, but leave a comment with the condition so it is clear which one it is

scripts/scil_filter_tractogram.py
```diff
@@ -158,11 +158,8 @@ def main():
...
-            filtered_streamlines, indexes = filter_grid_roi(
-                sft,
-                mask,
-                filter_mode,
-                is_not)
+            filtered_sft, indexes = filter_grid_roi(sft, mask,
+                                                    filter_mode, is_not)
 
         elif filter_type == 'atlas_roi':
             img = nib.load(filter_arg_1)
```
EmmaRenauld: Would this be more intuitive and less repetitive? (changing filter_arg_1 for filter_arg line 148)

```
        if filter_type == 'drawn_roi' or filter_type=='atlas_roi':
            img = nib.load(filter_arg)
            if not is_header_compatible(img, sft):
                parser.error('Headers from the tractogram and the mask are '
                             'not compatible.')
            if filter_type == 'drawn_roi':
                mask = img.get_data()
            else:
                atlas = img.get_data().astype(np.uint16)
                mask = np.zeros(atlas.shape, dtype=np.uint16)
                mask[atlas == int(filter_arg_2)] = 1
            filtered_sft, indexes = filter_grid_roi(sft, mask,
                                                    filter_mode, is_not)
```

scilpy/segment/streamlines.py
```diff
@@ -2,56 +2,71 @@
...
-    target_streamlines = []
-    for idx in np.where(streamline_index == [0, 1][include])[0]:
-        target_indices.append(idx)
-        target_streamlines.append(streamlines[idx])
-
-    return target_streamlines, target_indices
+    return np.where(streamlines_case == [0, 1][True])[0].tolist()
 
 
 def filter_grid_roi(sft, mask, filter_type, is_not):
```
EmmaRenauld: mask is target_mask in the description

wouldn't "is_exclude" be clearer than "is_not"? (sounds like python's is not). In the main script too.

scilpy/segment/streamlines.py
```diff
@@ -166,27 +227,55 @@ def filter_ellipsoid(sft, ellipsoid_radius, ellipsoid_center,
     elif filter_type == 'either_end':
         selected_by_ellipsoid = np.union1d(line_based_indices_1,
                                            line_based_indices_2)
-    # If the --not option is used, the selection is inverted
-    all_indices = range(len(sft))
+
+    # If the 'exclude' option is used, the selection is inverted
     if is_not:
```
EmmaRenauld: same comment. is_exclude?

EmmaRenauld:
Small comments on old lines that you haven't touched so I can't comment directly:

Script:
line 28: Now supports sequential filtering condition and mix**ED** filtering object.
line 41: _buildArgsParser should be _build_args_parser
line 56: '--bdo', dest='bdo'. dest unnecessary
line 69: --filtering_list. Description not clear for me. i.e. should be example. Can we use atlas_roi in that file?
line 183: sft.dimension

Streamlines:
The code `if both points, np.intersect1d(), else, np.union1d(). If exclude, inverse()` happens 3 times overall.
We could to a _select(line_based_indices_1,  line_based_indices_2, filter_type, is_not) function and call it?

scripts/scil_filter_tractogram.py
```diff
@@ -140,45 +139,30 @@ def main():
     sft = load_tractogram_with_reference(parser, args, args.in_tractogram)
 
     # TractCount before filtering
```
arnaudbore: Streamline count

PR #121, opened 2020-03-05
Update recobundlesx with lower RAM usage

frheault:
- Switch to memmap to reduce memory usage.
- Removed variables that became useless after that switch.
- Updated the docstring and the comments in the code
- Sparse matrix usage converted from dok to lil/csr, (saves 5 minutes at the end)
- Cleaner/clearer/shorter logging message (I listen to comments from collaborators)

Tested on big tractogram, before it was crashing due to a pickle limitation of iterator being bounded by int32 or was limited to 4 processes on a 32Gb machine due to enormous duplication of data.
Now the int32 bug was fixed and 12 processes can be used and need a little over 16Gb of RAM !


scilpy/segment/recobundlesx.py
```diff
@@ -13,13 +13,26 @@
...
+        streamlines.append(streamline.reshape((lengths[i], 3)))
+
+    return streamlines
+
+
 class RecobundlesX(object):
     """
     This class is a 'remastered' version of the Dipy Recobundles class.
-    Made to be used in synch with the voting_scheme
+    Made to be used in synch with the voting_scheme.
```
AntoineTheb: suuuuuuuuper nitpicky but `in sync~u~`

scilpy/segment/recobundlesx.py
```diff
@@ -73,22 +85,22 @@ def recognize(self, model_bundle,
...
         bundle_pruning_thr : int
-            Distance threshold (mm) for the final pruning
+            Distance threshold (mm) for the final pruning.
         slr_transform_type : str
-            Define the transformation for the local SLR
+            Define the transformation for the local SLR.
             [translation, rigid, similarity, scaling]
         identifier : str
-            Identify the current bundle being recognize for the logging
+            Identify the current bundle being recognize for the logging.
```
AntoineTheb: again, nitpicky stuff but "being recognize**d**"

scilpy/segment/recobundlesx.py
```diff
@@ -231,69 +229,65 @@ def _register_neighb_to_model(self, slr_num_thread=1,
...
+
+                slr = StreamlineLinearRegistration(metric=metric,
+                                                   x0=init_transfo_dof,
+                                                   bounds=bounds_dof[:9],
+                                                   num_threads=slr_num_thread)
+                slm = slr.optimize(static, moving)
+        self.model_centroids = transform_streamlines(self.model_centroids,
+                                                     np.linalg.inv(slm.matrix))
+
+    def _prune_what_not_in_model(self,
```
AntoineTheb: Maybe `prune_outside_model` ? or `prune_far_from_model` ? The name `_prune_what_not_in_model` feels weird

scilpy/segment/recobundlesx.py
```diff
@@ -135,66 +144,55 @@ def _cluster_model_bundle(self, model, model_clust_thr, identifier=None):
...
         centroid_matrix[centroid_matrix >
                         neighbors_reduction_thr] = np.inf
 
         mins = np.min(centroid_matrix, axis=0)
-        close_clusters_indices = list(np.where(mins != np.inf)[0])
-        if len(close_clusters_indices) < 1:
+        close_clusters_indices = np.array(np.where(mins != np.inf)[0],
+                                          dtype=np.int32)
+
+        if not len(close_clusters_indices):
```
AntoineTheb: You could remove lines 160, 161 and then at line 173
```
return len(close_clusters_indices) > 0
```
?

Would be a more meaningful return and I think the rest of the code would handle the situation well, loops just wouldn't be branched into

scilpy/segment/voting_scheme.py
```diff
@@ -56,7 +96,7 @@ def __init__(self, config, atlas_directory, transformation,
 
     def _init_bundles_tag(self):
         """
-        Using all bundles in the configuration file and the input models
+        Using all bundles in the configuration file and the input models.
```
AntoineTheb: If I understand the comment correctly, I don't think there should be a period there

scilpy/segment/voting_scheme.py
```diff
@@ -117,74 +155,57 @@ def _load_bundles_dictionary(self, bundles_filepath):
...
         Will find the maximum values of a specific row (bundle_id), make
         sure they are the maximum values across bundles (argmax) and above the
         min_vote threshold. Return the indices respecting all three conditions.
-        :param bundle_id, int, indices of the bundles in the dok_matrix
-        :param min_vote, int, minimum value for considering (voting)
-        :param streamlines_wise_vote, dok_matrix, streamlines-wise
-            sparse matrix for voting
-        :param bundles_wise_vote, dok_matrix, bundles-wise
-            sparse matrix for voting
+        :param bundle_id, int, indices of the bundles in the lil_matrix.
```
AntoineTheb: This is more a question on scilpy than a comment on the review, but are we using two different docstring types for public methods/classes and private methods ?

AntoineTheb:
@frheault could you send me (or point to me to) some data so I could test this ?

jchoude:
I don't know how you want this to be managed, but either the syntax of the json would need to be explained somewhere, or a reference models / config dataset should be made available.

scilpy/segment/voting_scheme.py
frheault: This is laziness on my part, but I felt like a full-blown giga docstring would be useless for some functions. So a smaller, simpler version felt appropriate at the time (it was WIP)

I will do it properly in a different PR

PR #141, opened 2020-03-11
Transfert binary measures scripts

frheault:
Follow up from Tractostorm, this is related to PR #67 but for binary measures.

Here some testing data
https://drive.google.com/open?id=1AGnBd5fQIJYGB84LIw1rGDRxaWNqn5QV


AntoineTheb:
Some of the files in the testing data are symbolic links instead of files, can't test for now

scripts/scil_summarize_bundles_binary_measures.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+
+    if args.files_exist:
```
arnaudbore: I don't understand this.

scripts/scil_summarize_bundles_binary_measures.py
```diff
@@ -0,0 +1,241 @@
...
+        assert_inputs_exist(parser, args.in_bundles)
+    assert_outputs_exist(parser, args, [args.out_json])
```
arnaudbore: does not need to be a list here.

scripts/scil_summarize_bundles_binary_measures.py
frheault: That was something for me (special request) because I auto-generate A LOT of commands using that script and the requirement for an existing file was a pain sometimes.

I removed it. Thanks for pointing that out

PR #144, opened 2020-03-11
Fix endpoint map

GuillaumeTh:
FIx endpoint map assert


GuillaumeTh:
An other bug is present. Give me few minutes

PR #146, opened 2020-03-11
Changing arguments in argparser

karanphil:
Making changes to add add_sh_basis_arg, add_verbose_arg and add_overwrite_arg in argparser for every scripts.


frheault:
There was a major refactor of `scil_compute_connectivity.py`, you need to update with upstream/master.

If you are not sure how/what to do, ask me

karanphil:
I updated with upstream/master and changed the scripts that Alex mentionned.

PR #147, opened 2020-03-11
 added a test to check if output folder exist

StongeEtienne:
Added a check for "assert_outputs_exist", it test of the directory for the output file exist.
Since most our script assume that the folder for the output file exist, only few of them test it.
I think the default should be "check_dir_exist=True", however it would change the current behavior if so.

This could be move to annother function if prefered, eg
"assert_outputs_directories_exist(...)"

@GuillaumeTh  @jchoude  @frheault

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -146,7 +146,8 @@ def assert_inputs_exist(parser, required, optional=None):
             check(optional_file)
 
 
-def assert_outputs_exist(parser, args, required, optional=None):
+def assert_outputs_exist(parser, args, required, optional=None,
+                         check_dir_exist=False):
     """
     Assert that all outputs don't exist or that if they exist, -f was used.
     If not, print parser's usage and exit.
@@ -155,11 +156,18 @@ def assert_outputs_exist(parser, args, required, optional=None):
     :param required: string or list of paths
     :param optional: string or list of paths.
                      Each element will be ignored if None
+    :param check_dir_exist: test if output directory exist
     """
     def check(path):
         if os.path.isfile(path) and not args.overwrite:
             parser.error('Output file {} exists. Use -f to force '
                          'overwriting'.format(path))
+
+        if check_dir_exist:
+            path_dir = os.path.dirname(path)
+            if path_dir and not os.path.isdir(path_dir):
+                parser.error('Directory {}  for a given output file '
+                             'does not exists.'.format(path_dir))
 
     if isinstance(required, str):
         required = [required]
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -156,7 +156,9 @@ def assert_outputs_exist(parser, args, required, optional=None,
     :param required: string or list of paths
     :param optional: string or list of paths.
                      Each element will be ignored if None
-    :param check_dir_exists: test if output directory exists
+    :param check_dir_exists: bool
+                             Test if output directory exists
+
     """
     def check(path):
         if os.path.isfile(path) and not args.overwrite:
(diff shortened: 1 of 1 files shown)


scilpy/io/utils.py
```diff
@@ -155,12 +156,19 @@ def assert_outputs_exist(parser, args, required, optional=None):
...
     """
     def check(path):
         if os.path.isfile(path) and not args.overwrite:
             parser.error('Output file {} exists. Use -f to force '
                          'overwriting'.format(path))
 
+        if check_dir_exist:
+            path_dir = os.path.dirname(path)
+            if path_dir and not os.path.isdir(path_dir):
+                parser.error('Directory {}  for a given output file '
```
AntoineTheb: Extra space after {}

scilpy/io/utils.py
```diff
@@ -155,12 +156,19 @@ def assert_outputs_exist(parser, args, required, optional=None):
     :param required: string or list of paths
     :param optional: string or list of paths.
                      Each element will be ignored if None
+    :param check_dir_exist: test if output directory exist
```
AntoineTheb: exists

scilpy/io/utils.py
```diff
@@ -146,7 +146,8 @@ def check(path):
             check(optional_file)
 
 
-def assert_outputs_exist(parser, args, required, optional=None):
+def assert_outputs_exist(parser, args, required, optional=None,
+                         check_dir_exist=False):
```
AntoineTheb: imo it'd be better to set the default for `check_dir_exist` to True
I can't think of a use case where you would not want to make sure the directory exists beforehand

StongeEtienne: I think the only problem is for script that migh create the folder,
However I feel most of them simply use "assert_output_dirs_exist_and_empty(...)"

scilpy/io/utils.py
```diff
@@ -155,12 +156,19 @@ def assert_outputs_exist(parser, args, required, optional=None):
...
     """
     def check(path):
         if os.path.isfile(path) and not args.overwrite:
             parser.error('Output file {} exists. Use -f to force '
                          'overwriting'.format(path))
 
+        if check_dir_exists:
+            path_dir = os.path.dirname(path)
+            if path_dir and not os.path.isdir(path_dir):
+                parser.error('Directory {} for a given output file '
```
frheault: Not sure is needed, but a / could be put after {} since it is a directory.
I actually don't care, it is just in case. Your call

scilpy/io/utils.py
AntoineTheb: Fixed

scilpy/io/utils.py
AntoineTheb: Fixed

scilpy/io/utils.py
AntoineTheb: Discussed offline

scilpy/io/utils.py
```diff
@@ -155,12 +156,19 @@ def assert_outputs_exist(parser, args, required, optional=None):
     :param required: string or list of paths
     :param optional: string or list of paths.
                      Each element will be ignored if None
+    :param check_dir_exists: test if output directory exists
```
GuillaumeTh:
```suggestion
    :param check_dir_exists: bool
                             Test if output directory exists
```

StongeEtienne:
@frheault @GuillaumeTh  @AntoineTheb 
Should we change the default "check_dir_exist=False"
@arnaudbore @jchoude  what do you think on your side ?
is it too risky even tho 3 persons think it should be de default ?

scilpy/io/utils.py
StongeEtienne: done

scilpy/io/utils.py
StongeEtienne: done

StongeEtienne:
added an issues to discuss the change of the default behavior https://github.com/scilus/scilpy/issues/167
For now it's safer to put it to False

PR #160, opened 2020-03-11
Replaced create_header function

fullbat:
Delete scilpy.io.utils.create_header_from_anat and replace by dipy.io.utils.create_tractogram_header in conjunction with dipy.io.utils.get_reference_info

### DIFF
--- a/scripts/scil_summarize_bundles_similarity_measures.py
+++ b/scripts/scil_summarize_bundles_similarity_measures.py
@@ -11,7 +11,7 @@ import multiprocessing
 import os
 import shutil
 
-from dipy.io.stateful_tractogram import Space, StatefulTractogram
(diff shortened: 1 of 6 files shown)


frheault:
In order to be free of older functions that are now in Dipy, `scilpy.io.image.get_reference_info` should be removed. All references to it should be fixed.

scripts/scil_summarize_bundles_similarity_measures.py
```diff
@@ -11,7 +11,7 @@
 import os
 import shutil
 
-from dipy.io.stateful_tractogram import Space, StatefulTractogram
+from dipy.io.stateful_tractogram import StatefulTractogram
```
frheault: Already warn him about not modifying file unrelated to his PR

PR #162, opened 2020-03-11
Revert import ordering

frheault:
The auto-ordering from Fred modified the agreed import convention.
Reverted the 8 commits rather than reverting individually from GitHub interface.

Numerous instance of `from future import ...`, `from past import ...` or `from builtins import ...` would have to be fixed in the future.


GuillaumeTh:
Finally we take the PEP8 official ordering

PR #105, opened 2020-02-12
Change return of remove_loops_and_sharp_turns

AntoineTheb:
`remove_loops_and_sharp_turns` now returns streamlines and their "ids"
This is so data associated with the streamlines can be kept alongside the valid and invalid streamlines. I also added a utility function to filter the data associated with the returned ids.

~This conflicts with #84~

data: https://drive.google.com/file/d/1B7wi6vFpLLXg_Kd9F68U_nHzX3hb4QwV/view?usp=sharing

PR #165, opened 2020-03-12
Apply autopep8

frheault:
Unknown origin, but all operators were missing space.
Don't know how its possible since I have autopep8 running.


GuillaumeTh:
Duplicated with #105

PR #166, opened 2020-03-12
ENH: Added script to search keywords in all of SCILPY scripts

ppoulin91:


scripts/scil_find_script.py
```diff
@@ -0,0 +1,151 @@
...
+    parser = ArgumentParser(description=__doc__,
+                            formatter_class=RawTextHelpFormatter)
+    parser.add_argument('keywords', nargs='+',
+                        help="Search keywords")
+    return parser
+
+
+def main():
+    parser = _build_arg_parser()
+    add_verbose_arg(parser)
```
EmmaRenauld: Should go in _build_arg_parser I think

scripts/scil_find_script.py
```diff
@@ -0,0 +1,151 @@
...
+            print(filename, '"{}"'.format(display_info))
+
+    if not matches:
+        print("No results found!")
+
+
+def _is_matching_keywords(keywords, texts):
+    """Test multiple texts for matching keywords. Returns True only if all
+    keywords are present in any of the texts.
+
```
EmmaRenauld: I'm not sure "is" is the good name.

_find?

scripts/scil_find_script.py
```diff
@@ -0,0 +1,151 @@
...
+def _build_arg_parser():
+    parser = ArgumentParser(description=__doc__,
+                            formatter_class=RawTextHelpFormatter)
+    parser.add_argument('keywords', nargs='+',
+                        help="Search keywords")
+    add_verbose_arg(parser)
+    return parser
+
+
+def main():
```
jhlegarreta: Is there an agreed convention on the order of the methods in the scripts, e.g.  `_build_arg_parse`r and `main` always first, then the rest of methods, or else are the majority of the scripts making these two be last methods?

ppoulin91: I would guess that `build_arg_parser` should come first, but I don't think there is any convention for the rest

jhlegarreta: OK. I guess having one would improve the readability of the scripts, but it should not be a show-stopper.

scripts/scil_find_script.py
ppoulin91: It doesn't return aything, son I don't think '_find' is the right one. What do you think of '_test_matching_keywords'?

ppoulin91:
@jhlegarreta Yes, I would be nice to add more options if necessary in the future

arnaudbore:
Add a test please.

arnaudbore:
Tested: 
```
scil_find_script.py convert
No results found!
```

mdesco:
On my mac, same thing as @arnaudbore . Cannot find anything, no matter what keyword I enter. 

scil_find_script.py quick
No results found!

scil_find_script.py bundle
No results found!

scil_find_script.py track
No results found!

ppoulin91:
I overall made it nicer to look at, and use bold instead of color when printing to support different terminal configurations. We also now detect when a script will not run and fall back to the docstring in that case.

I fixed the script directory, could @arnaudbore or @mdesco test it now?

arnaudbore:
It works ! Now, can you add a test and change all your prints to logging. 
I may be wrong but I feel like the final print is a little bit too much.
You could have logging.info by default that print the different scripts names and when using verbose (i.e logging.debug) it would also show you the docstring.

ppoulin91:
@arnaudbore Ready for a last check!

PR #170, opened 2020-03-13
ENH: Add FRF load/save utils

jhlegarreta:
Add FRF load/save utils to avoid boilerplate code.

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -257,3 +257,42 @@ def read_info_from_mb_bdo(filename):
     radius = np.asarray(radius, dtype=np.float32)
     center = np.asarray(center, dtype=np.float32)
     return geometry, radius, center
+
+
+def load_frf(fname):
+    """Load the Fiber Response Function (FRF) from a text file.
+
+    Parameters
+    ----------
+    fname: str
+        Filename where the values are stored.
+
+    Returns
+    -------
+    frf_response: tuple (1D array-like, float)
+        Fiber Response Function containing the eigenvalues and the mean b0
+        value.
+    """
+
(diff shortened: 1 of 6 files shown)


jhlegarreta:
Does `pep8speaks` interpret correctly embedded `LaTeX` math syntax?

GuillaumeTh:
@jhlegarreta thanks for your PR however after discussion with @arnaudbore and @jchoude this PR is just wrapping of a numpy function. Doing this give us more code in scilpy and so more testing to do. I keep the PR open until your response

jhlegarreta:
Some context about the PR:
- Methods assume that the values are stored in a [specific way](https://github.com/scilus/scilpy/pull/170/files#diff-56c1b7cb00a4894ec222d6ddd3e13d7bR298), and not in any other way. Doing this at a single place allows people to just provide the array, and avoids them to wonder how they should do the stacking.
- It avoids people needing to know how it is written and how it should be read.
- It avoids people needing to check the shape, and writing the code to raise the exception if the check fails.

Until I diged a little bit into the code I did not know who you were writing the FRF. This method exposes the method to the API and the documentation: many of the users of the tool will read the documentation, and if such function does not exist, most probably they would not know how the FRF is saved/read within the context of the tool. 

I am fine if this requires to write a test; I simply did not do it because generally they have not been added. Let me know if writing one is required.

arnaudbore:
I agree with @GuillaumeTh this seems nice but at the same time it doesn't help much and add an extra wrapper that we don't need. I close this PR. Thank you for your code @jhlegarreta. I would encourage you to submit new PRs but before coding and wasting time please talk to us on slack or raise an issue. :+1:

PR #171, opened 2020-03-13
ENH: Remove `future` `dvision` and `print_function` imports

jhlegarreta:
Remove `future` `division` and `print_function` imports, which are not necessary
in Python 3.x.


jhlegarreta:
The PR does not mean to be exhaustive. I just had to have a look at these files for other reasons.

arnaudbore:
Please be exhaustive otherwise it's useless.

frheault:
@arnaudbore Well, I would not advise changing every single file at once unless you will merge it right away. Because if every file get modified and the PR waits there for 2 weeks it is possible a ton of conflict could appear if stuff gets merge in the meantime.

PR #173, opened 2020-03-13
DOC: Use NumPy docstring style

jhlegarreta:
Use NumPy docstring style.


jhlegarreta:
If I'm not mistaken we prefer to document using the NumPy style vs. the Google style.

scilpy/io/utils.py
```diff
@@ -146,12 +151,16 @@ def validate_sh_basis_choice(sh_basis):
...
-    :param optional: string or list of paths.
-                     Each element will be ignored if None
+    """Assert that all inputs exist. If not, print parser's usage and exit.
+
+    Parameters
+    ----------
+    parser: argparse.ArgumentParser object
+        Parser.
+    required: string or list of paths
+        Paths to be checked.
```
GuillaumeTh: Modify the doc please to be more explicit. For example: `Requiered paths to be checked`.

scilpy/io/utils.py
```diff
@@ -146,12 +151,16 @@ def validate_sh_basis_choice(sh_basis):
...
+    optional: string or list of paths
+        Each element will be ignored if None.
```
GuillaumeTh: Modify the doc please to be more explicit. For example: `Optional paths to be checked`.

scilpy/io/utils.py
```diff
@@ -174,11 +183,17 @@ def assert_outputs_exist(parser, args, required, optional=None):
...
-    :param optional: string or list of paths.
-                     Each element will be ignored if None
+
+    Parameters
+    ----------
+    parser: argparse.ArgumentParser object
+        Parser.
+    args: list
+        Argument list.
+    required: string or list of paths
```
GuillaumeTh: Same comment here for required and optional

scilpy/io/utils.py
```diff
@@ -203,9 +218,15 @@ def assert_output_dirs_exist_and_empty(parser, args, *dirs, create_dir=False):
...
     If exists and not empty, and -f used, delete dirs.
-    :param parser: argparse.ArgumentParser object
-    :param args: argparse namespace
-    :param dirs: list of paths
+
+    Parameters
+    ----------
+    parser: argparse.ArgumentParser object
+        Parser.
+    args: list
```
GuillaumeTh: list -> argparse namespace

scilpy/io/utils.py
```diff
@@ -203,9 +218,15 @@ def assert_output_dirs_exist_and_empty(parser, args, *dirs, create_dir=False):
...
+        Argument list.
+    dirs: list
+        Paths to be checked.
```
GuillaumeTh: Requiered directory paths to be checked

GuillaumeTh:
@mdesco can you check english please. I will do a last review after yours.

scilpy/io/utils.py
jhlegarreta: I would then change the variable name to a more meaningful name, since `required` along is not meaningful to me. Wouldn't `required_paths` be more appropriate?

scilpy/io/utils.py
jhlegarreta: Same comment as as above: `required_paths` vs `optional_paths` could be an option, but may be a more elaborate naming would be better.

scilpy/io/utils.py
jhlegarreta: Same as above. I limited myself to changing the docstring syntax and/or adding the type where necessary without changing the variable names. Let me know what you would propose as an accurate naming of the variables.

arnaudbore:
@jhlegarreta are you still working on this ?

jhlegarreta:
> @jhlegarreta are you still working on this ?

I am willing to wrap this up, but there were unanswered questions to Guillaume's comments ([here](https://github.com/scilus/scilpy/pull/173#discussion_r393264845) and [here](https://github.com/scilus/scilpy/pull/173#discussion_r393271869)): I formatted the docstrings to NumPy style, as it is the style that we are sticking to, but I did not modify the existing argument docstrings.

[This comment](https://github.com/scilus/scilpy/pull/173#discussion_r393214573) and [this other one](https://github.com/scilus/scilpy/pull/173#discussion_r393216666) are clear enough and I could address them, but for the ones above you may want to think and propose a better naming or a more consistent one with the rest of the scripts. I will be willing to change them if you come up with the proposal in here.

scilpy/io/utils.py
arnaudbore: We are going to keep these var names. Please update the doc ;)

arnaudbore:
Please update doc as suggested by @GuillaumeTh and fix conflict. I will merge it right after :+1:

scilpy/io/utils.py
jhlegarreta: Done in 4a5df3b.

scilpy/io/utils.py
jhlegarreta: Done in 4a5df3b.

scilpy/io/utils.py
jhlegarreta: Done in 4a5df3b.

scilpy/io/utils.py
jhlegarreta: Done in 4a5df3b.

scilpy/io/utils.py
jhlegarreta: Done in 4a5df3b.

jhlegarreta:
> Please update doc as suggested by @GuillaumeTh and fix conflict. I will merge it right after 👍

Addressed in 4a5df3b and fixed conflicts.

PR #174, opened 2020-03-13
WIP: ENH: Add Sphinx documentation config files

jhlegarreta:
Add Sphinx documentation config files.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = .
+BUILDDIR      = _build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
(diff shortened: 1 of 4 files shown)


jhlegarreta:
This is still a first draft. Does not yet generate the package documentation properly. The index looks like this:

scilpy_package_documentation_index

boiteaclou:
If I remember well, there is some package that can generate the autodoc .rst files automatically. Also, since the documentation is written in Numpy/Google format, you'll need to add sphinx.ext.napoleon to sphinx extensions in conf.py

jhlegarreta:
@boiteaclou thanks for having had a look !! I will investigate the suggested aspects in the following days. Thanks again Alex !

jhlegarreta:
The documentation gets successfully generated (with some rough edges -29 warnings) with
```
cd doc
sphinx-apidoc -f -o . ../scilpy
make html
```

which gives:

scilpy_package_documentation_module_index

and

scilpy_package_module_doc_example

arnaudbore:
@jhlegarreta can you find a way this doc would auto build ? Where would it live ? Very interesting way of checking where documentation is missing.

jhlegarreta:
> @jhlegarreta can you find a way this doc would auto build ?

I guess I could find a way to add a GitHub Actions workflow to build the doc for every PR.

I think it would be best to add that automation in a separate PR, since it can spur a somewhat long debate about the burden it may involve, whether it can be triggered only when the doc changes, etc. 

That is to say, I can remove the `WIP` label to this PR and it would be ready to be merged provided that you want to test it locally and works for you. Further improvements and fixes can come afterwards, 

> Where would it live ?

Best would be that an administrator of the organization creates a generic account in readthedocs to host this. I would not host this anybody's personal readthedocs account (even if that person is an admin). Maybe we can even come up with a script that automates generating and pushing the new documentation to readthedocs before a release.

GuillaumeTh:
I find a way to do it automatically on scil-documentation. Otherwise we can create a doc folder like this in this repository

arnaudbore:
Documentation has been merge using #247.

PR #175, opened 2020-03-14
Add script compute mean afd map from streamlines

arnaudbore:
@mdesco can you test the script and check the docstring and help?
@frheault  can you check the code?


mdesco:
I don't have the same results scilpy_private vs scilpy_public. On my mac.

Here is a subject to test.
https://www.dropbox.com/s/uh7ol7qsvmifexh/test_afd_arnaud.zip?dl=0 

test/afd_tracks.sh

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,132 @@
...
+    Returns
+    -------
+    afd_sum : np.array
+        AFD map (weighted if length_weighting)
+    rd_sum : np.array
+        rdAFD map (weighted if length_weighting)
+    """
+
+    sft.to_vox()
+
```
frheault: Max said he is not getting the same results as before, could be that you are missing `sft.to_corner()`

It all depends on the behaviour of `grid_intersections`
What was the function being used in the old scilpy ?

scripts/scil_compute_mean_afd_map_from_streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+        Crozier, S., Salvado, O., & Connelly, A. (2012).
+        Apparent Fibre Density: a novel measure for the analysis of
+        diffusion-weighted magnetic resonance images. NeuroImage, 59(4), 3976--3994.
+"""
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
+                                formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_tractogram',
```
frheault: Is this script made mostly (only) for bundles and not whole brain?

@mdesco should we phrase the docstring, argument and name of the script so we promote using bundles rather than tractogram?

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,132 @@
...
+def afd_and_rd_sums_along_streamlines(streamlines, fodf, fodf_basis,
+                                      length_weighting):
+    """
+    Compute the mean Apparent Fiber Density (AFD) and mean Radial fODF (radfODF)
+    maps along a bundle.
+
+    Parameters
+    ----------
+    streamlines :
```
frheault: Doc needed, it needs to be streamlines (as ArraySequence) in voxel space, most likely origin at the corner. OR you could pass the sft down and do that in grid intersection.

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,132 @@
...
+
+    fodf : np.ndarray
+        fODF with shape (X, Y, Z, #coeffs).
+        coeffs depend on the sh_order.
```
frheault: coeffs -> #coeffs

scripts/scil_compute_mean_afd_map_from_streamlines.py
arnaudbore: @mdesco I changed the documentation, I'm open to suggestions to rename the script.

mdesco: Thanks @arnaudbore for changing tractogram for bundles. This is good. As for the name of the script, we could change it to 
scil_compute_mean_afd_map_from_bundles.py 

Also, as this is our only, fixel-based script, maybe we should be more explicit:
scil_compute_mean_fixel_afd_map_from_bundles.py
or
scil_compute_mean_fixel_afd_from_bundles.py

I'm good with all three.

scripts/scil_compute_mean_fixel_afd_from_bundles.py
```diff
@@ -0,0 +1,84 @@
...
+a Funk-Radon transform. Hence, radfODF is the fixel-based or HARDI-based
+generalization of the DTI radial diffusivity and AFD
+the generalization of axial diffusivity.
+
+Please use a bundle file rather than a whole tractogram.
+"""
+
+import argparse
+
+import nibabel as nb
```
frheault: In the rest of scilpy, the vast majority is now `nib` (I think)

scripts/scil_compute_mean_fixel_afd_from_bundles.py
```diff
@@ -0,0 +1,84 @@
...
+        diffusion-weighted magnetic resonance images. NeuroImage, 59(4), 3976--3994.
+"""
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
+                                formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_bundle',
+                   help='Path of the bundle file.')
+    p.add_argument('in_fODF', metavar='fodf',
```
frheault: Is the metavar really necessary ? I would remove it
I would also use `in_fodf`

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,135 @@
...
+def afd_map_along_streamlines(sft, fodf, fodf_basis, length_weighting):
+    """
+    Compute the mean Apparent Fiber Density (AFD) and mean Radial fODF
+    (radfODF) maps along a bundle.
+
+    Parameters
+    ----------
+    sft : StatefulTractogram
+        StatefulTractogram containing the streamlines needed.
+    fodf : np.ndarray
```
frheault: From the code, this isn't an array. It is a nibabel.image (I think)

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,135 @@
...
+    rd_sum_map : np.array
+        fdAFD map.
+    weight_map : np.array
+        Segment lengths.
+    """
+
+    sft.to_vox()
+    sft.to_corner()
+
+    fodf_data = np.asanyarray(fodf.dataobj)
```
frheault: Why not` fodf.get_fdata() `?

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,135 @@
...
+                                      length_weighting):
+    """
+    Compute the mean Apparent Fiber Density (AFD) and mean Radial fODF (radfODF)
+    maps along a bundle.
+
+    Parameters
+    ----------
+    sft : StatefulTractogram
+        StatefulTractogram containing the streamlines needed.
+    fodf : np.ndarray
```
frheault: Again, not an array

scripts/scil_compute_mean_fixel_afd_from_bundles.py
```diff
@@ -0,0 +1,84 @@
...
+    sft = load_tractogram_with_reference(parser, args, args.in_bundle)
+
+    fodf_img = nb.load(args.in_fODF)
+
+    afd_mean_map, rd_mean_map = afd_map_along_streamlines(sft,
+                                                          fodf_img,
+                                                          args.sh_basis,
+                                                          args.length_weighting)
+
+    nb.Nifti1Image(afd_mean_map.astype('float32'),
```
frheault: Maybe use `np.float32`
(It is more common in scilpy, I think)

scilpy/reconst/afd_along_streamlines.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
```
frheault: I think we removed all shebang (`#!/usr/bin/env python`)
(It is not necessary for module, just for script)

scilpy/reconst/afd_along_streamlines.py
arnaudbore: We had a discussion about this with @GuillaumeTh I think it's safer to get the data that way.

PR #177, opened 2020-03-16
Fix bvecs bvals functions

arnaudbore:
Here is a pretty drastic cleaning of gradients related scripts.

After a discussion with @mdesco we have decided to only support fsl and mrtrix gradients files format.
I also added a function to check the filename for each case (mrtrix and fsl). 
This PR does not clean for the duplicated code with dipy for generation of sampling schemes.

Here are some data to test: [data](https://box.criugm.qc.ca/f/9692a7381c5845e29cd0/?dl=1)


mdesco:
@arnaudbore , we did have a chat and I agree with that you wrote above. However, the fsl standard is to call the files bval/bvec not something.bval/something.bvec... Hence, this PR will break everything we used to do, which I don't like, including tractofow... @jchoude @GuillaumeTh what do you think?

I would vote to maintain support for bval/bvec as well (or not) .bval .bvec

Is the BIDS standard just not supporting bval/bvec?


Here is a bash script and bval/bvec files to show you what works and what does not work.

./cmd_bash_scilpy_public.sh

https://www.dropbox.com/s/s1xnhvqxed5kbb2/bval_bvec_test.zip?dl=0

scilpy/io/utils.py
```diff
@@ -56,6 +55,64 @@ def check_tracts_same_format(parser, filename_list):
...
+        filename_list = [filename_list]
+
+    if gradient_format == 'fsl':
+        if len(filename_list) == 2:
+            filename_1 = filename_list[0]
+            filename_2 = filename_list[1]
+            basename_1, ext_1 = split_name_with_nii(filename_1)
+            basename_2, ext_2 = split_name_with_nii(filename_2)
+
+            if ext_1 == '' or ext_2 == '':
```
frheault: Right there, we could handle both case.
Either files have an extension (`.bval` or .`bvec`), OR they can have an empty extension and have a basename being exactly (`bval` or `bvec`).

Only last question, how can we handle `.bval` .`bvec` as well as `.bvals` .`bvecs` (same thing with basename)
I think a lot people uses bvals and bval interchangeably 

What do you think @mdesco

scripts/scil_visualize_multi_shell.py
```diff
@@ -70,6 +70,15 @@ def main():
...
+        assert_gradients_filenames_valid(parser, args.scheme_file, 'fsl')
+    elif len(args.scheme_file) == 1:
+        assert_gradients_filenames_valid(parser, args.scheme_file, 'mrtrix')
+    else:
+        parser.error('Depending on the gradient format you should have '
+                     'two files for FSL format and one file for MRtrix')
+
+    out_basename = None
     if args.out:
         out_basename, ext = split_name_with_nii(args.out)
```
frheault: Since the output is expected to be a `.png`, I think being robust to `.gz` is useless

scilpy/io/utils.py
```diff
@@ -56,6 +55,64 @@ def check_tracts_same_format(parser, filename_list):
...
+            else:
+                parser.error('fsl gradients filenames must have the same '
+                             'basename.')
+        else:
+            parser.error('You should have two files for fsl format.')
+
+    elif gradient_format == 'mrtrix':
+        if len(filename_list) == 1:
+            curr_filename = filename_list[0]
+            basename, ext = split_name_with_nii(curr_filename)
```
frheault: Same thing related to robustness to `.gz`

scilpy/samplingscheme/save_scheme.py
```diff
@@ -106,65 +44,16 @@ def save_scheme_bvecs_bvals(points, shell_idx, bvals, filename=None,
...
+        output bval filename.
+    filename_bvec: str
+        output bvec filename.
     ------
     """
-    if filename:
-        fullfilename, ext = split_name_with_nii(filename)
-        filename_bval = fullfilename + '.bval'
-        filename_bvec = fullfilename + '.bvec'
+    fullfilename, ext = split_name_with_nii(filename_bval)
```
frheault: fullfilename would actually be basename, that or at least full_filename

scilpy/io/utils.py
frheault: Personally I don't mind switching to an extension based convention. My only question is:
- In tractoflow and most simple organisation we name everything `dwi.nii.gz` `bvals` `bvecs` `rev_b0.nii.gz`
If we have an extension based convention, do we call them `bvals.bval` `bval.bval` or `scheme.bval` ? (or anything else)

arnaudbore:
@mdesco BIDS doesn't not support bvec/bval or even (.)bvals/(.)bvecs.
I can modify the script if the consensus is to accept `*bvec(s)/*bval(s)`

I will respond to @frheault comments later this morning.

GuillaumeTh:
I think we can handle both. .bval/.bvec and bval/bvec with "s" or not too.

arnaudbore:
I still think we should move forward with the BIDS flow but I hear you and  I will let people do want they want but if it's not BIDS I will raise a warning.

mdesco:
I personnally don't care and I will adjust but I would really like @GuillaumeTh and @jchoude to comment on this...

arnaudbore:
After a short discussion with @jchoude , @frheault  and @GuillaumeTh we decided to use only .bvec/.bval  and .b for theses commands (conversion and visualisation)

scripts/scil_visualize_gradients.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=__doc__)
+
+    p.add_argument(
+        'scheme_file', metavar='scheme_file', nargs='+',
+        help='Sampling scheme filename. (only accepts .bvecs and .bvals '
+             'or .b).')
```
mdesco, line 33: This help does not match the docstring above. 
Moreover, it is not clear. It works for a .b file, that's fine.
But for the fsl format, it requires both files .bval and .bvec. In any order, it works. But .bvals/.bvecs fails...

Please clarify.

scripts/scil_generate_sampling_scheme.py
```diff
@@ -117,13 +104,10 @@ def main():
...
     args = parser.parse_args()
 
-    caru = args.caru
-    phil = args.phil
     fsl = args.fsl
-    siemens = args.siemens
     mrtrix = args.mrtrix
 
-    if not (caru or phil or fsl or siemens or mrtrix):
+    if not (fsl or mrtrix):
```
mdesco, line 110: Really not of fan of the outfile specification. Any way we can put the extension instead of having to put --fsl or --mrtix at the bottom?

If too complicated, just say, output prefix without extension (please see --fsl or --mrtrix option below)

scripts/scil_visualize_multi_shell.py
```diff
@@ -27,9 +27,9 @@ def _build_args_parser():
         description=__doc__)
 
     p.add_argument(
-        'scheme_file', metavar='scheme_file',
-        help='Sampling scheme filename. (only accepts .txt or .caru or '
-             '.bvecs and .bvals or .b or .dir or .dvs)')
+        'scheme_file', metavar='scheme_file', nargs='+',
+        help='Sampling scheme filename. (only accepts .bvecs and .bvals '
```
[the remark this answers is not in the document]
arnaudbore: This message is not clear I will fix this. Either you put both .bvec and .bval or .b but using only new.bvec won't work.

scripts/scil_generate_sampling_scheme.py
arnaudbore: I decided to leave it that way so we make sure the extension(s) are correct and be able to output mrtrix and fsl formats at the same time.

scripts/scil_visualize_gradients.py
arnaudbore: Done

scripts/scil_visualize_gradients.py
```diff
@@ -0,0 +1,144 @@
...
+
+    shell_idx[shell_idx != -1] -= 1
+
+    sym = args.enable_sym
+    sph = args.enable_sph
+    same = args.same_color
+
+    ms = build_ms_from_shell_idx(points, shell_idx)
+    if proj:
+        plot_proj_shell(ms, use_sym=sym, use_sphere=sph, same_color=same,
```
frheault, line 134: Not sure if possible. But do you think it is possible to cast the shell bval as an int before using it in the output filename?
I do not like having `lol_shell_2000.0.png` if it is not necessary.

Anyway, the centroid approach is matching the similar bval in the first place so no conflict is possible.
I think simply casting it as an int would work 99.99% of the case.

scripts/scil_generate_sampling_scheme.py
```diff
@@ -151,6 +146,11 @@ def main():
     # Compute bval list
     if args.bvals is not None:
         bvals = args.bvals
+        unique_bvals = np.unique(bvals)
+        if len(unique_bvals) != S:
+            parser.error('You have provided {} shells '.format(S) +
+                         'but only {} unique bvals.'.format(len(unique_bvals)))
```
frheault, line 152: I would remove the word only, since you can give more bval than bvec (and then the sentence make less sense)

arnaudbore:
> Last comment, why do we generate a sampling_scheme but visualise gradient ?
> We flip or convert gradient, so should we rename all scil_generate_sampling_scheme or scil_generate_sampling_gradient ? @mdesco
> (Just asking why the two terminologies)

I agree it doesn't make sense here. @mdesco what is your suggestion?

Could be:
```
scil_generate_sampling_gradient.py
scil_generate_gradient.py
```

or something else...

mdesco:
I like scil_generate_gradient_sampling.py.

PR #186, opened 2020-03-26
Normalize connectivity scripts

frheault:
For Sami's project, we needed a script to normalize connectivity matrices.
Some of the options could be achieved through multiple calls to `scil_connectivity_math.py`, but not all (i.e `parcel_volume` and `parcel_surface`)

Other scripts have been modified to handle the usage of logarithm correctly (Around 0) and to make it easier to load/save connectivity matrices. Now that there are more 3-4 scripts related to connectivity, some refactoring was required to avoid duplicated code.

Minor modification to scil_compute_connectivity.py and scil_visualize_connectivity.py (scripts typically being called before and after so they work well together for Sami's project)

These scripts will likely be modified in the future, right now it is being designed with only two projects in mind as an example use case (retroactively Jasmeen, currently Sami)

Data for testing:
https://drive.google.com/open?id=1kdAZlTmz667XdgKkYysQmksH_HZPSGPY

(Sorry about `scripts/scil_compute_connectivity.py`, pep8 was messed up (by me) in a previous PR and now autopep8 added back all the space between operators, 99% of diff in that file are just that. The 1% was to switch the similarity measure from `w_dice` to `bundle_adjency_voxel`)

### DIFF
--- a/scilpy/image/operations.py
+++ b/scilpy/image/operations.py
@@ -9,7 +9,6 @@ to apply simple operations on nibabel images or numpy arrays.
 from collections import OrderedDict
 from copy import copy
 import logging
-import sys
 
 import numpy as np
 from scipy.ndimage.filters import gaussian_filter
@@ -19,7 +18,7 @@ from scipy.ndimage.morphology import (binary_closing, binary_dilation,
 from scilpy.utils.util import is_float
 
 
-EPSILON = sys.float_info.epsilon
+EPSILON = np.finfo(float).eps
 
 
 def get_array_ops():
--- a/scilpy/image/operations.py
+++ b/scilpy/image/operations.py
@@ -9,6 +9,7 @@ to apply simple operations on nibabel images or numpy arrays.
 from collections import OrderedDict
 from copy import copy
 import logging
+import sys
 
 import numpy as np
 from scipy.ndimage.filters import gaussian_filter
@@ -16,6 +17,9 @@ from scipy.ndimage.morphology import (binary_closing, binary_dilation,
                                       binary_erosion, binary_opening)
 
 from scilpy.utils.util import is_float
+
+
+EPSILON = sys.float_info.epsilon
 
 
 def get_array_ops():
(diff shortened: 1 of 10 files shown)


scilpy/image/operations.py
```diff
@@ -18,6 +19,9 @@
 from scilpy.utils.util import is_float
 
 
+EPSILON = sys.float_info.epsilon
```
arnaudbore: EPSILON = np.finfo(float).eps

scilpy/io/utils.py
```diff
@@ -265,3 +266,25 @@ def read_info_from_mb_bdo(filename):
...
+        data = np.loadtxt(filepath)
+    elif ext == '.npy':
+        data = np.load(filepath)
+    else:
+        raise ValueError('Extension {} is not supported'.format(ext))
+
+    return data
+
+
+def save_matrix_in_any_format(filepath, output_data):
```
arnaudbore: You can save without extension but then you can't read it ?

scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -284,6 +282,35 @@ def compute_dice_voxel(density_1, density_2):
...
+def compute_correlation(density_1, density_2):
+    """
+    Compute the overlap (dice coefficient) between two density maps (or binary).
+    Correlation being less robust to extreme case (no overlap, identical array),
+    a lot of check a needed to prevent NaN.
+    Parameters
+    ----------
+    density_1: ndarray
+        Density (or binary) map computed from the first bundle
+    density_1: ndarray of ndarray
```
arnaudbore: density_2

scilpy/io/utils.py
frheault: Well for saving I put a default type, but for loading I didn't want to do try/catch and guessing the kind of file format

scripts/scil_normalize_connectivity.py
```diff
@@ -0,0 +1,195 @@
...
+Normalize a connectivity matrix coming from scil_decompose_connectivity.py.
+3 categories of normalization are available:
+-- Edge attributes
+ - length: Multiply each edge by the average bundle length.
+   Compensate for far away connection when using INT seeding.
+   Cannot be used with inverse_length.
+
+ - inverse_length: Divide each edge by the average bundle length.
+   Compensate for big connection when using WM seeding.
+   Cannot be used with length.
```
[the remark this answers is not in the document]
frheault: Yes, they are left at 0 (untouched)

scripts/scil_normalize_connectivity.py
```diff
@@ -0,0 +1,195 @@
...
+                        help='Length matrix use for edge-wise multiplication.')
+    length.add_argument('--inverse_length', metavar='LENGTH_MATRIX',
+                        help='Length matrix use for edge-wise division.')
+    edge_p.add_argument('--bundle_volume', metavar='VOLUME_MATRIX',
+                        help='Volume matrix use for edge-wise division.')
+
+    vol = edge_p.add_mutually_exclusive_group()
+    vol.add_argument('--parcel_volume', nargs=2,
+                     metavar=('ATLAS', 'LABELS_LIST'),
+                     help='Atlas and labels list for edge-wise division.')
```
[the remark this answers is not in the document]
frheault: In the PR fixing connectivity warning, I added
```
p.add_argument('--out_labels_list', metavar='OUT_FILE',
                   help='Save the labels list as text file.\n'
                        'Needed for scil_compute_connectivity.py and others.')
```
to scil_decompose_connectivity.py

When the script related to connectivity are more stable. I will add the txt file to the data (with the LUT) in scilpy/data/ and add this information to all scripts that needs it.

scilpy/io/utils.py
arnaudbore: default type but without extension so it won't be read with load_matrix_in_any_format.

scripts/scil_normalize_connectivity.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
+        description=__doc__,)
+
+    p.add_argument('in_matrix',
+                   help='Input connectivity matrix. This is typically a '
+                        'streamline_count matrix.')
```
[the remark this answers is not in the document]
arnaudbore: solved but not solved ?? same for a lot of them...

scilpy/io/utils.py
frheault: Ahhh, 1000 times sorry. I though I was adding the extension in the else...

PR #187, opened 2020-03-27
Summarize bundles measures

frheault:
Transfert of the last script of the trio, futurize, switch to sft.
Can be tested with any set of bundles or even tractograms (slower)

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -13,13 +13,13 @@ from scilpy.utils.bvec_bval_tools import DEFAULT_B0_THRESHOLD
 
 def link_bundles_and_reference(parser, args, input_tractogram_list):
     """
-    Associate the bundle to their reference (if they require a reference)
+    Associate the bundle to their reference (if they require a reference).
     Parameters
     ----------
     parser: argparse.ArgumentParser object
-        Parser as created by argparse
+        Parser as created by argparse.
     args: argparse namespace
-        Args as created by argparse
+        Args as created by argparse.
     input_tractogram_list: list
         List of tractogram paths.
     Returns
@@ -126,7 +126,25 @@ def add_sh_basis_args(parser, mandatory=False):
                         help=help_msg)
 
 
-def validate_nbr_processes(parser, args, default_nbr_cpus=None):
+def validate_nbr_processes(parser, args, default_nbr_cpu=None):
+    """ Check if the passed number of processes arg is valid.
+    If not valid (0 < nbr_cpu_to_use <= cpu_count), raise parser.error.
+
+    Parameters
+    ----------
+    parser: argparse.ArgumentParser object
+        Parser as created by argparse.
+    args: argparse namespace
+        Args as created by argparse.
+    default_nbr_cpu: int (or None)
+        Number of cpu to use, default is cpu_count (all).
+
+    Results
+    ------
+    nbr_cpu
+        The number of CPU to be used.
+    """
+
     if args.nbr_processes:
         nbr_cpu = args.nbr_processes
     else:
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -300,3 +300,27 @@ def read_info_from_mb_bdo(filename):
     radius = np.asarray(radius, dtype=np.float32)
     center = np.asarray(center, dtype=np.float32)
     return geometry, radius, center
+
+
+def load_matrix_in_any_format(filepath):
+    _, ext = os.path.splitext(filepath)
+    if ext == '.txt':
+        data = np.loadtxt(filepath)
+    elif ext == '.npy':
+        data = np.load(filepath)
+    else:
+        raise ValueError('Extension {} is not supported'.format(ext))
+
+    return data
+
+
+def save_matrix_in_any_format(filepath, output_data):
+    _, ext = os.path.splitext(filepath)
+    if ext == '.txt':
+        np.savetxt(filepath, output_data)
+    elif ext == '.npy':
+        np.save(filepath, output_data)
+    elif ext == '':
+        np.save('{}.npy'.format(filepath), output_data)
+    else:
+        raise ValueError('Extension {} is not supported'.format(ext))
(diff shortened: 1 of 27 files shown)


mdesco:
All the script names are very similar and I get confused and lost.
In fact, I discover what the scripts is doing by looking at the output json. 

Can you please add the list of measures being computed in the --help?

Also, can you think of better names of script? More explicit to the actual measures computed? (maybe too hard)...

frheault:
@mdesco I added the measures being computed to the description.
And the name are similar because they work in a very similar way (same input, similar output, even the code was made as similar as possible). Any suggestion on a single keyword to add so the name is more explicit?

As for the --keep_tmp, only the similarity measures are combinatorial (NxN) in nature it is needed only for that one. Binary measures are computed only against a gold standard (Nx1) et the normal measurements are for one bundle at the time (N). Since it makes the scripts super complex and it is not needed in terms of computation we decided (last year in bitbucket) that it was better not to have this feature unless it is necessary.

scilpy/io/utils.py
```diff
@@ -125,6 +126,21 @@ def add_sh_basis_args(parser, mandatory=False):
                         help=help_msg)
 
 
+def validate_nbr_processes(parser, args, default_nbr_cpus=None):
```
arnaudbore, line 129: doc

scripts/scil_summarize_bundles_measures.py
```diff
@@ -0,0 +1,119 @@
...
+        if measure_dict is not None:
+            for measure_name in measure_dict.keys():
+                # Create an empty list first
+                if measure_name not in output_measures_dict:
+                    output_measures_dict[measure_name] = []
+                output_measures_dict[measure_name].append(
+                    measure_dict[measure_name])
+
+    with open(args.out_json, 'w') as outfile:
+        json.dump(output_measures_dict, outfile, indent=1)
```
arnaudbore, line 115: could be usefull here to use add_json_args and take these options here.

scripts/scil_summarize_bundles_measures.py
```diff
@@ -0,0 +1,119 @@
...
+from scilpy.tractanalysis.reproducibility_measures import get_endpoints_density_map
+from scilpy.tractanalysis.streamlines_metrics import compute_tract_counts_map
+
+
+def _build_args_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_bundles', nargs='+',
+                   help='Path of the input bundles')
```
arnaudbore, line 40: .

scripts/scil_summarize_bundles_measures.py
```diff
@@ -0,0 +1,119 @@
...
+    p.add_argument('out_json',
+                   help='Path of the output file')
```
arnaudbore, line 42: .

scripts/scil_summarize_bundles_measures.py
```diff
@@ -0,0 +1,119 @@
...
+                             assert_inputs_exist,
+                             assert_outputs_exist,
+                             link_bundles_and_reference,
+                             validate_nbr_processes)
+
+from scilpy.tractanalysis.reproducibility_measures import get_endpoints_density_map
+from scilpy.tractanalysis.streamlines_metrics import compute_tract_counts_map
+
+
+def _build_args_parser():
```
arnaudbore, line 35: args -> arg

scripts/scil_summarize_bundles_measures.py
frheault: These scripts are already pretty complex, are you sure adding two extra options, that likely no one will play with, would be useful?

arnaudbore: it will add just 2 lines in your code. I don't think it will make the scripts more complex.

frheault: I mean only the help,. but sure I will add it

mdesco:
Thanks @frheault for the corrections in the docstring. It helps a lot. 

So, for the first script, my thoughts:
- scil_summarize_bundles_binary_measures.py computes measures to compare bundles. 
i.e.    sensitivity_streamlines, specificity_streamlines, precision_streamlines,
accuracy_streamlines, dice_streamlines, kappa_streamlines, youden_streamlines

You are missing the "voxel" equivalent, right? sensitivity_streamlines/voxels, etc...?

For me, this is script does "streamline-or-voxel-based-bundle-comparison". A name similar to that would be better.

mdesco:
scil_summarize_bundles_individual_measures.py computes basic measurements of bundles
i.e. volume, volume_endpoints, streamlines_count, avg_length, std_length,
min_length, max_length, mean_curvature

To me, there are basic morphology / shape / geometry measures. Maybe a name around this
scil_bundle_morphology_measure.py
scil_bundle_shape_measure.py
... would be clearer

mdesco:
scil_summarize_bundles_similarity_measures.py computes pair-wise similarity measures of bundles bundle_adjacency_voxels, dice_voxels, w_dice_voxels, volume_overlap,
volume_overreach, dice_voxels_endpoints, w_dice_voxels_endpoints,
volume_overlap_endpoints, volume_overreach_endpoints, density_correlation,
density_correlation_endpoints


for me, this is a scil_compute_pair-wise_bundle_shape_measures.py or something like this.

frheault:
To be honest I find most name confusing without reading the help.
For example, to me, morphology or shape does not indicate its computing streamlines count or length
I could make the help even clearer and just explain in a paragraph that the first one (the one with length, volume, streamlines_count) are just the intrinsic measure of bundles, like attributes.

I could name the second one scil_summarize_pairwise_agreement_measures.py. Would that be ok
And the last one scil_summarize_binary_classification_measures.py

mdesco:
Yes. This is better. But maybe worth polling the scil team in the slack and maybe ask Kurt for his thoughts. 

Thanks

PR #193, opened 2020-03-30
Connectivity filtering from population

frheault:
Script to facilitate filtering of connectivity matrices.
The same could be achieved through a complex sequence of `scil_connectivity_math.py`.

For example, a simple filtering (Jasmeen style) would be:
```
scil_filter_connectivity.py out_mask.npy
    --greater_than */sc.npy 1 0.90
    --lower_than */sim.npy 2 0.90
    --greater_than */len.npy 40 0.90 -v;
```
This will result in a binary mask where each node with a value of 1 represents
a node with at least 90% of the population having at least 1 streamline,
90% of the population is similar to the average (2mm) and 90% of the
population having at least 40mm of average streamlines length.

Testing data:
https://drive.google.com/file/d/1o51lTV_ek_cib303ANQMsegxDiZD9kwa/view?usp=sharing

### DIFF
--- a/scripts/scil_filter_connectivity.py
+++ b/scripts/scil_filter_connectivity.py
@@ -53,7 +53,6 @@ def _build_arg_parser():
                    help='Lower than condition using the VALUE_THR in '
                         'at least POPULATION_PERC (from MATRICES_LIST).\n'
                         'See description for more details.')
-
     p.add_argument('--greater_than', nargs='*', action='append',
                    help='Greater than condition using the VALUE_THR in '
                         'at least POPULATION_PERC (from MATRICES_LIST).\n'
--- a/scripts/scil_filter_connectivity.py
+++ b/scripts/scil_filter_connectivity.py
@@ -5,7 +5,7 @@
 Script to facilitate filtering of connectivity matrices.
 The same could be achieved through a complex sequence of scil_connectivity_math.py.
 
-Can be used with any connectivity matrices from scil_compute_connectivity.py.
+Can be used with any connectivity matrix from scil_compute_connectivity.py.
 
 For example, a simple filtering (Jasmeen style) would be:
 scil_filter_connectivity.py out_mask.npy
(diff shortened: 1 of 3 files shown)


scripts/scil_filter_connectivity.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Script to facilitate filtering of connectivity matrices.
+The same could be achieved through a complex sequence of scil_connectivity_math.py.
+
+Can be used with any connectivity matrices from scil_compute_connectivity.py.
```
mdesco: matrices -> matrix

scripts/scil_filter_connectivity.py
```diff
@@ -0,0 +1,153 @@
...
+                   help='Lower than condition using the VALUE_THR in '
+                        'at least POPULATION_PERC (from MATRICES_LIST).\n'
+                        'See description for more details.')
+    p.add_argument('--greater_than', nargs='*', action='append',
+                   help='Greater than condition using the VALUE_THR in '
+                        'at least POPULATION_PERC (from MATRICES_LIST).\n'
+                        'See description for more details.')
+
+    p.add_argument('--keep_condition_count', action='store_true',
+                   help='Report the number of condition that pass/fail rather '
```
mdesco: condition(s)

frheault:
@mdesco about the data and example you mentioned earlier. I think we would need to fetch data like Dipy and do an example that can be run. Maybe every time there is a merge these examples could be run and if they crash we have a notification to modify them.

I say that because the problems with separate example is that after a month some will stop working (fixing parameters name or mandatory parameters order). It would act as tutorials as well as very high level 'tests' that do not actually verify output. But since we could run them in sequence the output of each is kind of tested as an input for the next.

I would prefer that to documentation or example since these 'tests' would force us to keep them up to date. They could be run on-demand or after every merge. The only problem with that idea is the size of the data required and the execution time (would be a couple Gb and likely that 2-3 hours to run)

PR #199, opened 2020-04-03
Histogram plot and print filtering filename

frheault:
Maxime requested a script to visualize a histogram from connectivity matrice. 
I added the option in `scil_visualize_connectivity.py` 
@mdesco you can try it with any matrices (.npy), you may have to use `--log`

From the same request, Maxime wanted a script that can output the filename of .trk that successfully passed the filtering from `scil_filter_connectivity.py`. A script was added to go from a binary connectivity matrix (mask most likely) to the filename using the label_lists.txt that is saved by `scil_decompose_connectivity.py`. This is needed to go from matrix coordinate to filename following ROIs naming convention.

Typical use to isolate the _'nice'_ connections from the bad:
```
scil_print_connectivity_filenames.py filtering_mask.npy labels_list.txt pass.txt
for file in $(cat pass.txt); do mv ${SOMEWHERE}/${FILE} ${SOMEWHERE_ELSE}/; done
```


scripts/scil_print_connectivity_filenames.py
```diff
@@ -0,0 +1,64 @@
...
+def _build_arg_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=__doc__)
+
+    p.add_argument('in_matrix',
+                   help='Binary matrix in numpy (.npy) format.\n'
+                        'Typically from scil_filter_connectivity.py')
+    p.add_argument('labels_list',
+                   help='List saved by the decomposition script.')
+    p.add_argument('out_txt',
```
arnaudbore, line 33: I think we should output a json file instead of txt.

arnaudbore:
@mdesco can you test it? The code is okay, waiting for the conversion into json file tha'ts all.

scripts/scil_print_connectivity_filenames.py
frheault: Why? What would be the keys and adding the {} would make the primary use (in bash) fail

frheault: I could use the parameter out_file, mention I support both.
But for the key/val in the json I don't know what to put

arnaudbore: Ok leave it that way. :+1:

scripts/scil_visualize_connectivity.py
```diff
@@ -30,8 +30,7 @@
 
 def _build_arg_parser():
     p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
-                                description='Flip one or more axes of the '
-                                            'encoding scheme matrix.')
+                                description=__doc__)
```
mdesco, line 33: Line 7 above, you probably mean scil_reorder_connectivity.py, right?

scripts/scil_visualize_connectivity.py
```diff
@@ -184,6 +189,22 @@ def main():
...
+        nbr_bins = len(patches)
+        color = plt.cm.get_cmap(args.colormap)(np.linspace(0, 1, nbr_bins))
+        for i in range(0, nbr_bins):
+            patches[i].set_facecolor(color[i])
+
+        if args.show_only:
+            plt.show()
+        else:
+            plt.savefig(args.histogram, dpi=300, bbox_inches='tight')
+
```
mdesco, line 207: Any chance we can add a binning option? 
For example, this call
 scil_visualize_connectivity.py CM/track_count.npy tc.png --show_only --log --histogram tc_hist.png --exclude_zeros 
looked really good.

But this call looked pretty bad. I would have like to reduce the bin size to see better around 0 and 1000.
scil_visualize_connectivity.py CM/track_count.npy tc.png --show_only  --histogram tc_hist.png --exclude_zeros

tc_hist

scripts/scil_print_connectivity_filenames.py
```diff
@@ -0,0 +1,64 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Output the list of filename using the coordinates for a binary connectivity
```
mdesco, line 5: filename-> filenames
for -> from

scripts/scil_print_connectivity_filenames.py
```diff
@@ -0,0 +1,64 @@
...
+matrix. Typically used to move around files that are considered valid after
+the scil_filter_connectivity.py script.
+
+Example:
+scil_print_connectivity_filenames.py filtering_mask.npy labels_list.txt pass.txt
+for file in $(cat pass.txt); do mv ${SOMEWHERE}/${FILE} ${SOMEWHERE_ELSE}/; done
```
mdesco, line 11: Maybe even add a scil_filter_connectivity.py example before.
scil_filter_connectivity.py filtering_mask.npy --greater_than track_count.npy 1000 1  -v -f. # keep connections with more than 1000 streamlines

scripts/scil_print_connectivity_filenames.py
```diff
@@ -0,0 +1,64 @@
...
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=__doc__)
+
+    p.add_argument('in_matrix',
+                   help='Binary matrix in numpy (.npy) format.\n'
+                        'Typically from scil_filter_connectivity.py')
+    p.add_argument('labels_list',
+                   help='List saved by the decomposition script.')
```
mdesco, line 32: Just looked at scil_decompose_connectivity.py -h again. Should the --out_labels_list be mandatory??? I didn't even know it could be saved. (just a question)

frheault: Once generated it is useless to save it everytime and after the end of my many PR this label list (at least for Freesurfer and Brainnetome will be in scilpy/data/)

frheault:
Issue #200 will be solved once this is merged.

PR #202, opened 2020-04-07
Futurize metrics stats in roi

arnaudbore:
Not a WIP anymore,
Please use a mask + fa 
Next to last script for nf-tractometrie :checkered_flag:


scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +212,30 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
 
     plt.close(fig)
     return fig
+
+
+def get_metrics_stats_over_volume(weights, metrics_files):
```
frheault, line 217: @EmmaRenauld What was the nomenclature you used to replace <over_streamlines> since it was not clear ?
The same should be used here

scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +212,30 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
...
+    Returns
+    ---------
+    stats : list
+        list of tuples where the first element of the tuple is the mean
+        of a metric, and the second element is the standard deviation
+
+    """
+
+    return map(lambda metric_file:
+               weighted_mean_stddev(weights,metric_file.get_fdata(np.float64)),
```
frheault, line 240: stddev -> std_dev ?
Also should we used float32?

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -37,27 +36,26 @@ def _build_arg_parser():
                         'supported format.\nCan be a binary mask or a ' +
```
frheault, line 36: Should we use a capital letter at the beginning of each help/sentence?
I think this is the case everywhere

scilpy/utils/metrics_tools.py
arnaudbore: I really don't know it was cast as np.float64 in the private repository.
I don't feel like it will make a huge difference and should be cast as float32...

@jchoude, @GuillaumeTh  ?

arnaudbore: change for a cast as float32 same for L71.

scilpy/utils/metrics_tools.py
EmmaRenauld: I used along to mean point per point.

For "over" meaning the whole bundle, I just removed over. 
get_metrics_stats_over_streamlines became get_bundle_metrics_mean_std and in the description I said "Returns the mean value of each metric for the whole bundle"

EmmaRenauld: so... Get_volume_metrics_stats?

arnaudbore: get_roi_metrics_stats final answer !

scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +211,31 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
 
     plt.close(fig)
     return fig
+
+
+def get_roi_metrics_stats(density_map, metrics_files):
+    """
```
EmmaRenauld, line 217: I don't really mind, but based on François's previous question about trying to have the same nomenclature everywhere:

Everywhere else, it is called _mean_std, not _stats. I think I actually prefer stats but then maybe we should change it everywhere else?

scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +211,31 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
...
+    Returns the mean and standard deviation of each metric, using the
+    density map provided. This can be a binary mask,
+    or contains weighted values between 0 and 1.
```
EmmaRenauld, line 220: english question: using the density map provided or using the provided density map?

This can ... contain weighted values. No s on contain.

scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +211,31 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
...
+
+    Parameters
+    ------------
+    density_map : ndarray
+        3D numpy array containing a density map.
+
```
EmmaRenauld, line 226: One "enter" not necessary.

scilpy/utils/metrics_tools.py
```diff
@@ -212,3 +211,31 @@ def plot_metrics_stats(mean, std, title=None, xlabel=None,
...
+    ---------
+    stats : list
+        list of tuples where the first element of the tuple is the mean
+        of a metric, and the second element is the standard deviation.
+
+    """
+
+    return map(lambda metric_file:
+               weighted_mean_std_dev(density_map,
+                                     metric_file.get_fdata(dtype=np.float32)),
```
EmmaRenauld, line 240: Idem for dev.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Compute the statistics (mean, stddev) of scalar maps, which can represent
```
EmmaRenauld, line 5: idem for dev

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+diffusion metrics, in a ROI.
+
+The mask can either be a binary mask, or a weighting mask. If a weighting mask
+should either contain floats between 0 and 1, or should be normalized with
```
EmmaRenauld, line 9: If it is a weighting mask, it should...

Or at least add a comma? If a weighting mask, should.... But that's computerish, not english.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+from scilpy.utils.filenames import split_name_with_nii
+from scilpy.utils.metrics_tools import get_roi_metrics_stats
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(description=__doc__,
+                                formatter_class=argparse.RawTextHelpFormatter)
+
+    p.add_argument('in_mask',
+                   help='Mask volume file name, formatted in any nibabel ' +
```
EmmaRenauld, line 34: We usually use filename I think. Also I think usually we don't put the '+' at the end.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+                        'supported format.\nCan be a binary mask or a ' +
+                        'weighted mask.')
+
+    g = p.add_mutually_exclusive_group(required=True)
+    g.add_argument('--metrics_dir',
+                   help='Metrics files directory. name of the directory ' +
+                        'containing the metrics files.')
```
EmmaRenauld, line 41: Name with a capital

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+    g.add_argument('--metrics', dest='metrics_file_list', nargs='+',
+                   help='Metrics nifti file name. list of the names of the ' +
+                        'metrics file, in nifti format.')
```
EmmaRenauld, line 44: idem for filename and capital

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if not issubclass(mask_img.get_data_dtype().type, np.floating) and \
+            not args.normalize_weights:
+        parser.error('The mask file must contain floating point numbers.')
+
+    # Can be a weighted image
+    mask_data = mask_img.get_fdata(dtype=np.float32)
+
+    if args.normalize_weights:
+        mask_data /= np.sum(mask_data)
```
EmmaRenauld, line 79: Hum. So send to [0,1], shouldn't you divide by the max, not the sum? Am I mixed up?

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, args.in_mask, [os.listdir(args.metrics_dir),
+                                               args.metrics_file_list])
+
+    # Load mask and validate content depending on flags
+    mask_img = nib.load(args.in_mask)
+
```
EmmaRenauld, line 70: Would it be intelligent to check that's it's a 3D data, not 4D?

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+    if args.normalize_weights:
+        mask_data /= np.sum(mask_data)
+
+    if np.min(mask_data) < 0.0 or np.max(mask_data) > 1.0:
+        parser.error('Mask data should only contain values between 0 and 1. '
+                     'Try --normalize_weights.')
+
+    if args.bin:
+        mask_data[np.where(mask_data > 0.0)] = 1.0
+
```
EmmaRenauld, line 87: Should you add `mask_data[np.where(mask_data < 0.0)] = 0.0`, just to be sure?

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+    # Load all metrics files.
+    if args.metrics_dir:
+        metrics_files = [nib.load(args.metrics_dir + f)
+                         for f in sorted(os.listdir(args.metrics_dir))]
+    elif args.metrics_file_list:
```
EmmaRenauld, line 92: Technically, just else is enough, but ok if it's for readability. But my pycharm doesn't like it in line 96, it thinks metrics_file might not have been set.
`else: # args.metrics_file_list is not None`

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+
+    p.add_argument('--bin', action='store_true',
+                   help='If set, will consider every value of the mask ' +
+                        'higher than 0 to be part of the mask, and set to 1 ' +
+                        '(equivalent weighting for every voxel).')
+
+    p.add_argument('--normalize_weights', action='store_true',
+                   help='If set, the weights will be normalized to the [0,1] '
+                        'range.')
+
```
EmmaRenauld, line 54: bin and normalize_weights should probably be exclusive. Or test lower that user did not use both.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,112 @@
...
+    add_json_args(p)
+
+    return p
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, args.in_mask, [os.listdir(args.metrics_dir),
```
EmmaRenauld, line 65: BUG here. If user used --metrics, then args.metrics_dir is None but this (`os.listdir(None)` still gives the list of dirs in the current directory.

If user used --metrics_dir (I tried with --metrics_dir test_metrics, a folder with one FA inside), then what you send to the optional parameter is [['fa.nii.gz'], None]. So two bugs: 1- the dir name is not attached (should be test_metrics/fa.nii.gz) and 2-it's a list inside a list so it tries to see if the file ['fa.nii.gz'] exists and crashed saying it's a list, not a string.

scilpy/utils/metrics_tools.py
arnaudbore: Good call, for an other PR.

scripts/scil_compute_metrics_stats_in_ROI.py
arnaudbore: You're right it should not be sum (legacy code :weary: ). But `/ max` or `val - min / (max-min)` ? @mdesco

scripts/scil_compute_metrics_stats_in_ROI.py
arnaudbore: check L80 ish...

scripts/scil_compute_metrics_stats_in_ROI.py
arnaudbore: I'm going to keep it.

scripts/scil_compute_metrics_stats_in_ROI.py
mdesco: I believe this is ok but would like to the guru to validate @jchoude.

Normalization, here, should not be taken in the usual way. The weighted mask is a track-count or tdi map normally. And the goal is to weigh the average FA by local TDI, i.e. FA weighs more where there are more tracks... So, / max or val - min / (max-min) is not what we want. I think it should be left as is...

frheault: Here is an example:
```
import numpy as np
value = np.random.rand(10) # Imagine this is the FA (between 0 and 1)
weight = np.ones((10,)) # Imagine this is a streamline count (equal density everywhere)

# Our two possibilities
weight_1 = weight / np.max(weight)
weight_2 = weight / np.sum(weight)

np.average(value) # 0.431
np.average(value*weight_1)  # 0.431
np.average(value*weight_2) # 0.0431
```

Not sure the version #2 with the sum make sense, the more voxel you have the smaller your results get in relation to your original data while option #1 keep the same scaling.
(I personally use the max() version in scil_compute_connectivity())
```
if weighted:
    density = density / np.max(density)
```

@Arnaud, I thought about it and I think you should not do the -min() part because having a weight between 100 and 1000 would end up putting a 0 weight on the lowest part when in fact you really want to give it 10%. Also, simply dividing by the max() is negative value friendly !

arnaudbore: I agree the min-max normalisation is not really what we are looking for here but @mdesco dividing by sum does not seem to make sense to me. I would go for /max.

Waiting for @jchoude validation

EmmaRenauld: > You're right it should not be sum (legacy code 😩 ). But `/ max` or `val - min / (max-min)` ? @mdesco

If you decide to keep your change, careful, you wrote max instead of min `mask_data = mask_data - np.max(mask_data) / (np.max(mask_data) - np.min(max_data)) ` and also you are missing a parenthesis `(val - min) / (max-min)`

scilpy/utils/metrics_tools.py
jchoude: I was casting to f64 in the old version because, as some point, we had issues with metrics that have a small range (AD, MD and such), where the `stddev` was not stable when using f32 for computations.

scilpy/utils/metrics_tools.py
jchoude: Stats is larger than just mean and std, no? Not sure I would change it to "stats", it is quite large and doesn't precisely represent the content of the method.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,121 @@
...
+    # Load mask and validate content depending on flags
+    mask_img = nib.load(args.in_mask)
+
+    if len(mask_img.shape) > 3:
+        logging.error('Mask should be a 3D image.')
+
+    # Can be a weighted image
+    mask_data = mask_img.get_fdata(dtype=np.float32)
+
+    if np.min(mask_data)<0:
```
jchoude, line 84: Spaces between the operators (`(mask_data) < 0`)

scripts/scil_compute_metrics_stats_in_ROI.py
jchoude: Ok so. 

The reasoning of the sum is exactly what @mdesco said. Suppose you have a ROI and want to weigh each voxel differently, but still have a total sum of weights that go to 1. Then, you would need to divide by the sum, not by the max, to ensure that the sum of weights is one.

Now. I rechecked the old PRs, and dug down into my memory. I really remember that, at the time, there was a reason to do this, specifically. I can't seem to find it now, and I don't want to BS anything.

However: @frheault your example doesn't fit with the current code. When we call `np.average`, we specifically use the `weights` argument, which you don't. If I reuse your example with the correct syntax:

```
import numpy as np
value = np.random.rand(10) # Imagine this is the FA (between 0 and 1)
weight = np.ones((10,)) # Imagine this is a streamline count (equal density everywhere)

# Our two possibilities
weight_1 = weight / np.max(weight)
weight_2 = weight / np.sum(weight)

np.average(value) # 0.503
np.average(value, weights=weight_1)  # 0.503
np.average(value, weights=weight_2)  # 0.503
```

Notice the use of `weights` instead of multiplying directly. In your example, you were then missing a division by the sum of the weights.

If you compare the scaled ratios (`weight_1` and `weight_2`), you see that the weights are not the same, but their scaling to each other is the same (eg: `weight_1[0] / weight_1[1]` == `weight_2[0] / weight_2[1]`). This means that using either `sum` or `max` will yield the same (or almost the same, up to something like `10^-8`) end result.

So. It was not a bug before. If people really absolutely prefer using `max`, I'm ok with that for the sake of not having to redo all this discussion in a year or two. However, we shouldn't use min-max normalization, since it zeroes out the minimal values of the array. This is not a large error source, but not preferable.

frheault: So for consistency, I will use that function in my scil_compute_connectivity.py (once merged)
Thanks for the explanation/correction!

jchoude: :+1: and it will be modified to use `max` I guess?

jchoude:
Also, there was a comment about making `normalize` and `bin` mutually exclusive, but there was a good reason not to do so.

Imagine you have TDI map, but you would like to use it as a binary mask. In this case, just using `--bin` wouldn't work because there is a check of the range before the check for binarization. In this case, using `--normalize` and `--bin` at the same time allows one to use TDI maps directly as binary without creating 2 intermediate files.

scripts/scil_compute_metrics_stats_in_ROI.py
arnaudbore: I put max and add a link to this discussion into the code as comment. My guess, in a year or two we will ask our self the same question.

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -0,0 +1,122 @@
...
+
+IMPORTANT: if the mask contains weights (and not 0 and 1 exclusively), the
+standard deviation will also be weighted.
+"""
+
+import argparse
+import glob
+import json
+import logging
+import nibabel as nib
```
jchoude, line 20: So close to having it all :) Nibabel and Numpy are 3rd party, should go in second block.

PR #203, opened 2020-04-07
Futurize tractprofiles

arnaudbore:
Still a WIP but here is the last script for nf-tractometrie :1st_place_medal:


scripts/scil_tractprofiles.py
```diff
@@ -16,11 +16,14 @@
...
-from scilpy.tracking.tools import subsample_streamlines
+from scilpy.io.utils import (assert_inputs_exist,
+                             add_json_args,
+                             add_overwrite_arg,
+                             add_reference_arg)
 from scilpy.utils.filenames import split_name_with_nii
 from scilpy.utils.metrics_tools import get_metrics_profile_over_streamlines
-
+from scilpy.tracking.tools import (resample_streamlines_num_points,
+                                   resample_streamlines_step_size)
```
frheault, line 26: I think this option makes no sense `resample_streamlines_step_size` the tractprofile has a set number of point, all streamlines need to have the same number of point for it to work, right?

GuillaumeTh:
Please fix PEP8. It will be a test for Travis in the sametime

scilpy/utils/metrics_tools.py
```diff
@@ -10,6 +10,58 @@
 from scilpy.utils.filenames import split_name_with_nii
 
 
+def get_metrics_profile_over_streamlines(streamlines, metrics_files):
```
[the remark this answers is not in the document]
arnaudbore: -> get_bundle_metrics_profile

scilpy/utils/metrics_tools.py
```diff
@@ -10,6 +10,58 @@
...
+    """
+    Returns the profile of each metric along each streamline.
+    This is used to create tract profiles.
+
+    Parameters
+    ------------
+    streamlines : sequence
```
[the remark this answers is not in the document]
arnaudbore: Done

scripts/scil_tractprofiles.py
arnaudbore: Agree

scripts/scil_tractprofiles.py
```diff
@@ -88,7 +90,7 @@ def main():
     # Make sure all streamlines go in the same direction. We want to make
     # sure point #1 / 20 of streamline A is matched with point #1 / 20 of
     # streamline B and so on
-    num_streamlines = len(bundle_subsampled)
+    num_streamlines = len(new_sft.streamlines)
```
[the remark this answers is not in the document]
mdesco: There is no reason for the 1st streamline to be a good reference for all others. 

I agree with Francois on this one and this is a nightmare to deal with with collaborators using tractometry whenever bundles don't come from RBX. 

I would even suggest here to be drastic and stop the script if we think streamlines haven't been uniformized before hand (possible to check?)

Ping-me in the slack if we need to discuss this.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Compute tract profiles and their statistics along streamlines.
```
mdesco, line 5: This script is called bundle_profiles. Please replace all tract profiles for bundle profiles in the docstring.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,125 @@
...
+                                description=__doc__)
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file to compute the tract profiles on.')
+    p.add_argument('in_metrics', nargs='+',
+                   help='Nifti metric(s) on which to compute '
+                        'the tract profiles.')
+    p.add_argument('--in_centroid',
+                   help='If provided it will be used to make sure all '
+                        'streamlines go in the same direction. Otherwise, a '
+                        'centroid will be automatically computed.')
```
mdesco, line 48: This script breaks if centroid does not have the right number of --nb_pts_per_streamline.

Please robustify.

Also, even if centroid has the right number of points it breaks right now. See data here:
https://www.dropbox.com/s/xrra0c3084xvlyu/bundle_profile_bug.zip?dl=0

 scil_compute_bundle_profiles.py output/AF_L_m.trk --in_centroid AF_L_centroid_10pts.trk  t1_converted.nii.gz --nb_pts_per_streamline 10
scil_compute_bundle_profiles.py output/AF_L_m.trk --in_centroid AF_L_centroid.trk  t1_converted.nii.gz --nb_pts_per_streamline 10

arnaudbore: Do you want to force resampling of the centroid if `in_centroid and --nb_pts_per_streamline` or raise an error ?

frheault: I personally think they should be mutually exclusive to be really clear (but required). If nothing is provided we computed it with the nbr_of_points, and if we provide the centroids we use them as is.
Anyway if we want to resample centroids we should use `scil_resample_streamlines.py`

I would also be ok with removing that parameter entirely, since we rarely call that script out of the blue, we usually follow steps which include `scil_compute_centroids.py` (like in the tractometry pipeline)

arnaudbore: Everything should be fixed.
@max your centroid has 99 points.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, [args.in_bundle] + args.in_metrics,
+                        optional=args.in_centroid)
+
+    if args.nb_pts_per_streamline <= 1:
+        parser.error('--nb_pts_per_streamline {} needs to be greater than '
+                     '1'.format(args.nb_pts_per_streamline))
+
+    assert_same_resolution(args.in_metrics)
```
frheault, line 75: Could you bring `sft = load_tractogram_with_reference(parser, args, args.in_bundle)` before that assert and include it in the check ?
You can keep `assert_same_resolution` as is, for now, we already talked about changing the name and docstring in a future PR.

Right now you can load a tractogram that do not fit with the metrics and that's problematic.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+    metrics = [nib.load(m) for m in args.in_metrics]
+
+    sft = load_tractogram_with_reference(parser, args, args.in_bundle)
+
+    bundle_name, _ = os.path.splitext(os.path.basename(args.in_bundle))
+    stats = {}
+    if len(sft) == 0:
```
frheault, line 82: We should have an explicit agreement on that.
I know some people like `if len(sft.streamlines) == 0` while I personally use `if len(sft)` and here you are using a mix of both. I don't care which one we choose.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+        stats[bundle_name] = None
+        print(json.dumps(stats, indent=args.indent, sort_keys=args.sort_keys))
+        return
+
+    # Centroid - will be use as reference to reorient each streamline
+    if args.in_centroid:
+        sft_centroid = load_tractogram_with_reference(parser, args,
```
frheault, line 89: To be really robust, calling `is_header_compatible` between both `sft` would be needed.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+        centroid_streamlines = get_streamlines_centroid(sft.streamlines,
+                                                        args.nb_pts_per_streamline)
+        nb_pts_per_streamline = args.nb_pts_per_streamline
+
+    resampled_sft = resample_streamlines_num_points(sft, nb_pts_per_streamline)
+
+    # Make sure all streamlines go in the same direction. We want to make
+    # sure point #1 / args.nb_pts_per_streamline of streamline A is matched
+    # with point #1 / 20 of streamline B and so on
+    num_streamlines = len(resampled_sft.streamlines)
```
frheault, line 103: Related to previous comment `len(sft)` and `len(sft.streamlines)` are the same, but one is a bit clearer

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file to compute the bundle profiles on.')
+    p.add_argument('in_metrics', nargs='+',
+                   help='Nifti metric(s) on which to compute '
+                        'the bundle profiles.')
+
+    g = p.add_mutually_exclusive_group()
+    g.add_argument('--in_centroid',
+                   help='If provided it will be used to make sure all '
+                        'streamlines go in the same direction. \nOtherwise, a '
```
frheault, line 49: Since the two parameters are now mutually exclusive I would remove the text from _Otherwise_ and adapt the help of --nbr_pts_per_streamline accordingly.

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+    return np.mean(norm_l2(x - y))
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=__doc__)
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file to compute the bundle profiles on.')
+    p.add_argument('in_metrics', nargs='+',
+                   help='Nifti metric(s) on which to compute '
```
frheault, line 43: Do we go for the style: _Metric(s) on which to compute the bundle profiles (.nii or .nii.gz)_

scripts/scil_compute_bundle_profiles.py
```diff
@@ -0,0 +1,131 @@
...
+
+def average_euclidean(x, y):
+    return np.mean(norm_l2(x - y))
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter,
+                                description=__doc__)
+    p.add_argument('in_bundle',
+                   help='Fiber bundle file to compute the bundle profiles on.')
```
frheault, line 41: I don't like that sentence, I would prefer something more simple _Input bundle for profiling._

PR #204, opened 2020-04-07
Add get_data for masks and label

arnaudbore:
As discussed with @jchoude, @GuillaumeTh and @frheault here are two new functions to have a better control over the type of masks and labels inputs.

I will add a converted script with these new functions to have an example.

### DIFF
--- a/scilpy/io/image.py
+++ b/scilpy/io/image.py
@@ -3,6 +3,7 @@
 from dipy.io.utils import get_reference_info
 import logging
 import numpy as np
+import os
 
 
 def assert_same_resolution(images):
@@ -45,16 +46,16 @@ def get_data_as_mask(in_img):
     curr_type = in_img.get_data_dtype().type
     basename = os.path.basename(in_img.get_filename())
     if np.issubdtype(curr_type, np.signedinteger) or np.issubdtype(curr_type, np.unsignedinteger) or np.issubdtype(curr_type, np.bool):
-        data = np.asanyarray(in_img.object).astype(np.uint8)
+        data = np.asanyarray(in_img.dataobj).astype(np.uint8)
         unique_vals = np.unique(data)
         if len(unique_vals) == 2:
-            if np.all(unique_vals != np.array[0, 1]):
+            if np.all(unique_vals != np.array([0, 1])):
                 logging.warning('The two unique values in mask were not 0 and'
                                 ' 1. Tha mask has been binarised.')
                 data[data != 0] = 1
         else:
             raise IOError('The image {} contains more than 2 values. '
-                          'It can\t be loaded as mask.'.format(basename))
+                          'It can\'t be loaded as mask.'.format(basename))
 
     else:
         raise IOError('The image {} cannot be loaded as mask because '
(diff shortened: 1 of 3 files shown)


scilpy/io/image.py
```diff
@@ -23,3 +24,68 @@ def assert_same_resolution(images):
         aff_2, shape_2, _, _ = get_reference_info(i)
         if not (shape_1 == shape_2) and (aff_1 == aff_2).any():
             raise Exception("Images are not of the same resolution/affine")
+
+
+def get_data_as_mask(parser, in_img):
```
frheault, line 29: I would not use a parser.error(), but rather a `raise IOError` or something since this function will be used pretty much everywhere in the code and not just in the `main()` of a script.

scilpy/io/image.py
```diff
@@ -23,3 +24,68 @@ def assert_same_resolution(images):
...
+    return data
+
+
+def get_data_as_label(parser, in_img):
+    """
+    Get data as label (force type np.uint16), check data type before casting.
+
+    Parameters
+    ----------
+    parser: argparse.ArgumentParser object
```
frheault, line 72: Same thing

scripts/scil_remove_labels.py
```diff
@@ -63,7 +64,7 @@ def main():
 
     # Save final volume
     nii = nib.Nifti1Image(labels_volume, volume_img.affine, volume_img.header)
```
frheault, line 66: side note header is optional (I would prefer header=volume_img.header). We almost never do that in scilpy (providing the header), we usually simply give the affine. Should we start doing it ?
I am @jchoude has some good/bad experience with both.
I know that some output of freesurfer (and DSI Studio) write fancy stuff in the header, but it is usually more pain than joy to deal with them.

scilpy/io/image.py
```diff
@@ -23,3 +24,68 @@ def assert_same_resolution(images):
...
+        Data (dtype : np.uint8).
+    """
+
+    curr_type = in_img.get_data_dtype().type
+    basename = os.path.basename(in_img.get_filename())
+    if np.issubdtype(curr_type, np.signedinteger) or np.issubdtype(curr_type, np.unsignedinteger) or np.issubdtype(curr_type, np.bool):
+        data = np.asanyarray(in_img.object).astype(np.uint8)
+        unique_vals = np.unique(data)
+        if len(unique_vals) == 2:
+            if np.all(unique_vals != np.array[0, 1]):
```
frheault, line 52: A warning saying that the mask had two valy that were not  0 and 1 and converted to binary could be useful

scilpy/io/image.py
```diff
@@ -23,3 +26,66 @@ def assert_same_resolution(images):
...
+        Data (dtype: np.uint16).
+    """
+
+    curr_type = in_img.get_data_dtype()
+    basename = os.path.basename(in_img.get_filename())
+    if np.issubdtype(curr_type, np.signedinteger) or np.issubdtype(curr_type, np.unsignedinteger):
+        return np.asanyarray(in_img.dataobj).astype(np.uint16)
+    else:
+        raise IOError('The image {} cannot be loaded as label because '
+                     'its format {} is not compatible with a label '
```
frheault, line 90: Is it me (in the browser) or this has a missing space (not aligned)

PR #212, opened 2020-04-16
Mosaic bundles support binary nifti

frheault:
Addition for Kurt's project and useful in general.
This script can now handle (in any resolution) binary nifti file. It can even be mixed with tck/trk !

Exact same behavior as before for default usage. The mosaic was printing the number of streamlines, the text was replaced with _elements_ to handle printing number of voxels for masks and number of streamlines for tractograms.

Two very simple coloring option were added uniform_coloring (e.g [1,1,0] for yellow) and random_coloring that assign a random color for each bundle.
Default is endpoints coloring for bundles and white surface for binary mask.

Example image (mask and streamlines, random coloring)
mosaic_vol_2

mosaic_vol_1

### DIFF
--- a/scripts/scil_visualize_bundles_mosaic.py
+++ b/scripts/scil_visualize_bundles_mosaic.py
@@ -31,19 +31,19 @@ from scilpy.utils.filenames import split_name_with_nii
 def _build_arg_parser():
     p = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
-    p.add_argument('anat_reference',
-                   help='Image used as background (e.g. T1, FA, b0).')
-    p.add_argument('inputs', nargs='+',
-                   help='List of streamline or files supported by nibabel '
-                        'or binary mask file.')
-    p.add_argument('output_name',
+    p.add_argument('in_volume',
+                   help='Volume used as background (e.g. T1, FA, b0).')
+    p.add_argument('in_bundles', nargs='+',
+                   help='List of tractography files supported by nibabel '
+                        'or binary mask files.')
+    p.add_argument('out_image',
                    help='Name of the output image mosaic '
                         '(e.g. mosaic.jpg, mosaic.png).')
 
     p.add_argument('--uniform_coloring', nargs=3,
                    metavar=('R', 'G', 'B'), type=float,
                    help='Assign an uniform color to streamlines (or ROIs).')
-    p.add_argument('--random_coloring', metavar='SEED', type=int, default=0,
+    p.add_argument('--random_coloring', metavar='SEED', type=int,
                    help='Assign a random color to streamlines (or ROIs).')
     p.add_argument('--zoom', type=float, default=1.0,
                    help='Rendering zoom. '
@@ -67,7 +67,7 @@ def _build_arg_parser():
 
 
 def get_font(args):
-    """Returns a ttf font object."""
+    """ Returns a ttf font object. """
     if args.ttf is not None:
         try:
             font = ImageFont.truetype(args.ttf, args.ttf_size)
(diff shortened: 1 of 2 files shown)


scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -10,44 +10,60 @@
...
-                        help='Font size (int) to use for the legends.')
-    parser.add_argument('--opacity_background', type=float, default=0.4,
-                        help='Opacity of background image [0, 1.0]')
-    parser.add_argument('--resolution_of_thumbnails', type=int, default=300,
-                        help='Resolution of thumbnails used in mosaic.')
-
-    add_overwrite_arg(parser)
-    return parser
+    p.add_argument('anat_reference',
+                   help='Image used as background (e.g. T1, FA, b0).')
```
mdesco, line 35: Maybe volume instead of image? (suggestion)

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -10,44 +10,60 @@
...
+    p.add_argument('inputs', nargs='+',
+                   help='List of streamline or files supported by nibabel '
```
mdesco, line 37: is the 'or' supposed to be there?
I would suggest
List of streamline files supported by ...

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -10,44 +10,60 @@
...
 from PIL import ImageFont
 from PIL import ImageDraw
 
-from scilpy.io.utils import (
-    add_overwrite_arg, assert_inputs_exist, assert_outputs_exist)
+from scilpy.io.image import get_data_as_mask
+from scilpy.io.utils import (add_overwrite_arg,
+                             assert_inputs_exist,
+                             assert_outputs_exist)
+from scilpy.utils.filenames import split_name_with_nii
```
mdesco, line 28: While we are working this script can you please add 1-2 examples in the docstring that cover options and summarize your PR comment. Thanks

mdesco:
Quand meme tres cool by the way! Can put wm_mask, interface masks, etc... and looks very good! Could be useful for QA/QC @GuillaumeTh maybe?

frheault:
@mdesco Its in 3D, so interface would look like a brain surface obscuring the inside.
I think it would would only to QA stuff like the different nuclei, something convex that is not empty inside.

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -10,48 +10,64 @@
...
+    p.add_argument('--zoom', type=float, default=1.0,
+                   help='Rendering zoom. '
+                        'A value greater than 1 is a zoom-in,\n'
+                        'a value less than 1 is a zoom-out [%(default)s].')
+
+    p.add_argument('--ttf', default=None,
+                   help='Path of the true type font to use for legends.')
+    p.add_argument('--ttf_size', type=int, default=35,
+                   help='Font size (int) to use for the legends. '
+                        '[%(default)s]')
```
arnaudbore, line 57: move . after default

scripts/scil_visualize_bundles_mosaic.py
```diff
@@ -200,13 +224,28 @@ def main():
...
-            bundle_tractogram_file = nib.streamlines.load(bundle_file)
-            streamlines = bundle_tractogram_file.streamlines
-
-            tubes = actor.line(streamlines)
-
-            number_streamlines = len(streamlines)
+            if bundle_ext in ['.tck', '.trk']:
+                if args.random_coloring is None and args.uniform_coloring is None:
+                    colors = None
+                bundle_tractogram_file = nib.streamlines.load(bundle_file)
```
arnaudbore, line 235: why not load_with_ref ? or at least check compatibility between anat and bundles.

PR #215, opened 2020-04-19
improve json output when using filtering list

arnaudbore:
Data for testing: [here](https://drive.google.com/open?id=1FTDlAngdEuN1odEad61e9aZcKKhhkQwV)

### DIFF
--- a/scripts/scil_filter_tractogram.py
+++ b/scripts/scil_filter_tractogram.py
@@ -232,6 +232,11 @@ def main():
             curr_dict['streamline_count_after_filtering'] = len(sft.streamlines)
             o_dict[filtering_Name] = curr_dict
 
+    # Streamline count after filtering
+    o_dict['streamline_count_final_filtering'] = len(sft.streamlines)
+    if args.display_counts:
+        print(json.dumps(o_dict, indent=args.indent))
+
     if not filtered_sft:
         if args.no_empty:
             logging.debug("The file {} won't be written (0 streamline)".format(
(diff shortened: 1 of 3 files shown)


scripts/scil_filter_tractogram.py
```diff
@@ -224,11 +244,9 @@ def main():
     save_tractogram(sft, args.out_tractogram)
 
     # Streamline count after filtering
-    sc_af = len(sft.streamlines)
+    o_dict['streamline_count_final_filtering'] = len(sft.streamlines)
     if args.display_counts:
-        print(json.dumps({'streamline_count_before_filtering': int(sc_bf),
-                          'streamline_count_after_filtering': int(sc_af)},
-                         indent=args.indent))
+        print(json.dumps(o_dict, indent=args.indent))
```
frheault, line 249: To make that option usable you should fix the warning at line
` mask = img.get_data().astype(np.uint16)`

Right now if you redirect the print to a file you get the warning, making the json usuable

scripts/scil_filter_tractogram.py
```diff
@@ -211,6 +226,11 @@ def main():
 
         sft = filtered_sft
 
+        if only_filtering_list:
+            filtering_Name = 'Filter_' + str(i)
+            curr_dict['tract_count_after_filtering'] = len(sft.streamlines)
```
frheault, line 231: streamline_count_after_filtering

scripts/scil_filter_tractogram.py
```diff
@@ -133,20 +140,28 @@ def main():
...
     for i, roi_opt in enumerate(roi_opt_list):
+        curr_dict = {}
         # Atlas needs an extra argument (value in the LUT)
         if roi_opt[0] == 'atlas_roi':
             filter_type, filter_arg, filter_arg_2, \
                 filter_mode, filter_criteria = roi_opt
         else:
             filter_type, filter_arg, filter_mode, filter_criteria = roi_opt
+
+        curr_dict['Filename'] = os.path.abspath(filter_arg)
```
frheault, line 160: Are you sure about the upper case ? I personally don't like it, I prefer all lower case (optional)

PR #217, opened 2020-04-24
Enhance streamlines math

frheault:
This will have to wait for Dipy 1.2 (my PR on adding and slicing SFT)
This script removes the hashing algorithm to match streamlines and replace it with an approach based on ckdtree (and some workaround the reduce the amount of processing).

This allows for the same behavior as before in roughly the same time, but also allows to match streamlines with shift and/or numerical impression. Can match even with a 1mm shift or a slight rotation (less than 1 degree) or a np.round(affine, 2) or the transformation matrix. This will enable the script to match streamlines that were modified sometime by accident or due to other software not handling the transformation well.

Would have been very useful for Kurt's project and/or white matter challenge 2015. Still keep the data_per_* and all. 
All you need to test is 2 tractograms or 2 bundles. Please try anything to find corner case.


frheault:
This is now working with Dipy master, the new 1.2 tag should come soon-ish (next month)

scripts/scil_streamlines_math.py
```diff
@@ -14,38 +14,36 @@
 
     concatenate:  Keep all streamlines with duplicates.
 
-For efficiency, the comparisons are performed using a hash table. This means
-that streamlines must be identical for a match to be found. To allow a soft
-match, use the --precision option to round streamlines before processing.
-Note that the streamlines that are saved in the output are the original
-streamlines, not the rounded ones.
+If a file 'duplicate.trk' have identical streamlines calling the script using
```
mdesco, line 17: has identical streamlines, calling

scripts/scil_streamlines_math.py
```diff
@@ -14,38 +14,36 @@
...
+the difference/intersection/union with a single input will remove these
+duplicated streamlines.
+
+To allow a soft match, use the --precision option to increase the allowed
+threshold for similarity. A precision of 1 represent 10**(-1), so a
```
mdesco, line 22: represents

scripts/scil_streamlines_math.py
```diff
@@ -14,38 +14,36 @@
...
+maximum distance of 0.1mm  is allowed. If the streamlines are identical, the
```
mdesco, line 23: extra space after mm

scripts/scil_streamlines_math.py
```diff
@@ -14,38 +14,36 @@
...
+default value of 3 or 0.001mm distance) should work. If there is a 0.5mm shift,
```
mdesco, line 24: (or

arnaudbore (approved):

PR #222, opened 2020-05-01
BF: Load mask as bool

GuillaumeTh:
DTI metrics script crash when we give a mask. Setting the type as bool fixed the issue.


frheault:
Well the code is very easy, but is the problem coming from Dipy or something?
Is it new from dipy 1.1 ?

PR #226, opened 2020-05-05
Commit integration

frheault:
Integration of user-friendly COMMIT script.
The requirements.txt was the only clean way I found to install what I needed (I used two specific COMMIT rather than user master as a tag)

This script computes commit for multi-shells data by default and the `--disable_zeppelin` revert makes it usable for single-shell.

@mdesco I actually don't know if you always use default parameters or if you want some COMMIT parameters available in the argsparser, if you want some and not other I would be happy to add them.

I don't have any ground truth or easy data to test, @mdesco do you think we could as Matteo or Mario to test it in some way to check our params or something, or you have everything yourself?

Implementation details:
-  highjack the stdout at the python and C level to control COMMIT logging (that is actually just print). So without `-v` there is nothing and `-v` only contains few logging and the rest is just the debugging from COMMIT.
- I also use a temporary directory to control the output of files that I couldn't deal with and I move only the relevant output and added our own (tractogram with weights for each streamline and/or filtered tractogram)
- Since COMMIT only handles .trk and .tck there is no reason to include anything from the `StatefulTractogram` in that script. However, to have a sanity check I verify if the input tractogram (if it is a .trk fits the DWI nifti file) just to try catching something simple.

Here is some testing data:
https://drive.google.com/open?id=1OPHTWRKjEomlBftpRoSg8_kjDdTpfP12


scilpy/io/streamlines.py
```diff
@@ -102,3 +96,75 @@ def load_tractogram_with_reference(parser, args, filepath,
...
+    lengths_filename = os.path.join(tmp_dir.name, 'lengths.dat')
+    lengths = np.memmap(lengths_filename, dtype='int32', mode='w+',
+                        shape=input_streamlines._lengths.shape)
+    lengths[:] = input_streamlines._lengths[:]
+
+    return tmp_dir, (data_filename, offsets_filename, lengths_filename)
+
+
+def reconstruct_streamlines_from_memmap(memmap_filenames, indices=None):
+    data = np.memmap(memmap_filenames[0],  dtype='float32', mode='r')
```
arnaudbore, line 134: docstring

scilpy/io/streamlines.py
```diff
@@ -102,3 +96,75 @@ def load_tractogram_with_reference(parser, args, filepath,
...
+    offsets = np.memmap(memmap_filenames[1],  dtype='int64', mode='r')
+    lengths = np.memmap(memmap_filenames[2],  dtype='int32', mode='r')
+
+    return reconstruct_streamlines(data, offsets, lengths, indices=indices)
+
+
+def reconstruct_streamlines_from_hdf5(hdf5_filename, key=None):
+    if isinstance(hdf5_filename, str):
```
arnaudbore, line 142: docstring

scilpy/io/streamlines.py
```diff
@@ -102,3 +96,75 @@ def load_tractogram_with_reference(parser, args, filepath,
...
+        group = hdf5_file
+
+    data = np.array(group['data']).flatten()
+    offsets = np.array(group['offsets'])
+    lengths = np.array(group['lengths'])
+
+    return reconstruct_streamlines(data, offsets, lengths)
+
+
+def reconstruct_streamlines(data, offsets, lengths, indices=None):
```
arnaudbore, line 161: docstring

mdesco:
Ping me when this is ready for me to have a go.

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Estimate the fit between a provided tractogram and DWI. Assign a weight to each
```
mdesco, line 5: COMMIT means Convex Optimization Modeling for Microstructure Informed Tractography. COMMIT estimates, globally, how a given tractogram explains the DWI in terms of signal fit, assuming a certain forward microstructure model. It assigns a weight to each streamline, which represents how well it explains the DWI signal globally. The default forward microstructure model is stick-zeppelin-ball, which required multi-shell data. It is possible to use the ball-and-stick model for single-shell data.

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+streamline that represent how well they explain the signal.
+Default is stick-zeppelin-ball for multi-shells data.
+
+The real output from COMMIT is:
```
mdesco, line 9: remove 'real'

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+- fit_NRMSE.nii.gz
+    fiting error (Normalized Root Mean Square Error)
+- fit_RMSE.nii.gz
+    fiting error (Root Mean Square Error)
+- results.pickle
+    Dictionary containing the experiment parameters and final weights
+- compartment_EC.nii.gz
+- compartment_IC.nii.gz
+- compartment_ISO.nii.gz
+    Each of COMMIT compartments
```
mdesco, line 19: EC := Extra-Cellular comparment. IC := Intra-Cellular compartment. ISO := isotropic volume fraction (or freewater comportment)

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+
+This script can divide the input tractogram in two using a threshold to apply
+on the streamlines' weight.
```
mdesco, line 22: Typically, the threshold should be 0, keeping only streamlines that have non-zero weight and that contribute to explain the DWI signal. Streamlines with 0 weight are essentially not necessary according to COMMIT.

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(description=__doc__,
+                                formatter_class=argparse.RawTextHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help='Input tractogram (.trk or .tck or .h5).')
+    p.add_argument('in_dwi',
+                   help='Diffusion from which the fodf were computed.')
```
mdesco, line 64: Not necessarily... I would just write Diffusion-weighted images used by COMMIT

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    p.add_argument('in_bvals',
+                   help='Bvals in the FSL format.')
```
mdesco, line 66: (.bval)

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    p.add_argument('in_bvecs',
+                   help='Bvecs in the FSL format..')
```
mdesco, line 68: (.bvec)

Normally we don't use capital B, no? 
I would write b-values, b-vectors

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    p.add_argument('in_peaks',
+                   help='Peaks extracted from the fodf.')
```
mdesco, line 70: These are optional in the case of B&Stick... Should this be mandatory or optional? 

Should we have 2 scripts? scil_run_multi-shell_commit.py and scil_run_single-shell_commit.py to differentiate the two more easily?

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    p.add_argument('out_dir',
+                   help='Output directory for the COMMIT maps.')
+
+    p.add_argument('--tracking_mask',
+                   help='Binary mask were tratography was allowed.\n'
```
mdesco, line 75: where

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    g1.add_argument('--perpendicular_diff', nargs='+', type=float,
+                    help='Perpendicular diffusivity in mm^2/s.\n'
+                    'Default for ball_stick: None\n'
+                    'Default for stick_zeppelin_ball: '
+                    '[1.19E-3, 0.85E-3, 0.51E-3, 0.17E-3]')
+    g1.add_argument('--isotropic_diff', nargs='+', type=float,
+                    help='Istropic diffusivity in mm^2/s.\n'
+                    'Default for ball_stick: 2.0E-3\n'
+                    'Default for stick_zeppelin_ball: [1.7E-3, 3.0E-3]')
+
```
mdesco, line 96: Ok finally, I find this pretty clear. Maybe no need for 2 scripts. But peaks are not mandatory for B&S

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    g2 = p.add_argument_group(title='Tractogram options')
+    g2.add_argument('--assign_weights', action='store_true',
+                    help='Store the streamlines weights in the '
+                    'data_per_streamline.')
+    g2.add_argument('--threshold_weights', type=float, metavar='THRESHOLD',
+                    help='Split the tractogram in two. Valid and invalid, '
+                    'based on the provided threshold.')
```
mdesco, line 103: Lets use other terms than valid and invalid. In fact, in the 'invalid' recobundle finds all bundles... so, it's not that the streamlines are invalid but not necessary to explain the signal. 

Not sure what the best terms are. 
'streamlines_needed' and 'unneeded_streamlines'...I'm sure we can do better

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+
+    g3 = p.add_argument_group(title='Kernels options')
+    kern = g3.add_mutually_exclusive_group()
+    kern.add_argument('--save_kernels', metavar='DIRECTORY',
+                      help='Output directory for the COMMIT kernels.')
+    kern.add_argument('--load_kernels', metavar='DIRECTORY',
+                      help='Input directory where the COMMIT kernels are '
+                           'located.')
```
mdesco, line 111: I think we need to add a boolean for regenerate_kernels=true or false. True by default.

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,344 @@
...
+    add_processes_arg(p)
```
mdesco, line 112: Other 2 crucial parameters we need access to;
i) ndirs=500 in the dictionnary
ii) max_iter = 500

scripts/scil_run_commit.py
mdesco: We also need a --tolerance on bvals as we have in other scripts.
COMMIT needs rounded bvals otherwise, this explodes the number of kernels and atoms in the dictionnary. Maybe default = 20

scripts/scil_run_commit.py
frheault: That would mean that they are always saved in the out_dir and if you use regenerate_kernels=True its going to look for them in the out_dir.
Also it would mean that --regenerate_kernels=True cannot be used with --load_kernels
Also it would mean that --save_kernel save a copy of the kernel elsewhere than out_dir on top of the copy in out_dir.

Is that an ok behavior ?

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,370 @@
...
+                         'Default for stick_zeppelin_ball: [1.7E-3, 3.0E-3]')
+
+    g2 = p.add_argument_group(title='Tractogram options')
+    g2.add_argument('--assign_weights', action='store_true',
+                    help='Store the streamlines weights in the '
+                         'data_per_streamline.')
+    g2.add_argument('--threshold_weights', type=float, metavar='THRESHOLD',
+                    default=None,
+                    help='Split the tractogram in two. essential and\n'
+                         'nonessential, based on the provided threshold.')
```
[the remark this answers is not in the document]
frheault: Well I think this quickly "explodes" in terms of naming (adding 4-5 news arguments to the script). I think having a single folder output is good enough at the moment. The script is already becoming a monster along scil_decompose_connectivity.py.

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,378 @@
...
+- fit_NRMSE.nii.gz
+    fiting error (Normalized Root Mean Square Error)
+- fit_RMSE.nii.gz
+    fiting error (Root Mean Square Error)
+- results.pickle
+    Dictionary containing the experiment parameters and final weights
+- compartment_EC.nii.gz (Extra-Cellular)
+- compartment_IC.nii.gz (Intra-Cellular)
+- compartment_ISO.nii.gz (isotropic volume fraction (freewater comportment))
+    Each of COMMIT compartments
```
mdesco, line 26: Can you add the essential.trk and non_essential.trk in the list here and
commit_weights.txt

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,378 @@
...
+    p.add_argument('in_bvals',
+                   help='Bvals in the FSL format.')
+    p.add_argument('in_bvecs',
+                   help='Bvecs in the FSL format..')
+    p.add_argument('out_dir',
+                   help='Output directory for the COMMIT maps.')
+
+    p.add_argument('--b0_thr', type=float, default=0.0,
+                   help='All b-values with values less than or equal '
+                        'to b0_thr are considered as b0s.')
```
mdesco, line 84: Add the defaultl in the help please

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,378 @@
...
+    p.add_argument('--nbr_dir', type=int, default=500,
+                   help='Number of directions, on the half of the sphere,\n'
+                        'representing the possible orientations of the response '
+                        'functions [%(default)s].')
+    p.add_argument('--nbr_iter', type=int, default=500,
+                   help='Maximum number of iterations [%(default)s].')
+    p.add_argument('--in_peaks',
+                   help='Peaks file representing principal direction(s) '
+                        'locally,\ntypically coming from fODFs. This file is '
+                        'mandatory for the default\n stick-zeppelin-ball model, '
```
mdesco, line 94: extra space to remove after \n

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,378 @@
...
+                         'Default for stick_zeppelin_ball: '
+                         '[1.19E-3, 0.85E-3, 0.51E-3, 0.17E-3]')
+    g1.add_argument('--iso_diff', nargs='+', type=float,
+                    help='Istropic diffusivity in mm^2/s.\n'
+                         'Default for ball_stick: [2.0E-3]\n'
+                         'Default for stick_zeppelin_ball: [1.7E-3, 3.0E-3]')
+
+    g2 = p.add_argument_group(title='Tractogram options')
+    g2.add_argument('--keep_whole_tractogram', action='store_true',
+                    help='Save a tractogram copy streamlines weights in the '
```
mdesco, line 121: Save a tractogram copy with streamline weights in the

scripts/scil_run_commit.py
```diff
@@ -0,0 +1,378 @@
...
+                         'data_per_streamline\n[default: False].')
+    g2.add_argument('--threshold_weights', type=float, metavar='THRESHOLD',
+                    default=0.,
+                    help='Split the tractogram in two. essential and\n'
```
mdesco, line 125: Split the tractogram in two, essential 

maybe ; or , not sure

scilpy/io/streamlines.py
```diff
@@ -29,6 +29,20 @@ def check_tracts_same_format(parser, tractogram_1, tractogram_2):
             'Input and output tractogram files must use the same format.')
 
 
+def lazy_streamlines_count(in_tractogram_path):
```
arnaudbore, line 32: docstring

PR #230, opened 2020-05-07
Added the cut options for streamlines

frheault:
Here is a small addition to `scil_remove_invalid_streamlines.py` that will cut streamlines that are out of the volume bounding box (and remove the streamline if it is 100% out). It is optional with `--cut_invalid`, however it could be argued that it should be our default.

In order to keep the data_per_point on the rest of the streamline, I am not resampling or using a line-based strategy. This will simply remove the entire segment if one of their points is out. So if the step-size is 10mm, and the segment is out of the bounding box by 1mm it means the 9mm that is perfectly valid will be removed too.

This is only a slight upgrade from removing the entire streamline.
[tiny_test_cut_PR.zip](https://github.com/scilus/scilpy/files/4595447/tiny_test_cut_PR.zip)

### DIFF
--- a/scilpy/utils/streamlines.py
+++ b/scilpy/utils/streamlines.py
@@ -289,7 +289,22 @@ def compress_sft(sft, tol_error=0.01):
 
 
 def cut_invalid_streamlines(sft):
-    """ Cut streamlines so their longest segment are within the bounding box """
+    """ Cut streamlines so their longest segment are within the bounding box.
+    This function keeps the data_per_point and data_per_streamline.
+
+    Parameters
+    ----------
+    sft: StatefulTractogram
+        The sft to remove invalid points from.
+
+    Returns
+    -------
+    new_sft : StatefulTractogram
+        New object with the invalid points removed from each streamline.
+    cutting_counter : int
+        Number of streamlines that were cut.
+    """
+
     sft.to_vox()
     sft.to_corner()
 
@@ -302,6 +317,7 @@ def cut_invalid_streamlines(sft):
     for key in sft.data_per_point.keys():
         new_data_per_point[key] = []
 
+    cutting_counter = 0
     for ind in range(len(sft.streamlines)):
         # No reason to try to cut if all points are within the volume
         if ind in indices_to_remove:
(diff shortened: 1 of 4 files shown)


scilpy/utils/streamlines.py
```diff
@@ -285,3 +286,48 @@ def compress_sft(sft, tol_error=0.01):
     compressed_sft.to_space(orig_space)
 
     return compressed_sft
+
+
+def cut_invalid_streamlines(sft):
+    """ Cut streamlines so their longest segment are within the bounding box """
```
arnaudbore, line 292: clean docstring

arnaudbore: should it be cut_invalid_points ?

scilpy/utils/streamlines.py
```diff
@@ -285,3 +286,48 @@ def compress_sft(sft, tol_error=0.01):
...
+    copy_sft = copy.deepcopy(sft)
+    indices_to_remove, indices_to_keep = copy_sft.remove_invalid_streamlines()
+    epsilon = 0.001
+
+    new_streamlines = []
+    new_data_per_point = {}
+    for key in sft.data_per_point.keys():
+        new_data_per_point[key] = []
+
+    for ind in range(len(sft.streamlines)):
```
arnaudbore, line 305: instead of looking at each index why not taking all indices_to_keep and then loop just the indices_to_remove ?

mdesco:
Tried it on your tiny example but everything I try does not cut anything. Are there streamlines to cut in your example?

scil_remove_invalid_streamlines.py CC_AGWM.trk test.trk --cut_invalid --reference mni_masked.nii.gz -f
WARNING:root:Removed 0 invalid streamlines.

scil_remove_invalid_streamlines.py CC_AGWM_lin.trk test.trk --cut_invalid --reference mni_masked.nii.gz -f
WARNING:root:Removed 0 invalid streamlines.

scilpy/utils/streamlines.py
frheault: Well, technically, you are cutting streamlines, points are indivisible entities

frheault:
@mdesco that's because a cut streamline is not removed from the tractogram.
I will add a new message about the cut one.

scilpy/utils/streamlines.py
frheault: Because since I want to keep data_per_point and data_per_streamline I need to generate a new list of streamlines and two new dictionaries of data that have the same ordering as before. 

So I need the full for loop to append in the right order the cut as well as the intact streamlines

mdesco:
Ready to go!

PR #237, opened 2020-06-03
Supports dps for connectivity computation

frheault:
Final step for COMMIT integration and connectivity matrices.
Supporting any data_per_streamline and generate an average of that data for each connection.

### DIFF
--- a/scripts/scil_compute_connectivity.py
+++ b/scripts/scil_compute_connectivity.py
@@ -70,9 +70,10 @@ def _processing_wrapper(args):
     labels_img = args[1]
     in_label, out_label = args[2]
     measures_to_compute = copy.copy(args[3])
-    weighted = args[4]
-    if args[5] is not None:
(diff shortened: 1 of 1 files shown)


scripts/scil_compute_connectivity.py
```diff
@@ -190,6 +197,10 @@ def _build_arg_parser():
                    help='Use density-weighting for the metric weighted matrix.')
     p.add_argument('--no_self_connection', action="store_true",
                    help='Eliminate the diagonal from the matrices.')
+    p.add_argument('--include_dps', action="store_true",
+                   help='Save matrices from data_per_streamline.')
+    p.add_argument('--force_labels_list', metavar='FILE',
```
arnaudbore, line 202: ??

arnaudbore: + help

PR #238, opened 2020-06-04
Skip indices with single points

frheault:
Solves bug (Max) with SET output for Sami's project.

`ValueError: Shape of array too small to calculate a numerical gradient, at least (edge_order + 1) elements are required.`


mdesco:
@arnaudbore we need this quickly please ;)

PR #241, opened 2020-06-08
Fix datatype for fdata and change requirements

GuillaumeTh:
Fix get_fdata() datatype to float32 to reduce memory and change requirements typo for dipy to be less strict

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ bz2file==0.98.*
 coloredlogs==10.0.*
 cycler==0.10.*
 Cython==0.29.*
-dipy==1.1.*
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@ bz2file==0.98.*
 coloredlogs==10.0.*
 cycler==0.10.*
 Cython==0.29.*
-dipy==1.1.*
+dipy>=1.1.*
```
frheault, line 5: dipy>=1.1.0 ? What the behavior of a * and a >= together

GuillaumeTh: I will do dipy>=1.1

PR #242, opened 2020-06-09
Reduce ram usage in bias field correction

frheault:
This script was taking more than 32GB of RAM on HCP data, so I switch to an approach with masking as early as possible and chunks.

This reduced by a factor 4-5 the amount of RAM needed when not using a mask and 2-3 (at least) when using one.

Otherwise, same behavior as before.


arnaudbore:
Multiple times tested by @frheault - Trust is powerful :+1:

PR #245, opened 2020-06-14
Update compute_pft

arnaudbore:


scripts/scil_compute_maps_for_particle_filter_tracking.py
```diff
@@ -64,39 +64,39 @@ def main():
...
                          [args.include, args.exclude, args.interface])
 
-    wm_pve = nib.load(args.wm)
-    logging.info('"{0}" loaded as WM PVE map.'.format(args.wm))
-    gm_pve = nib.load(args.gm)
-    logging.info('"{0}" loaded as GM PVE map.'.format(args.gm))
-    csf_pve = nib.load(args.csf)
-    logging.info('"{0}" loaded as CSF PVE map.'.format(args.csf))
+    wm_pve = nib.load(args.in_wm)
+    logging.info('"{0}" loaded as WM PVE map.'.format(args.in_wm))
```
frheault, line 72: Are these loggings really useful?

scripts/scil_compute_maps_for_particle_filter_tracking.py
```diff
@@ -64,39 +64,39 @@ def main():
...
-    interface[gm_pve.get_data() >= args.int_thres] = 1
-    interface[wm_pve.get_data() < args.int_thres] = 0
+    interface[gm_pve.get_fdata(dtype=np.float32) >= args.int_thres] = 1
+    interface[wm_pve.get_fdata(dtype=np.float32) < args.int_thres] = 0
 
     logging.info('The interface "{0}" contains {1} voxels.'.format(
         args.interface, int(np.sum(interface))))
 
-    nib.Nifti1Image(include_map.astype('float32'),
+    nib.Nifti1Image(include_map.astype(np.float32),
```
frheault, line 95: In scilpy, the vast majority of saving is done using nib.save(nib.Nifti1Image()). I think you should change it just for consistency.

scripts/scil_compute_maps_for_particle_filter_tracking.py
```diff
@@ -64,39 +64,39 @@ def main():
...
                     gm_pve.affine).to_filename(args.include)
-    nib.Nifti1Image(exclude_map.astype('float32'),
+    nib.Nifti1Image(exclude_map.astype(np.float32),
                     gm_pve.affine).to_filename(args.exclude)
-    nib.Nifti1Image(interface.astype('float32'),
+    nib.Nifti1Image(interface.astype(np.float32),
```
frheault, line 99: If you declare interface as float32 (in the zeros()), they will all be in float32 already.

scripts/scil_compute_maps_for_particle_filter_tracking.py
```diff
@@ -27,31 +27,31 @@ def _build_arg_parser():
     p = argparse.ArgumentParser(
         description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
     p.add_argument(
```
frheault, line 29: Maybe its just me, but I prefer when as much argument are on the same line (in this case, 'in_wm')

scripts/scil_compute_pft.py
```diff
@@ -54,23 +55,23 @@ def _build_arg_parser():
     p._optionals.title = 'Generic options'
 
     p.add_argument(
```
frheault, line 57: Same comment as above, I prefer when as many arguments are on the same line.

PR #248, opened 2020-06-14
Update some scripts 2

arnaudbore:


scripts/scil_convert_rgb.py
```diff
@@ -66,9 +67,7 @@ def main():
     elif original_im.ndim == 3:
         converted_im_float = decfa_to_float(original_im)
 
-        converted_data = converted_im_float.get_fdata()
-        converted_data_int = converted_data.astype("uint8")
-
+        converted_data_int = get_data_as_mask(converted_im_float)
```
frheault, line 70: Since an RGB is not a mask, I think this is (1) confusing and (2) will likely crash due to the fact that RGB contain more than one value. It should stay like before.

scripts/scil_evaluate_bundles_pairwise_agreement_measures.py
```diff
@@ -112,9 +113,8 @@ def load_data_tmp_saving(filename, reference, init_only=False,
         # If initilization, loading the data is useless
         if init_only:
             return None
-        density = nib.load(tmp_density_filename).get_fdata().astype(np.uint16)
-        endpoints_density = nib.load(
-            tmp_endpoints_filename).get_fdata().astype(np.uint16)
+        density = get_data_as_label(nib.load(tmp_density_filename))
```
frheault, line 116: A density map is not labels, this is again confusing when reading the code later. From other scripts, I think you could simply switch it to float32.

scripts/scil_convert_rgb.py
arnaudbore: Ok will switch back

scripts/scil_evaluate_bundles_pairwise_agreement_measures.py
arnaudbore: It was np.uint16 you really want to switch to float32

PR #258, opened 2020-06-18
Integration testing others

frheault:
Followup of testing, with utility scripts


arnaudbore:
A couple of warnings (46) that could be fixed at the same time, also I miss some from a previous PR.
Could be done in another PR as well. Your call. Otherwise LGTM

PR #262, opened 2020-06-24
Integration testing commit amico

frheault:
Follow up of test for commit/amico related script

Missing test for NODDI due to weird behavior


arnaudbore:
Not tested, @mdesco can you test it ?

PR #271, opened 2020-07-02
Support for peaks/evecs in local_tracking

frheault:
Added heuristic for peaks/fodf (EUDX) support in local tracking, uniformize the arg parse.

This is not "faster" but simply allows the user to use peaks directly. Actually, EUDX was not even in the option so it wasn't really available anyway.
This also allows good old DTI tracking if needed for a publication (baseline). However, it is not FACT as most papers do, but simply EUDX with single peak per voxel

The script is still made to work primarily with fodf, but support for peaks is useful.


scripts/scil_compute_local_tracking.py
```diff
@@ -48,69 +56,60 @@ def _build_arg_parser():
...
+                         help='Step size in mm. [%(default)s]')
+    track_g.add_argument('--min_length', type=float, default=10.,
+                         help='Minimum length of a streamline in mm. '
+                              '[%(default)s]')
+    track_g.add_argument('--max_length', type=float, default=300.,
+                         help='Maximum length of a streamline in mm. '
+                              '[%(default)s]')
+    track_g.add_argument('--theta', type=float,
+                         help='Maximum angle between 2 steps.\n'
+                              '["eudx"=60, det"=45, "prob"=20]')
```
arnaudbore, line 82: "det"

scripts/scil_compute_fodf_metrics.py
```diff
@@ -156,10 +156,10 @@ def main():
                                      nbr_processes=args.nbr_processes)
 
     # Computing maps
-    nufo_map, afd_max, afd_sum, rgb_map, \
-        _, _ = maps_from_sh(data, peak_dirs,
-                                       peak_values, peak_indices,
-                                       sphere, nbr_processes=args.nbr_processes)
+    if args.nufo or args.afd_max or args.afd_total or args.afd_sum or args.rgb:
+        nufo_map, afd_max, afd_sum, rgb_map, \
```
arnaudbore, line 160: remove import os at the begining + format parser the same way you did the others

GuillaumeTh:
@frheault can you fix Max's comments. I will do the last check after.

PR #279, opened 2020-07-14
Update readme for lapack et blas

frheault:
A simple line to direct people in the right direction for COMMIT


samuelstjean:
you can also put in libopenblas-dev, it's a replacement for both (and multithreaded)

Issue #285, opened 2020-07-21
scil_compute_connectivity allclose 'affine' too strict


mdesco:
This command breaks:
scil_compute_connectivity.py decompose_sub-1144_ses-2.h5 atlas/sub-1144_ses-2/brainnetome.nii.gz --volume volume.npy --streamline_count track-count.npy --length length.npy --metrics *fa.nii.gz fa.npy --density_weighting --no_self_connection -f

On this data:
https://www.dropbox.com/s/ursgtsbn26iakx1/conn.zip?dl=0

As recommanded by @frheault, if I comment the allclose 'affine' lines of the code, things run and seem to be fine.

frheault:
@mdesco I cant see the data, but I made a fix. Can you test it (see latest PR)? Or send me back the data?

mdesco:
Works all good now with P292

PR #292, opened 2020-08-03
Add atol to float comparison in scilpy

frheault:
Mainly to address issue #285 but similar code related to hdf5 was fixed at the same time (and other np.allclose)
Comparison between integer is now made using np.array_equal instead.


arnaudbore:
Need to be tested with @mdesco data

frheault:
> Need to be tested with @mdesco data

Anyone still have it ? The dropbox link does not work.

Issue #282, opened 2020-07-17
scil_reorder_connectivity better help


mdesco:
1) --labels_list LABELS_LIST
                        List saved by the decomposition script,
                        the json must contain labels rather than coordinates.

Please say this is a txt file and add (.txt) at the end.

2) Also, not a fan of --prefix  why not just fix the output. Is it because the input can infact be *npy ? 

3) would be cool that input is *npy because if I start working on a sub-connectome of the full connectome, I want to reorder my vol.npy, len.npy, fa.npy, track-count.noy, etc... 

Thanks @frheault

frheault:
@arnaudbore to close

PR #294, opened 2020-08-03
Update for suffix and more robust visualisation

frheault:
Final update for Maxime request of #282 
Cover more use case from connectoflow


arnaudbore:
LGTM @mdesco can you test it?

frheault:
> LGTM @mdesco can you test it?

You can test it using the data hidden in `~/.scilpy/connectivity` ;)

PR #297, opened 2020-08-12
update display streamline count

arnaudbore:


scripts/scil_detect_streamlines_loops.py
```diff
@@ -56,11 +58,15 @@ def _build_arg_parser():
     p.add_argument('-a', dest='angle', default=360, type=float,
                    help='Maximum looping (or turning) angle of\n' +
                         'a streamline in degrees. [%(default)s]')
+    p.add_argument('--display_counts', action='store_true',
+                   help='Print streamline count before and after filtering')
```
GuillaumeTh, line 62: Add default value at the end.

PR #298, opened 2020-08-12
Add containers

arnaudbore:
Yeah containers for scilpy :) Docker and Singularity


containers/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt-get update
+RUN apt-get -y upgrade
+
+RUN apt -y install libblas-dev
+RUN apt -y install liblapack-dev
+RUN apt -y install libgl1-mesa-glx
+
+WORKDIR /
+ENV SCILPY_VERSION="master"
```
GuillaumeTh, line 11: Why downloading a zip instead of doing a git clone to take automatically the master ?

GuillaumeTh: I merge. We can fix it later if necessary. Thanks !

PR #301, opened 2020-08-12
Fix condition

GuillaumeTh:


GuillaumeTh:
Fix the condition to build latest docker.

PR #302, opened 2020-08-12
Robustify parallel fit

frheault:
There was some overflow in term of int32 from multiprocessing due to the fact that the array_split was done on the whole volume rather than where values existed in the voxel. This was causing a crash for some HCP subject, now only voxels with values are split and send across processes.

Discover a small bug for the RGB map were a normalization was performed using the max value of the array, only when it is split into chunks each chunk has a different maximum value leading to a chunk intensity effect that was visible when --processes > 1.

All functions using chunks were corrected


frheault:
@karanphil Please test on HCP with your multishell stuff, also try 1/N processes with and without masks (for all 3 modified scripts)
This should be in the official scilpy release.

karanphil:
> @karanphil Please test on HCP with your multishell stuff, also try 1/N processes with and without masks (for all 3 modified scripts)
> This should be in the official scilpy release.

I ran `scil_compute_fodf.py` like you asked, using tournier07 to also test the `convert_sh_basis` changes. It all looks good. 

Sadly, I got an error while running `scil_compute_fodf_metrics.py`, at line 373 in `multi_processes.py`. @frheault 

Also, it seems to work on my msmt HCP data.

karanphil:
I tested `scil_compute_fodf.py` and `scil_compute_fodf_metrics.py` with tournier07, to test all your changes. Everything seems to work! Just note that I am not the most familiar with `scil_compute_fodf_metrics.py` outputs, so please have someone else check for the outputs. Thanks for this fix!

karanphil (approved):

arnaudbore (approved):
LGTM need to be tested

PR #306, opened 2020-08-19
add the possibility to add options to eddy or topup

arnaudbore:


frheault:
I would need an example command, I am not sure how I need to format the `--topup_options` and `--eddy_options`?

Do I need to put like that --eddy_options "--mb X --mb_offs Y --slspec Z". It is not clear from the code, but mainly I think an example or instruction should be included in the script help/description.

Otherwise, LGTM (not tested yet)

scripts/scil_prepare_eddy_command.py
```diff
@@ -77,6 +78,11 @@ def _build_arg_parser():
                    help='if set, will use the fixed seed strategy for eddy.\n'
                         'Enhances reproducibility.')
 
+    p.add_argument('--eddy_options',  default=None,
+                   help='Additional options you want to use to run eddy.'
```
frheault, line 82: Add \n at the end of the first sentence (to run eddy.)

scripts/scil_prepare_topup_command.py
```diff
@@ -66,6 +67,11 @@ def _build_arg_parser():
                         'else, will output the lines to the ' +
                         'terminal [%(default)s].')
 
+    p.add_argument('--topup_options',  default=None,
+                   help='Additional options you want to use to run topup.'
```
frheault, line 71: Add \n at the end of the first sentence (to run topup.)

arnaudbore:
> Is the goal to add this to tractoflow? If yes, some options related to output will have to be taken into account for the process to keep track of the right file. Otherwise,

Yes and no, it will be used so people can add extra options for topup and eddy like number of iterations or different factors but if a completely different behaviour is asked it will crash... I guess it's higher level of comprehension so people who will use it, will e careful. Neurostars: https://neurostars.org/t/pass-extra-arguements-to-processes/16504/7

GuillaumeTh (approved):
LGTM ! Thanks a lot !

Issue #235, opened 2020-06-02
Maybe we should detect this with a heuristic or send a warning if 100+ images given?


arnaudbore:
Maybe we should detect this with a heuristic or send a warning if 100+ images given?

_Originally posted by @mdesco in https://github.com/scilus/scilpy/pull/233_

frheault:
To close @arnaudbore

PR #307, opened 2020-08-19
Memory friendly scil_image_math.py

frheault:
Resolving issue #235 

I thought that a heuristic was not a robust/useful approach to common operations I know are possible (e.g adding 100+ labels in in16) so I modified all operations so they could be performed on the fly one image at the time.
This does not affect the speed in an observable way below 10 images, slightly slower than before above 25-50. To my surprise, since `AverageImages` is multiprocessed (1 main CPU and multiple readers) `scil_image_math.py ` is actually faster in CPU-times! So I think it's fine just like that.

Before all operations were capped at around 100 3D images in float32, now I can call any operation with hundreds of images!
Operations such as concatenate (of 3D images still rely on 4D image routine so this one have the same limitation as before

Operations having only one/two inputs (threshold, clip, difference, subtraction, division, convert) are now slightly slower than before due to the robustness increased (Not really observable, like 10% of 0.8sec).

Since all operations are performed using `nibabel` caching system and that operations were also used with `scil_connectivity_math.py` I had to switch to Nifti1Image for connectivity matrices before sending them to the operations.
This has no consequences on speed for `scil_connectivity_math.py` (None observable).

While testing everything I had to fix a few bugs & update the `scil_split_volume_by_*` scripts.
Since half of that code is also @ppoulin91 I think you could review/test it.

### DIFF
--- a/scilpy/image/operations.py
+++ b/scilpy/image/operations.py
@@ -417,16 +417,17 @@ def std(input_list, ref_img):
                 'This operation with 4D data only support one operand.')
     else:
         if len(input_list) == 1:
-            raise ValueError('This operation with only one operand requires 4D data.')
+            raise ValueError(
+                'This operation with only one operand requires 4D data.')
 
     if len(input_list[0].header.get_data_shape()) > 3:
         return np.std(input_list[0].get_fdata(dtype=np.float64), axis=-1)
     else:
-        mean_data=mean(input_list, ref_img)
-        output_data=np.zeros(input_list[0].header.get_data_shape())
+        mean_data = mean(input_list, ref_img)
+        output_data = np.zeros(input_list[0].header.get_data_shape())
         for img in input_list:
             if isinstance(img, nib.Nifti1Image):
-                data=img.get_fdata(dtype=np.float64)
+                data = img.get_fdata(dtype=np.float64)
                 output_data += (data - mean_data) ** 2
                 img.uncache()
             else:
@@ -440,8 +441,8 @@ def union(input_list, ref_img):
         Operation on binary image to keep voxels, that are non-zero, in at
         least one file.
     """
-    output_data=addition(input_list, ref_img)
-    output_data[output_data != 0]=1
+    output_data = addition(input_list, ref_img)
+    output_data[output_data != 0] = 1
 
     return output_data
 
@@ -452,8 +453,8 @@ def intersection(input_list, ref_img):
         Operation on binary image to keep the voxels, that are non-zero,
         are present in all files.
     """
-    output_data=multiplication(input_list, ref_img)
-    output_data[output_data != 0]=1
+    output_data = multiplication(input_list, ref_img)
+    output_data[output_data != 0] = 1
 
     return output_data
 
@@ -467,18 +468,18 @@ def difference(input_list, ref_img):
     _validate_length(input_list, 2)
     _validate_imgs(*input_list, ref_img)
 
-    output_data=np.zeros(ref_img.header.get_data_shape(), dtype=np.float64)
+    output_data = np.zeros(ref_img.header.get_data_shape(), dtype=np.float64)
     if isinstance(input_list[0], nib.Nifti1Image):
-        data_1=input_list[0].get_fdata(dtype=np.float64)
+        data_1 = input_list[0].get_fdata(dtype=np.float64)
     else:
-        data_1=input_list[0]
+        data_1 = input_list[0]
     if isinstance(input_list[1], nib.Nifti1Image):
-        data_2=input_list[1].get_fdata(dtype=np.float64)
+        data_2 = input_list[1].get_fdata(dtype=np.float64)
     else:
-        data_2=input_list[1]
+        data_2 = input_list[1]
 
-    output_data[data_1 != 0]=1
-    output_data[data_2 != 0]=0
+    output_data[data_1 != 0] = 1
+    output_data[data_2 != 0] = 0
     return output_data
 
 
@@ -490,10 +491,10 @@ def invert(input_list, ref_img):
     _validate_length(input_list, 1)
     _validate_type(input_list[0], nib.Nifti1Image)
 
-    data=input_list[0].get_fdata(dtype=np.float64)
-    output_data=np.zeros(data.shape, dtype=np.float64)
-    output_data[data != 0]=0
-    output_data[data == 0]=1
+    data = input_list[0].get_fdata(dtype=np.float64)
+    output_data = np.zeros(data.shape, dtype=np.float64)
+    output_data[data != 0] = 0
+    output_data[data == 0] = 1
 
     return output_data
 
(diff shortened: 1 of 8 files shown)


scilpy/image/operations.py
```diff
@@ -112,396 +113,463 @@ def _validate_float(x):
...
         not in the second file (non-zeros).
     """
     _validate_length(input_list, 2)
-    _validate_arrays(*input_list)
+    _validate_imgs(*input_list, ref_img)
 
-    output_data = copy(input_list[0]).astype(np.bool)
-    output_data[input_list[1] != 0] = 0
+    output_data = np.zeros(ref_img.header.get_data_shape(), dtype=np.float64)
+    if isinstance(input_list[0], nib.Nifti1Image):
```
ppoulin91, line 464: Minor issue: I feel like there are multiple instances of fetching data from either a Nifti1Image or an array itself, which could be put in a function like `load_data()` for readability

scilpy/image/operations.py
```diff
@@ -112,396 +113,463 @@ def _validate_float(x):
...
 
-def invert(input_list):
+def invert(input_list, ref_img):
     """
     invert: IMG
         Operation on binary image to interchange 0s and 1s in a binary mask.
     """
     _validate_length(input_list, 1)
-    _validate_arrays(*input_list)
+    _validate_type(input_list[0], nib.Nifti1Image)
```
ppoulin91, line 484: Why is the `Nifti1Image` type only sometimes enforced, but not always?

scripts/scil_connectivity_math.py
```diff
@@ -41,29 +42,33 @@ def _build_arg_parser():
...
-    p.add_argument('inputs', nargs='+',
+    p.add_argument('in_matrices', nargs='+',
                    help='The list of matrices files or parameters.')
     p.add_argument('out_matrix',
                    help='Output matrix path.')
 
     p.add_argument('--data_type',
                    help='Data type of the output image. Use the format: '
                         'uint8, float16, int32.')
+    p.add_argument('--exclude_background', action='store_true',
```
ppoulin91, line 53: Sneaking in a new feature in a refactoring PR ;)

ppoulin91 (commented):
Didn't test, but everything looks good, I only have a few minor comments

scripts/scil_connectivity_math.py
frheault: It already in the master, don't know why its green (But yeah I do that often)

scilpy/image/operations.py
frheault: Some operations requires the first arg to be a array/image (invert, lower_clip, etc) and other support number and array/image in any order (addition, multiplication, etc)

scilpy/image/operations.py
```diff
@@ -112,396 +113,463 @@ def _validate_float(x):
...
+            and not len(input_list[0].header.get_data_shape()) > 3:
         logging.error('This operation with only one operand requires 4D data.')
         raise ValueError
 
-    in_data = np.squeeze(np.rollaxis(np.array(input_list), 0,
-                                     input_list[0].ndim+1))
-
-    return np.std(in_data, axis=-1)
+    if len(input_list[0].header.get_data_shape()) > 3:
+        return np.std(input_list[0].get_fdata(dtype=np.float64), axis=-1)
```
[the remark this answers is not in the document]
frheault: So I should warn the user or crash? I would say crash, but I just want to be sure.

scilpy/image/operations.py
frheault: It is because I have to uncache() right after, so I need to load it, perform the operation, unload.
Even with a wrapper function, then I need another uncache() wrapping function, it adds up to a similar amount of line.

scilpy/image/operations.py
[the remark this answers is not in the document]
frheault: I now raise an error

scilpy/image/operations.py
ppoulin91: I see, fair enough!

scilpy/image/operations.py
ppoulin91: I get it for simple ops, but e.g in `difference()`, you can have either `Nifti1Image` or `np.array` as input, but here in `invert` only `Nifty1Image` is allowed, so what's the difference (no pun intended)?

scripts/scil_connectivity_math.py
ppoulin91: Fun fact: it's **not** in the master branch, you must have started your branch from another that isn't merged yet... Be careful, should this really be merged in this PR?

scilpy/image/operations.py
frheault: Because you can do a 1 - another image, 1 will be an array for broadcasting. Difference supports using value and image 

You cannot invert a 1, invert does not support value, just image

scripts/scil_connectivity_math.py
frheault: I just checked, it's there...
https://github.com/scilus/scilpy/blob/master/scripts/scil_image_math.py

ppoulin91: Your "new feature" is in `scil_connectivity_math.py`

ppoulin91:
I can test it tuesday morning, unless @GuillaumeTh or @arnaudbore want to test it before

scilpy/image/operations.py
```diff
@@ -112,396 +113,470 @@ def _validate_float(x):
...
-def union(input_list):
+def union(input_list, ref_img):
     """
     union: IMGs
         Operation on binary image to keep voxels, that are non-zero, in at
         least one file.
     """
-    output_data = addition(input_list)
-    output_data[output_data != 0] = 1
+    output_data=addition(input_list, ref_img)
```
ppoulin91, line 443: You had a formatting bug I guess, you are missing spaces around operators for no reason, check everywhere please.

ppoulin91 (changes requested):
Did a bunch of tests with 100s of images on scil_image_math.py, everything works fine. Small formatting bug to fix, then it's good for me.

scripts/scil_connectivity_math.py
frheault: It works as intended, but obviously less intuitive as the image background is a quite clearer concept.
For matrices it simply mean that everywhere you had zeros before you will still have 0 after, even if you do `scil_connectivity_math.py addition 17 *.npy masked_addition.npy --exclude background`
This would add 17 to all nonzeros values

Which would be equivalent to finding all zeros (like doing a binarization of all your npy, do an intersection of all binary mask, do the command `scil_connectivity_math.py addition 17 *.npy addition.npy` followed by `scil_connectivity_math.py multiplication addition.npy mask.npy masked_addition`

This is maybe overly specific as a use case, but I need it some time so I say we need it :100:

PR #308, opened 2020-08-20
add python 3.8 to test with Travis

arnaudbore:


arnaudbore:
Does not seem to be an easy fix.
python 3.8 ok with vtk 9.0.* but not python 3.5 but will require a lot of fix.

PR #311, opened 2020-09-02
Fix limitations of RBx for tractogram size

frheault:
There was a limit on the tractogram size (nb_points, nb_streamlines) due to multiprocessing. This is slightly slower for very big tractogram, but at least it does not crash.


arnaudbore (commented):
LGTM, I will need to test this.

boiteaclou (commented):
Everything good for me. Tested and it is robust. LGTM

boiteaclou (approved):

arnaudbore (approved):

PR #315, opened 2020-09-14
Adapting scil_compute_ssst_frf.py to changes in auto_response (WIP - dipy 1.2)

karanphil:
I adapted `scil_compute_ssst_frf.py` to my changes in Dipy, regarding the ssst response function calculation. Now, instead of calling `auto_response`, we call `mask_for_response_ssst` followed by `response_from_mask_ssst`. It does exactly the same thing, except now we als have the option to give a cuboid ROI instead of a cubic one.

I also thought it would be a good idea to change `scil_compute_fodf.py` to `scil_compute_ssst_fodf.py`, since `scil_compute_msmt_fodf.py` is coming in another PR and to follow `scil_compute_ssst_frf.py`'s logic.


frheault:
@karanphil is this to close?

PR #317, opened 2020-09-17
Fix RAM spike in decompose_connectivity

frheault:
Remove the `remove_invalid_streamlines` function, so no copy (spike in RAM)

Return immediately after knowing empty edge in compute_connectivity
A simple fix for empty edges, sometimes leading to np.nan for some metrics


arnaudbore (approved):

frheault:
@jchoude I change a voxel indices dtype in uncompress (which you coded I think) to uint16 instead of uint64.
I compared it in the two scripts it was called and it seem that it is nothing more than a voxels indices array which is always well below 1000 and casting it did not change the values and results.

Is there something I am missing? Is it safe to change it from uint64 to uint16?

jchoude:
As discussed on Slack, it will be ok like this.

PR #321, opened 2020-09-21
Update outliers removal and decompose speedup

frheault:
Update the outliers removal function and script so it is cleaner (dipy 1.2 sft slicing)
Decompose is SUPER slow and we only need an approx. for connectomics.
So I added parameters to control the removal of the outliers a bit better.
For decompose_connectivity I used less sampling (10 instead of 30) and instead of 25-30 levels of hierarchy, I use 12.
Making the overall function run 6 times faster on average for decompose, for similar results.


scripts/scil_outlier_rejection.py
```diff
@@ -11,34 +11,32 @@
...
+    p.add_argument('out_bundle',
+                   help='Fiber bundle without outliers.')
+    p.add_argument('--remaining_bundle',
+                   help='Removed outliers.')
+    p.add_argument('--alpha', type=float, default=0.6,
+                   help='Percent of the length of the tree that clusters '
+                   'of individual streamlines will be pruned.')
+    add_reference_arg(p)
+    add_overwrite_arg(p)
+    return p
```
arnaudbore, line 39: Why you don't have the fast option here ?

mdesco (commented):
scil_decompose_connectivity.py 100307_ensemble_prob_mni_10k.trk atlas_freesurfer.nii.gz decomposed_100307.h5 --save_raw_connections --save_intermediate --save_discarded --outlier_threshold 0.5 --loop_max_angle 325 --curv_qb_distance 8 -f --out_dir output
  File "/Users/desm2239/my_env/scilpy_public/bin/scil_decompose_connectivity.py", line 7, in <module>
  File "/Users/desm2239/Research/Source/scilpy.github/scripts/scil_decompose_connectivity.py", line 457, in <module>
    main()
  File "/Users/desm2239/Research/Source/scilpy.github/scripts/scil_decompose_connectivity.py", line 374, in main
    invalid_length_sft = raw_sft[invalid_length_ids]
TypeError: 'StatefulTractogram' object is not subscriptable


I get this error. Data is from your connectivity_PR from the past...

scripts/scil_outlier_rejection.py
frheault: Because we are more patient, 1-50 bundle that makes sense to be slow and reliable.
decompose is more like 3000-25000 that no one QA, so an approx is ok

mdesco:
Everything is running well. I'm still QC-ing a bit to make sure we are happy with the default.

mdesco (approved):
@frheault am ok with this PR. Please add a comment in the help about the outlier_rejection threshold. e.g. "The default is a recommended good trade-off for a freesurfer parcellation. With smaller parcels the threshold should most likely be reduced. Good candidate connections to QC are the brainstem to precentral gyrus connection and precentral left to precentral right connection, or equivalent in your parcellation."

arnaudbore (approved):

PR #323, opened 2020-09-22
Remove ascii characters

GuillaumeTh:


scripts/scil_compute_bundle_volume.py
```diff
@@ -2,11 +2,11 @@
 # -*- coding: utf-8 -*-
 
 """
-Compute bundle volume in mm³. This script supports anisotropic voxels
+Compute bundle volume in mm3. This script supports anisotropic voxels
```
AntoineTheb, line 5: _Maybe_ mm^3 could have been better but it's so inconsequential and this PR is urgent so no need to change it, but maybe we could keep it in mind and change it down the line, maybe

AntoineTheb (approved):

frheault (approved):

PR #330, opened 2020-09-30
Disconnectome mockup (WIP)

frheault:
@arnaudbore Could you test that?

You would need a tractogram (or many), labels (uint16), a lesions mask (uint8) and to read the help and try to figure it out.

I would need input before going forward. But I think this is going in the right direction.


frheault:
@arnaudbore if you think this is useful in scilpy, can you take care of that PR?
Like switch ownership style, just to remove it from my stack

arnaudbore:
Good try.

PR #333, opened 2020-10-07
FW b thr fix

mdesco:
Simply adding default b_thr in the --help


mdesco:
Would you guys agree to change --in_mask to --mask like in most scilpy scripts?

arnaudbore:
> Would you guys agree to change --in_mask to --mask like in most scilpy scripts?

Yes definitively in_ it's for mandatory arguments.

arnaudbore:
Too many merges, I'm starting a new PR.

PR #335, opened 2020-10-08
Update NODDI and freewater

arnaudbore:
data to test: https://www.dropbox.com/s/wktc2vbnb74efir/input.zip?dl=0

### DIFF
--- a/scripts/scil_compute_NODDI.py
+++ b/scripts/scil_compute_NODDI.py
@@ -11,8 +11,9 @@ from contextlib import redirect_stdout
 import io
 import logging
 import os
-import tempfile
+import shutil
 import sys
+import tempfile
 
 import amico
 from dipy.io.gradients import read_bvals_bvecs
@@ -22,6 +23,7 @@ from scilpy.io.utils import (add_overwrite_arg,
                              add_processes_arg,
                              add_verbose_arg,
                              assert_inputs_exist,
+                             assert_outputs_exist,
                              assert_output_dirs_exist_and_empty)
 from scilpy.utils.bvec_bval_tools import fsl2mrtrix, identify_shells
 
@@ -47,7 +49,7 @@ def _build_arg_parser():
     p.add_argument('in_bvec',
                    help='b-vectors filename, in FSL format (.bvec).')
 
-    p.add_argument('--in_mask',
+    p.add_argument('--mask',
                    help='Brain mask filename.')
     p.add_argument('--out_dir', default="results",
                    help='Output directory for the NODDI results. '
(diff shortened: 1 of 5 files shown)


scripts/tests/test_compute_freewater.py
```diff
@@ -24,11 +24,11 @@ def test_execution_commit_amico(script_runner):
...
                            'dwi.bvec')
-    in_mask = os.path.join(get_home(), 'commit_amico',
-                           'mask.nii.gz')
+    mask = os.path.join(get_home(), 'commit_amico',
+                        'mask.nii.gz')
     ret = script_runner.run('scil_compute_freewater.py', in_dwi,
-                            in_bval, in_bvec, '--in_mask', in_mask,
-                            '--out_dir', 'freewater/', '--b_thr', '30',
+                            in_bval, in_bvec, '--mask', mask,
+                            '--out_dir', 'freewater', '--b_thr', '30',
```
jchoude, line 31: Here, you removed the `/` of the output dir, but you didn't change it in the noddi test script. Normal?

scripts/scil_compute_freewater.py
```diff
@@ -47,11 +48,11 @@ def _build_arg_parser():
     p.add_argument('in_bvec',
                    help='b-vectors filename, in FSL format (.bvec).')
 
-    p.add_argument('--in_mask',
+    p.add_argument('--mask',
                    help='Brain mask filename.')
     p.add_argument('--out_dir', default="results",
                    help='Output directory for the Free Water results. '
-                        '[current_directory]')
+                        '[%(default)s]')
```
jchoude, line 55: You told me by chat that if the output directory is `./` the script will remove everything in the input dir, which of course is bad. Could you add a check to make sure output_dir is not the current directory?

scripts/scil_compute_NODDI.py
```diff
@@ -95,9 +96,16 @@ def main():
...
 
     assert_inputs_exist(parser, [args.in_dwi, args.in_bval, args.in_bvec],
-                        args.in_mask)
+                        args.mask)
+
+    basic_out_files = ['FIT_dir.nii.gz', 'FIT_ICVF.nii.gz',
+                       'FIT_ISOVF.nii.gz', 'FIT_OD.nii.gz']
+    out_files = [os.path.join(args.out_dir, f) for f in basic_out_files]
+
+    assert_outputs_exist(parser, args, out_files, check_dir_exists=False)
```
jchoude, line 105: Same question as in the FW script.

scripts/scil_compute_NODDI.py
```diff
@@ -47,7 +48,7 @@ def _build_arg_parser():
     p.add_argument('in_bvec',
                    help='b-vectors filename, in FSL format (.bvec).')
 
-    p.add_argument('--in_mask',
+    p.add_argument('--mask',
                    help='Brain mask filename.')
     p.add_argument('--out_dir', default="results",
                    help='Output directory for the NODDI results. '
```
jchoude, line 54: Add the display of the default in the help. And possible add the check as suggested in FW (about `./`).

frheault:
Just to be sure, both scripts have a results folder as default, do we want results_freewater and results_noddi to avoid overlap? (I really don't care, just an observation).

Both are working, I will approve after you do the check if --out_dir == './'

arnaudbore:
> Just to be sure, both scripts have a results folder as default, do we want results_freewater and results_noddi to avoid overlap? (I really don't care, just an observation).
I already had this discussion with @jchoude and he told me this was not suppose to happen.
> 
> Both are working, I will approve after you do the check if --out_dir == './'

doing the `assert_output_dirs_exist_and_empty` already does the trick here. I even change the message saying it could potentially delete some of the data if you are using this folder.

frheault:
I would personally not allow to use `'./'`, but yeah we can leave some responsibility to the user

jchoude:
As for having different defaults, I don't think that's absolutely needed... In most cases, people will either use a pipeline where this is already implicitly the case, or will use a non-custom out dir.

For the `./` check, it indirectly does it in the check that Arnaud mentioned. Would you still prefer an explicit check with the `pwd` to make sure?

frheault:
I fine with this as is, I will approve

frheault (approved):

arnaudbore:
@jchoude do you approve and merge ?

jchoude (approved):

PR #337, opened 2020-10-13
Chord chart connectivity

frheault:
Added options for visualization of connectivity, chord chart!
This is very hard to do in matplotlib (or even in fancy library like Plotly), I used some code I found online and modified it to our need.

This option works best with matrix containing around 25-250 connections, more than that and it requires heavy fine-tunning for it to look nice (likely not possible with the scripts).

It also has some difficulties if the biggest connection is 10x higher values than the second or weird extreme distribution.
Otherwise it works pretty well.

Only makes sense for streamlines count and volume matrix, does not make sense with length or FA.

sc_vol_normalized_chord-min

https://drive.google.com/file/d/1Zka8-xmb1xwJGT7QRW4NRO3LaQIkk3W6/view?usp=sharing

### DIFF
--- a/scilpy/viz/chord_chart.py
+++ b/scilpy/viz/chord_chart.py
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
 
+"""
+Most code here is from https://github.com/fengwangPhysics/matplotlib-chord-diagram.
+It was adapted to our specific needs: size of matrix, order of magnitude of
+max/min values, alpha for visualisation, etc.
+"""
+
 import math
 import matplotlib
 from matplotlib.path import Path
@@ -8,6 +14,20 @@ import numpy as np
 
 
 def polar2xy(r, theta):
+    """
+    Plot each shell
+
+    Parameters
+    ----------
+    r: float
+        The norm/radius
+    theta: float
+        The angle in radian
+    Return
+    ------
+    numpy.ndarray
+        the x/y coordinates
+    """
     return np.array([r*np.cos(theta), r*np.sin(theta)])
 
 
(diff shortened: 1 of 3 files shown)


scilpy/viz/chord_chart.py
```diff
@@ -0,0 +1,235 @@
...
+
+import math
+import matplotlib
+from matplotlib.path import Path
+import matplotlib.patches as patches
+import numpy as np
+
+
+def polar2xy(r, theta):
+    return np.array([r*np.cos(theta), r*np.sin(theta)])
```
arnaudbore, line 11: Please add/update docstring parameters and return

arnaudbore (commented):
Fancy visualisation !
Please add docstring.
There is some issues with your example (some files are missing). 
Something I may be missing, the histogram is weird to me. What you called normalised histogram should it be between 0 and 1 ? Is there a way to add some legend as well ?

scripts/scil_visualize_connectivity.py
```diff
@@ -74,6 +78,25 @@ def _build_arg_parser():
     histo.add_argument('--exclude_zeros', action='store_true',
                        help='Exclude the zeros from the histogram.')
 
+    chord = p.add_argument_group(title='Chord chart options')
+    chord.add_argument('--chord_chart', metavar='FILENAME',
+                       help='Compute and display/save a chord chart of weigth.')
+    chord.add_argument('--percentile_threshold', type=int, default=0,
+                       help='Discard connections below that percentile.'
+                            '[%(default)s]')
```
mdesco, line 86: Not sure what this percentile represents...

mdesco (approved):
Once minor comments addressed, everything looks good.

frheault:
@arnaudbore I added the docstring to the simple function (polar coord) that could be useful to someone else.

I didn't add docstring to the very specific function borrowed from the net, Laurent would need that script ASAP (I slowed him down because I moved)

arnaudbore (approved):
Go LGTM !

PR #342, opened 2020-10-30
add compute only option

arnaudbore:
scil_compute_NODDI.py
scil_compute_freewater.py


mdesco (approved):
Tested and works fine. Thanks!

PR #347, opened 2020-11-02
scil connectivity Normalization fix by zero labels

mdesco:
Sami had 4 subjects with label 90 with 0 voxels in them after warping to MNI.
Maybe sign of an issue but regardless, we should deal with the case label is empty. The volume normalization was doing this check but not the surface normalization.

This should fix it.

Data to test: https://www.dropbox.com/s/kuwb5lhs8339x5v/normalization_fix.zip?dl=0
Command ran:
scil_normalize_connectivity.py sc.npy sc_surface_normalized.npy --parcel_surface atlas_brainnetome_v2_mni.nii.gz brainnetome_mni.txt

### DIFF
--- a/scripts/scil_normalize_connectivity.py
+++ b/scripts/scil_normalize_connectivity.py
@@ -139,7 +139,7 @@ def main():
 
         voxels_vol = np.prod(atlas_img.header.get_zooms()[:3])
         voxels_sur = np.prod(atlas_img.header.get_zooms()[:2])
-
+        
         # Excluding background (0)
         labels_list = np.loadtxt(labels_filepath)
         if len(labels_list) != in_matrix.shape[0] \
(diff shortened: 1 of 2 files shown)


arnaudbore (approved):

scripts/scil_normalize_connectivity.py
```diff
@@ -174,8 +176,11 @@ def main():
                 factor_list.append(np.count_nonzero(
                     atlas_data == label) * voxels_vol)
             else:
-                factor_list.append(approximate_surface_node(
-                    atlas_data, label) * voxels_sur)
+                if np.count_nonzero(atlas_data == label):
+                    factor_list.append(approximate_surface_node(
+                        atlas_data, label) * voxels_sur)
+                else:
+                    factor_list.append(0)
```
frheault, line 183: Is not that going to divide by 0 right after? Which will put an infinite value in the matrix, which will be hard to visualize?

If possible, later I would catch the 0 and put -1 in the matrix. Which will be easy to spot while not screwing up the vizualisation

arnaudbore: it's divided by >0.001 do you still want to put -1 ?

mdesco: In fact, the behavior should be the same at the volume normalization, no? @frheault you were already doing the np.nonzero_count for the above if. We should have the same behavior for both.

frheault: Ok when I looked at the raw file I didn't see a few line. Forget about my comments... SOrry

PR #352, opened 2020-11-05
fix too fast merge of normalise_connectivity

arnaudbore:
fix if no matrix operation


mdesco (approved):
Now it works and properly tested by me..

PR #357, opened 2020-11-29
changing 20 to 100

mdesco:
In the HCP7T, b=0 image is between 55 and 65 in the dataset. It is a b0 image.
I relaxed the check a bit. Should not affect the rest.


mdesco:
En passant, je vois que nos autres scripts ont un truc comme --force_b0_threshold qui raise le warning mais qui passe quand meme... 

scil_compute_ssst_frf.py dwi.nii.gz bval bvec frf.txt --force_b0_threshold

arnaudbore:
Fix with #358

PR #359, opened 2020-12-08
New tractogram scoring script

alexjoa:
scil_score_tractogram is a tractogram scoring script written by @frheault , modified and mainly tested by myself @alexjoa . Scores input tractogram overall and bundlewise. Outputs a results.json containing a full report and splits the input tractogram into resulting .trk : *_tc.trk, *_fc.trk, nc.trk and *_wpc.trk, where * is the current bundle. Definitions: tc: true connections, streamlines joining a correct combination of ROIs. fc: false connections, streamlines joining an incorrect combination of ROIs. nc: no connections, streamlines not joining two ROIs. wpc: wrong path connections, streamlines that go outside of the ground truth mask, joining a correct combination of ROIs. Bundle overlap : ground truth voxels containing tc streamline(s). Either input gt_endpoints or gt_heads and gt_tails. Ground truth and ROIs must be in the same order i.e. groundTruth1.nii.gz ... groundTruthN.nii.gz --gt_tails tail1.nii.gz ... tailN.nii.gz --gt_heads head1.nii.gz ... headN.nii.gz.

### DIFF
--- a/scripts/scil_score_tractogram.py
+++ b/scripts/scil_score_tractogram.py
@@ -3,7 +3,7 @@
 
 import argparse
 
-from dipy.io.stateful_tractogram import Space, StatefulTractogram
+from dipy.io.stateful_tractogram import Space, Origin, StatefulTractogram
 from dipy.io.streamline import save_tractogram
 
 from scilpy.io.streamlines import load_tractogram_with_reference
@@ -25,7 +25,7 @@ from sklearn.cluster import KMeans
 import json
 from dipy.tracking.utils import length
 from scilpy.utils.streamlines import (perform_streamlines_operation,
-                                      subtraction)
+                                      difference)
 from scilpy.tractanalysis.features import remove_loops_and_sharp_turns
 import os
 from scilpy.utils.filenames import split_name_with_nii
@@ -51,24 +51,21 @@ def _build_arg_parser():
     p.add_argument('--gt_heads', nargs='+',
                    help='')
 
-    p.add_argument('--dilate_endpoints', metavar='NB_PASS',
-                   help='heuristic')
+    # p.add_argument('--dilate_endpoints', metavar='NB_PASS',
+    #                help='heuristic')
 
     p.add_argument('--gt_config', metavar='FILE',
                    help='heuristic')
-    p.add_argument('--output', default='./',
+    p.add_argument('--out_dir', default='gt_out/',
                    help='heuristic')
 
     p.add_argument('--wrong_path_as_separate', action='store_true',
                    help='heuristic')
 
-    # p.add_argument('--quantify_endpoints', action='store_true',
-    #                help='heuristic')
-
     add_reference_arg(p)
     add_overwrite_arg(p)
-    add_verbose_arg(p)
     add_json_args(p)
+
     return p
 
 
@@ -95,11 +92,14 @@ def get_binary_maps(streamlines, dimensions, sft):
         return np.zeros(dimensions), np.zeros(dimensions)
     elif len(streamlines) == 1:
         streamlines = [streamlines]
-    tmp_sft = StatefulTractogram(streamlines, sft, Space.RASMM)
+    tmp_sft = StatefulTractogram.from_sft(streamlines, sft)
     tmp_sft.to_vox()
     tmp_sft.to_corner()
     tmp_sft.remove_invalid_streamlines()
-    
+
+    if len(tmp_sft) == 1:
+        return np.zeros(dimensions), np.zeros(dimensions)
+
     bundles_voxels = compute_tract_counts_map(tmp_sft.streamlines,
                                               dimensions).astype(np.int16)
 
@@ -115,12 +115,19 @@ def get_binary_maps(streamlines, dimensions, sft):
 def main():
     parser = _build_arg_parser()
     args = parser.parse_args()
-    assert_output_dirs_exist_and_empty(parser, args, args.output)
+    assert_output_dirs_exist_and_empty(parser, args, args.out_dir,
+                                       create_dir=True)
+
+    if (args.gt_tails and not args.gt_heads) \
+            or (args.gt_heads and not args.gt_tails):
+        parser.error('')
+    if not args.gt_endpoints and (not args.gt_tails and not args.gt_heads):
+        parser.error('')
 
     sft = load_tractogram_with_reference(parser, args, args.in_tractogram)
-
     gt_bundle_masks = []
     gt_bundle_inv_masks = []
+
     for gt_bundle in args.gt_bundles:
 
         # Support ground truth as streamlines or masks
@@ -136,7 +143,7 @@ def main():
             gt_sft.remove_invalid_streamlines()
             gt_sft.to_vox()
             gt_sft.to_corner()
-            affine, dimensions, _, _ = gt_sft.space_attribute
+            affine, dimensions, _, _ = gt_sft.space_attributes
             gt_mask = compute_tract_counts_map(gt_sft.streamlines,
                                                dimensions).astype(np.int16)
         gt_inv_mask = np.zeros(dimensions, dtype=np.int16)
@@ -157,9 +164,9 @@ def main():
             basename = os.path.basename(
                 split_name_with_nii(mask_filename)[0])
             tail_filename = os.path.join(
-                args.output, '{}_tail.nii.gz'.format(basename))
+                args.out_dir, '{}_tail.nii.gz'.format(basename))
             head_filename = os.path.join(
-                args.output, '{}_head.nii.gz'.format(basename))
+                args.out_dir, '{}_head.nii.gz'.format(basename))
             nib.save(nib.Nifti1Image(head, affine), head_filename)
             nib.save(nib.Nifti1Image(tail, affine), tail_filename)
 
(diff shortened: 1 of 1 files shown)


alexjoa:
@mdesco @frheault @ppoulin91 can you guys have a look, thanks.

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+    gt_bundle_inv_masks = []
+
+    for gt_bundle in args.gt_bundles:
+
+        # Support ground truth as streamlines or masks
+        # Will be converted to binary masks immediately
+        _, ext = split_name_with_nii(gt_bundle)
+        if ext in ['.gz', '.nii.gz']:
+            gt_img = nib.load(gt_bundle)
+            gt_mask = gt_img.get_fdata().astype(np.int16)
```
arnaudbore, line 189: use get_data_as_mask or as_label

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+                   help="Bundles ground truth(.trk, .nii or .nii.gz).")
+    p.add_argument('--gt_endpoints', nargs='+',
+                   help="Bundles endpoints, both bundle's ROIs\
+                       (.nii or .nii.gz).")
+    p.add_argument('--gt_tails', nargs='+',
+                   help="Bundles tails, bundle's first ROI(.nii or .nii.gz).")
+    p.add_argument('--gt_heads', nargs='+',
+                   help="Bundles heads, bundle's second ROI(.nii or .nii.gz).")
+
+    # p.add_argument('--dilate_endpoints', metavar='NB_PASS',
```
arnaudbore, line 71: what do you want to do with that  ?

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+    roi2 = roi_2.get_fdata(dtype=np.float64)
+
+    rois = [roi1, roi2]
+    overlap = intersection(rois, roi_1)
+    nb_voxels = np.count_nonzero(overlap)
+
+    if nb_voxels > 0:
+        overlapping_roi.append((mask_1, mask_2))
+        overlapping_roi.append((mask_2, mask_1))
+
```
arnaudbore, line 141: return missing?

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+
+def remove_duplicate_streamlines(sft, fc_streamlines, roi1_name, roi2_name):
+    roi1 = nib.load(roi1_name).get_fdata().astype(np.int16)
```
arnaudbore, line 144: get_data_as_label or as_mask

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+
+    return mask_1, mask_2
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+    assert_output_dirs_exist_and_empty(parser, args, args.out_dir,
+                                       create_dir=True)
+
```
arnaudbore, line 171: assert_inputs_exist missing

arnaudbore (commented):
First round of comments:
Please add a test, if possible use zip already available if possible.
Please add some data to test with command line.
Great script !

alexjoa:
[pr_test_data.zip](https://github.com/scilus/scilpy/files/5680379/pr_test_data.zip)

Command line:
unzip pr_test_data
cd {your_data_directory}/pr_test_data

scil_score_tractogram.py 100307_concatenate_res.trk ground_truth/gt_AF_L/groundTruth_AF_L.trk ground_truth/gt_AF_R/groundTruth_AF_R.trk ground_truth/gt_CC/groundTruth_CC.trk ground_truth/gt_PT_L/groundTruth_PT_L.trk ground_truth/gt_PT_R/groundTruth_PT_R.trk --gt_tails ground_truth/gt_AF_L/tail_AF_L.nii.gz ground_truth/gt_AF_R/tail_AF_R.nii.gz ground_truth/gt_CC/tail_CC.nii.gz ground_truth/gt_PT_L/tail_PT_L.nii.gz ground_truth/gt_PT_R/tail_PT_R.nii.gz --gt_heads ground_truth/gt_AF_L/head_AF_L.nii.gz ground_truth/gt_AF_R/head_AF_R.nii.gz ground_truth/gt_CC/head_CC.nii.gz ground_truth/gt_PT_L/head_PT_L.nii.gz ground_truth/gt_PT_R/head_PT_R.nii.gz --out_dir score --wrong_path_as_separate -f

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+    fused_masks[fused_masks > 0] = 1
+
+    indices = np.unique(indices).tolist()
+    no_conn_streamlines = []
+    no_conn_streamlines.extend(sft.streamlines[indices])
+
+    overlapping_roi = []
+    all_rois = args.gt_heads + args.gt_tails
+    for roi1, roi2 in itertools.combinations(all_rois, 2):
+        identify_overlapping_roi(roi1, roi2, overlapping_roi)
```
AntoineTheb, line 260: Adding on @arnaudbore's comment elsewhere, it'd be less confusing if the function had an explicit return instead of using side-effects

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+    fc_streamlines_list = []
+    wpc_streamlines_list = []
+
+    # Again keep the keep the correct combinations
+    comb_filename = list(itertools.combinations(
+        itertools.chain(*zip(args.gt_tails, args.gt_heads)), r=2))
+    comb_data = list(itertools.combinations(all_data, r=2))
+
+    # Go through all the possible combinations of endpoints masks
+    for i, roi in enumerate(comb_data):
```
AntoineTheb, line 276: This for-loop is ~125 lines long. Without falling into "Clean Code" levels of abstractions and function segmentation, do you think you could break up some of the logic into separate functions ?  It would add to code reusability (if ever another script needs to reuse the application logic written below) and readability (the `main` function ends up being ~400 lines long).

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+            [x for x in wpc_streamlines_list if x])
+
+    final_results['total_streamlines'] = total_count
+    final_results["bundle_wise"] = {}
+    final_results["bundle_wise"]["true_connections"] = {}
+    final_results["bundle_wise"]["false_connections"] = {}
+    tractogram_overlap = 0.0
+
+    # Bundle-wise statistics, useful for more complex phantom
+    for i, filename in enumerate(comb_filename):
```
AntoineTheb, line 468: Same as above; if possible, I would try to wrap up "application logic" in separate functions (and files) to promote reusability, and keep the `main()`'s role to I/O.

AntoineTheb (changes requested):
@alexjoa I think you forgot to add the `trk`s to the testing data. Or maybe the usage example above could be adapted so it uses nifitis instead ? Also, I just ran it (with niftis) and the results are quite bad, is it normal ? Do you think you could provide testing data that is more representative ? I will run it with the ISMRM2015 dataset and report back.

Concerning the code, I left a few comments. Mostly, though, I think that most of the logic could be broken up into separate functions/files. I'm really not a fan of having monolithic `main()`s: It hampers readability and code-reusability. I much prefer removing as much logic as possible from the `main` and leaving only I/O stuff, if possible. Also, would it be possible to add a few logs (displayed with a `verbose` option) ? It'd be more user-friendly this way.

Otherwise, cool stuff and congrats !

scripts/scil_score_tractogram.py
AntoineTheb: I think the function uses side-effets instead of an explicit return, as `append` happens in-place.

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+import os
+import pprint
+from sklearn.cluster import KMeans
+
+from scilpy.image.operations import intersection, difference
+from scilpy.io.streamlines import load_tractogram_with_reference
+from scilpy.io.utils import (add_overwrite_arg,
+                             assert_inputs_exist,
+                             assert_outputs_exist,
+                             add_verbose_arg,
```
AntoineTheb, line 24: I would actually use the verbose flag and print logs as files get loaded and the processing goes along. For large dataset, the script takes a while and with no indication on how it's going, it can be a bit confusing.

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,544 @@
...
+
+    final_results['total_streamlines'] = total_count
+    final_results["bundle_wise"] = {}
+    final_results["bundle_wise"]["true_connections"] = {}
+    final_results["bundle_wise"]["false_connections"] = {}
+    tractogram_overlap = 0.0
+
+    # Bundle-wise statistics, useful for more complex phantom
+    for i, filename in enumerate(comb_filename):
+        timer = time()
```
AntoineTheb, line 469: Unused variable

AntoineTheb (changes requested):
Some more comments

scripts/scil_score_tractogram.py
alexjoa: @AntoineTheb Where/which scilpy script can I find an example of this? thanks

AntoineTheb: `scil_recognize_single_bundle` has some form of logging in the VotingScheme class. `scil_decompose_connectivity.py` ouputs messages as the processing goess along.

AntoineTheb:
I forgot to present an update but I tried to use the script to score the ISMRM2015 dataset, but after running it for 8 hours on my laptop and 4 on my lab PC,I ran out of ram everytime.
 
Perhaps some memory management would be in order ? Something like lazily-loading only some of the streamlines at the same time, for example.  

Here's some data to test: https://drive.google.com/file/d/1RUS4qsxssa16Gyr7oJMgbMBpH1tBCWEz/view?usp=sharing
Head and tail rois can be selected with wildcards.

mdesco:
@alexjoa will you be able to put some time on this so that we can merge this PR in the future?
Thanks

PR #365, opened 2020-12-21
FIX: Update fury to 0.6.*

AntoineTheb:
Self explanatory.

You can test that fury still works and displays streamlines by running `scil_visualize_bundle_mosaic.py` with the data here:
https://drive.google.com/drive/folders/1qgnXRayGJGPDOwuzqeC6clZWbOgc2b8x?usp=sharing

### DIFF
--- a/scripts/scil_visualize_seeds.py
+++ b/scripts/scil_visualize_seeds.py
@@ -46,7 +46,7 @@ def main():
         raise ValueError("Invalid input streamline file format " +
                          "(must be trk): {0}".format(args.tractogram_filename))
 
-    # Load files and data. TRKs can have 'same' as reference
+    # Load files and data. TRKs can have 'same' as refescenece
     tractogram = load_tractogram(args.tractogram, 'same')
     # Streamlines are saved in RASMM but seeds are saved in VOX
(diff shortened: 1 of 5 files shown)


scripts/scil_visualize_seeds.py
```diff
@@ -46,7 +46,7 @@ def main():
         raise ValueError("Invalid input streamline file format " +
                          "(must be trk): {0}".format(args.tractogram_filename))
 
-    # Load files and data. TRKs can have 'same' as reference
+    # Load files and data. TRKs can have 'same' as refescenece
```
CHrlS98, line 49: Typo, right?

CHrlS98 (commented):
Except for the typo, it all looks good to me!

scripts/scil_visualize_seeds.py
frheault: If it is about the Mediterranean recluse spider, it is rufescens, not refescenece ;)

AntoineTheb: Ha ! My regex might have been too permissive. I'll fix that asap.

arnaudbore (approved):

PR #366, opened 2020-12-31
ENH: Add ratio to pairwise_agreement

AntoineTheb:
Add tractometer-style overlap and overreach output as a ratio over the "gold-standard" instead of absolute values.

@arnaudbore @frheault @GuillaumeTh


frheault (approved):
I assume you ran pytest

AntoineTheb:
@frheault I did, and even *added some tests*

arnaudbore (approved):

PR #367, opened 2021-01-05
Resample bvals

mdesco:
Small script to create a new bval text file with rounded values according to the tolerance factor. 
Developed specifically for hcp_7T data but could be useful for other datasets.

### DIFF
--- a/scripts/scil_resample_bvals.py
+++ b/scripts/scil_resample_bvals.py
@@ -33,6 +33,9 @@ def _build_arg_parser():
     parser.add_argument('bvals',
                         help='The b-values in FSL format.')
 
+    parser.add_argument('bvecs',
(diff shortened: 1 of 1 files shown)


frheault (commented):
I see a small/big problem, this script assumes the ordering found by identify_shells() will be the same as the one provided by the user in bvals_to_extract.

If the bvals of the files goes 0 2000 1000 3000 [...] and the user uses --bvals_to_extract 0 1000 2000 3000 it will switch the values accidentally.

It would be important to fix this, either by ordering centroids (and swapping shell_indices accordingly)
OR
At least verify that np.abs(centroids[i]- bvals_to_extract[i]) < to (and crash if not)

Solution #1 is more robust.

PR #369, opened 2021-01-06
Update to save RAM and parameters optims

frheault:
I realized I was generating all of thousands of parameters combinations AND including the models bundle in it. Meaning that if you run 18 parameters * 5 models * 2 seeds you are copying the model bundles that many times in memory before launching sub-processes. Oversight on my part, also I realized that copies (as many as # thread) of QBX indices was kept during the sub-processes despite not being needed for the whole time, so now it is deleted.

These two additions save quite a lot of RAM, enough that now I am fairly sure RBx can be run on the 192Gb node at 40 CPU !

Other tiny fixes on the side, I can explain if needed.

### DIFF
--- a/scilpy/io/image.py
+++ b/scilpy/io/image.py
@@ -61,6 +61,8 @@ def get_data_as_mask(in_img, dtype=np.uint8):
                 logging.warning('The two unique values in mask were not 0 and'
                                 ' 1. Tha mask has been binarised.')
                 data[data != 0] = 1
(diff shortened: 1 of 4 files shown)


mdesco (approved):
This works very well. Tested on single beluga node with 185G and 40 CPUs.
As well as my office linux desktop. Ready to be merged! Thanks!

arnaudbore (approved):

PR #370, opened 2021-01-11
Infinite endpoints computation

frheault:
Change datatype of mask to uint8,
faster saving due to endpoints compuation using (get data as mask in loop)


mdesco (approved):
Tested. Good to go.

AntoineTheb (approved):
Tested with the fibercup, LGTM

arnaudbore (approved):

Issue #361, opened 2020-12-11
pairwise bundle adjacency of two bundle always higher than one


MengxingLiu:
Hi,

I have been checking the bundle adjacency of two bundles from test-retest, no matter how the two bundles are close to each other, none of the values fall between 0 and 1, it's always equal to or higher to 1. 
Here is a result
https://drive.google.com/file/d/1VZL5w8Y66dn21P72n_sSeRLQd-1k8EKp/view?usp=sharing
the minimum value is exactly 1.0, which makes me wonder if there is a systematic reason that the optimal value of bundle_adjacency_voxel is one?  Although I tried to compute two identical bundle and it gave me a value of zero.

FYI, when I'm computing the bundle_adjacency_voxel, the reference I passed to it is a T1 image with 1mm isotropic voxel size.

frheault:
Hi, this could indeed be made clearer. And there is maybe an inconsistency in the code I will check.
I think that right now the bundle_adjacency_voxel is in relation to the voxel_size, so not in mm directly. I will soon update the code to make it clearly in mm.

About the distance unit, this is a good observation. It came from a decision in a previous paper of mine where I was using dice_voxel AND bundle_adjacency_voxel at the same time. The goal of dice was to quantify overlapping voxel and the bundle_adjacency_voxel was to quantify voxels that do not overlap (basically excluding zeros).
So it is "normal" (but not intuitive at first) that bundle_adjacency_voxel is always above 1mm (smallest distance possible 1 voxel, or 1mm).

In order to make this more intuitive in the future (since the current configuration was fine tune for my own use) I will change the code so it does include overlap. 

In the meantime, if you want to interpret your results you can look at the dice_voxel and bundle_adjacency_voxel. Both values together will tell you how much overlap is achieved and how close to overlapping regions are the non-overlapping voxels.

arnaudbore:
@MengxingLiu 
This new feature has been merged.

PR #373, opened 2021-01-13
Optional overlap in pairwise measures

frheault:
As mentioned in #361 the current behavior was not intuitive and maybe oriented too much toward my own use (from Tractostorm). So I add a parameter to allow to switch from one the other directly in the script.


AntoineTheb (approved):
Nothing too fancy, LGTM

arnaudbore (approved):

PR #374, opened 2021-01-18
FIX: Add warning to reference and .trk

AntoineTheb:
Add a warning when loading a .trk with a reference that the reference will be discarded. `scilpy.io.streamlines.load_tractogram_with_reference` was pulling a sneaky and discarding the reference without telling the user if the tractogram was a .trk. Atleast the user should be told that the reference is discarded.

@frheault @GuillaumeTh @arnaudbore

### DIFF
--- a/scilpy/io/streamlines.py
+++ b/scilpy/io/streamlines.py
@@ -85,7 +85,10 @@ def load_tractogram_with_reference(parser, args, filepath,
 
     _, ext = os.path.splitext(filepath)
     if ext == '.trk':
-        if args.reference:
+        if (
+            args.reference or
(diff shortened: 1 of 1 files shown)


frheault (approved):

arnaudbore:
I would put this warning only if reference has been set otherwise it means that the user knew about it.

AntoineTheb:
Good point @arnaudbore, fixed.

AntoineTheb:
Some tests are failing, investigating. I'll update when they're fixed.

AntoineTheb:
@arnaudbore @frheault @GuillaumeTh ready for review

arnaudbore (approved):

Issue #379, opened 2021-01-25
Trouble with scil_compute_bundle_profiles.py


Rad-dude:
Dear Scilpy developers,

First of all, thanks a lot for making the code openly available.

I am getting an error and failed execution of scil_compute_bundle_profiles.py with tck inputs.

Here's my command:
scil_compute_bundle_profiles.py $(pwd)/sub-PT_TCKs_output/PyT_all_LT_output/PyT_all_LT_filt3_BT.tck $(pwd)/fa_reg2T1w.nii --reference $(pwd)/fa_reg2T1w.nii --in_centroid $(pwd)/trial_centroid.tck

Here's the error message:
Traceback (most recent call last):
  File "/path_to/opt/anaconda3/bin/scil_compute_bundle_profiles.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/path_to/scilpy/scripts/scil_compute_bundle_profiles.py", line 131, in <module>
    main()
  File "/path_to/scilpy/scripts/scil_compute_bundle_profiles.py", line 74, in main
    assert_same_resolution(args.in_metrics + [args.in_bundle])
  File "/path_to/scilpy/scilpy/io/image.py", line 25, in assert_same_resolution
    if not is_header_compatible(images[0], curr_image):
  File "path_to/opt/anaconda3/lib/python3.8/site-packages/dipy/io/utils.py", line 323, in is_header_compatible
    affine_2, dimensions_2, voxel_sizes_2, voxel_order_2 = get_reference_info(
  File "path_to/opt/anaconda3/lib/python3.8/site-packages/dipy/io/utils.py", line 292, in get_reference_info
    raise TypeError('Input reference is not one of the supported format')
TypeError: Input reference is not one of the supported format

Scilpy is installed under anaconda3 on Mac OSX 10.13.6.
Other scil scripts are working fine, for example:
scil_compute_centroid.py --reference ./sub-PT_prep/sub-PT_T1brain_MSinFA_Warped.nii.gz ./sub-PT_TCKs_output/PyT_all_LT_output/PyT_all_LT_filt3_BT.tck ./trial_centroid.tck

This makes me think that the problem is not with the environment, nor with the reference image itself.

Any help would be appreciated.

Best,
Ahmed Radwan

AntoineTheb:
Hi @Rad-dude, thanks for submitting this issue ! 

[...]

PR #382, opened 2021-01-26
FIX: tck as sft for header check

AntoineTheb:
Address #379.

`dipy.io.utils.is_header_compatible` does not handle `.tck`s. As such I went through scripts to make sure none passed `.tck` filenames to this function and loaded them as `StatefulTractogram` if they could have.

@frheault @arnaudbore @GuillaumeTh 

~Tests are still running, will update once it's done~ All tests have passed


scripts/scil_label_and_distance_maps.py
```diff
@@ -51,11 +51,11 @@ def main():
     assert_outputs_exist(parser, args, [args.output_label,
                                         args.output_distance])
 
-    is_header_compatible(args.in_bundle, args.in_centroid)
-
     sft_bundle = load_tractogram_with_reference(parser, args,
                                                 args.in_bundle)
 
+    is_header_compatible(sft_bundle, args.in_centroid)
```
arnaudbore, line 57: here args.in_centroid can also be a tck right ?

arnaudbore (approved):
One last fix but otherwise LGTM.

scripts/scil_label_and_distance_maps.py
AntoineTheb: Indeed ! Fixed in 32dfe8f

arnaudbore (approved):
LGTM

arnaudbore:
Quick discussion with Francois.

We should have a custom: check_compatibility_tractogram_with_reference()
We should check the extension before going for a sft.

AntoineTheb:
Well if the idea is to have a cleaner solution, I thought about writing a wrapper around dipy's `is_header_compatible` to make sure that only tcks would be loaded for a header check, and that they would be loaded with their reference. While it seems like a simple enough idea, you may notice that there are many more references to `is_header_compatible` in scilpy than what was modified here. This is because most scripts didn't need this added check, and so the added function would only save a few lines at best. So it would require a lot of work, potentially breaking many things, without too much added value.

If the idea was to save computation, you may notice that most scripts would have loaded the tractograms right after the check anyways, so it wouldn't save that much ressources.

All in all, I agree it would be cleaner to have a function like you're proposing, but I think it is out of scope for this PR (which let's remember addresses a bug that a user encountered).

arnaudbore (approved):

PR #383, opened 2021-01-27
[WIP] update dipy

arnaudbore:


CHrlS98:
Probably worth noting,
Using the option `--full_basis` in `scil_visualize_fodf.py` will crash with error `Invalid basis name.` when calling `sh_to_sf` because full SH bases are accessed by specifying `full_basis=True` rather than by appending `_full` to the basis name since DIPY 1.3. The exact line that will fail is https://github.com/scilus/scilpy/blob/2086275493349a396754eab6c28790c394b26c49/scilpy/viz/scene_utils.py#L91
The same problem could happen with `scil_denoise_sf.py` from PR #368 if it were merge before this one.

Also, there will probably be some deprecation warnings when using `sh_to_sf`, `sh_to_sf_matrix`, etc. because bases definitions have been updated (https://github.com/dipy/dipy/pull/2206).

PR #388, opened 2021-02-01
FIX: `window.snapshot` flips images

### LINKED ISSUE #386: Trouble with scil_visualize_bundles_mosaic.py
Greetings,

I was trying to use this function as follows:
scil_visualize_bundles_mosaic.py ./sub-x_prep/fa.nii.gz ./AF_all_LT_fin_WB_iFOD2.tck ./trial.png

But I keep getting this error, might this be related to the Statefultractogram issue again?

  File "/home/rad/anaconda3/bin/scil_visualize_bundles_mosaic.py", line 7, in <module>
  File "/NI_apps/scilpy/scripts/scil_visualize_bundles_mosaic.py", line 355, in <module>
    main()
  File "/NI_apps/scilpy/scripts/scil_visualize_bundles_mosaic.py", line 166, in main
    tractogram = load_tractogram_with_reference(parser, args, filename)
  File "/NI_apps/scilpy/scilpy/io/streamlines.py", line 104, in load_tractogram_with_reference
    elif args.reference is None:
AttributeError: 'Namespace' object has no attribute 'reference'

Any input would be appreciated.

Thanks in advance.

AntoineTheb:
- Un-flip the thumbnails
- Remove the need for intermediate images
- scene_utils: flip image if saved using window.snapshot

@arnaudbore @CHrlS98 @frheault @GuillaumeTh 

Fixes #386. All tests have passed


CHrlS98 (commented):
Looks all good to me!

AntoineTheb:
@arnaudbore Good to go !

scilpy/io/utils.py
```diff
@@ -494,3 +495,8 @@ def parser_color_type(arg):
         raise argparse.ArgumentTypeError(
             "Argument must be < " + str(MAX_VAL) + "and > " + str(MIN_VAL))
     return f
+
+
+def save_image(img_array, filename):
```
CHrlS98, line 500: Maybe we should add a comment saying `img_array[0, 0]` correspond to the bottom left corner of the image?

AntoineTheb: I realize now that the name of the function might be too general and misleading as it might lead someone to use it without having an image that comes from `window.snapshot`. Which would potentially flip an image that doesn't need to be flipped. I'll adapt

AntoineTheb: @CHrlS98 fixed in b2ce7f6

mdesco (approved):

AntoineTheb:
New "bundle.zip" file
https://drive.google.com/file/d/153Oj4lYcS6PTCKKi30cQxZttsesOit_o/view?usp=sharing

AntoineTheb:
Tests most likely fail because I have modified the testing data. However I can't know because I don't have access to the CI *wink wink*

arnaudbore (approved):

PR #395, opened 2021-02-05
Update jenkins files

GuillaumeTh:
I keep travis file for now to transfert the docker part later

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,6 +28,18 @@ pipeline {
                         }
                     }
                 }
+                stage('Python3.8') {
+                    steps {
+                        withPythonEnv('CPython-3.8') {
+                            sh '''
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -2,19 +2,41 @@ pipeline {
...
+                            '''
+                        }
+                    }
+                }
+                stage('Python3.7') {
+                    steps {
+                        withPythonEnv('CPython-3.7') {
+                            sh '''
+                                pip3 install numpy==1.18.* wheel
+                                pip3 install -e .
```
arnaudbore, line 24: Pourquoi python setup.py install does not work ?

GuillaumeTh: It work but in the end it do the same thing and I use more this command than setup.py

arnaudbore (approved):

Issue #408, opened 2021-02-16
scil_compute_todi.py defaults 


mdesco:
Can we please add the defaults in the -help. I do not see what the important defaults are by doing -h. This is annoying and I must go to the code.
--sphere
--sh_order
--sh_normed
--smooth

Also, not 100% sure what is the "length" weighted thingy... Should the non-length weighted be outputed too?

@StongeEtienne or @frheault  thanks

StongeEtienne:
@mdesco I'm working on it,
added the "sphere" and "sh_order" default, however "smooth" and "sh_normed" are boolean.
I can rename the "sh_normed" to "normalize_per_voxel", to be simplier.

However for "smooth" it apply a spatial blur (gaussian sigma), and a directionnal blur (dot order).
do you want me to split it into 2 separate input ?   spatial_sigma  & directionnal_smoothing_order  (0 default for no blurr).

StongeEtienne:
@mdesco *also, the "length" is to support compress streamlines (with non-uniform segment) so it weight each segment by their length.

mdesco:
Thanks for the precision. No need to split the boolean smooth option.

Is the length the length of the track or length within the voxel? Please clearly the text at the beginning of the script.

Also, please clarify the --out_lw_todi option. This is a spherical function output, with the fourth dimension as big as the number of points on the sphere. This is become crazy big output...

PR #409, opened 2021-02-17
Todi defaults

StongeEtienne:
Simplified TODI script, added details and default values in the help
 for issue #408

### DIFF
--- a/scripts/scil_compute_todi.py
+++ b/scripts/scil_compute_todi.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 """
-Compute a length-weighted Track Orientation Density Image (TODI).
-This script can afterwards output a length-weighted Track Density Image
-(TDI) or a length-weighted TODI, based on streamlines' segments.\n\n
+Compute a Track Orientation Density Image (TODI).
+Each segment of the streamlines is weighted by length,
+this is done to support compressed streamlines.
+This script can afterwards output a Track Density Image (TDI)
+or a TODI with SF or SH representation, based on streamlines' segments.\n\n
 """
 
 import argparse
@@ -33,13 +35,14 @@ def _build_arg_parser():
     p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
                                 formatter_class=argparse.RawTextHelpFormatter)
 
-    p.add_argument('in_bundle',
+    p.add_argument('input_tracto',
                    help='Input streamlines file.')
 
     add_reference_arg(p)
 
     p.add_argument('--sphere', default='repulsion724',
-                   help='sphere used for the angular discretization.')
+                   help='sphere used for the angular discretization. '
+                        '[%(default)s]')
 
     p.add_argument('--mask',
                    help='Use the given mask.')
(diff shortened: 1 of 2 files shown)


scripts/scil_compute_todi.py
```diff
@@ -33,39 +35,39 @@ def _build_arg_parser():
     p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
                                 formatter_class=argparse.RawTextHelpFormatter)
 
-    p.add_argument('in_bundle',
+    p.add_argument('input_tracto',
```
GuillaumeTh, line 38: Do you have any use case where it is not a bundle ? Moreover, the standard in scilpy is in_. Can you change it ?

scripts/scil_compute_todi.py
```diff
@@ -33,39 +35,39 @@ def _build_arg_parser():
...
-    p.add_argument('--out_lw_tdi',
-                   help='Output length-weighted TDI map.')
+    p.add_argument('--out_tdi',
+                   help='Output Track Density Image (TDI).')
 
-    p.add_argument('--out_lw_todi',
-                   help='Output length-weighted TODI map.')
+    p.add_argument('--out_todi_sf',
+                   help='Output TODI, with SF (each directions\n'
+                        'on the sphere), require a lot of memory)')
```
GuillaumeTh, line 58: The parenthesis seems to not fit.

scripts/scil_compute_todi.py
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 """
-Compute a length-weighted Track Orientation Density Image (TODI).
-This script can afterwards output a length-weighted Track Density Image
-(TDI) or a length-weighted TODI, based on streamlines' segments.\n\n
+Compute a Track Orientation Density Image (TODI).
+Each segment of the streamlines is weighted by length,
```
GuillaumeTh, line 5: Sound weird. Maybe something like this: Each segment of the streamlines is weighted by length (to support compressed streamlines)

GuillaumeTh (changes requested):

scripts/scil_compute_todi.py
mdesco: Agree with this suggestion

scripts/scil_compute_todi.py
mdesco: remove ) after sphere

require -> requires

scripts/scil_compute_todi.py
```diff
@@ -33,39 +35,39 @@ def _build_arg_parser():
...
 
-    p.add_argument('--sh_normed', action='store_true',
-                   help='Normalize SH.')
+    p.add_argument('--normalize_per_voxel', action='store_true',
+                   help='Normalize each SF or SH at each voxel.')
 
-    p.add_argument('--smooth', action='store_true',
-                   help='Smooth todi (angular and spatial).')
+    p.add_argument('--smooth_todi', action='store_true',
+                   help='Smooth TODI (angular and spatial).')
```
mdesco, line 70: please add the defaults for the boolean options too. [false]

mdesco (commented):
Minor edits from Guillaume and I, then good to go.

scripts/scil_compute_todi.py
StongeEtienne: It can be used for any tractogram, similar to a TDI map or image

scripts/scil_compute_todi.py
StongeEtienne: done

scripts/scil_compute_todi.py
StongeEtienne: done

mdesco:
@StongeEtienne  Can you please add the boolean defaults to false and then I approve. Otherwise, we cannot know they are boolean options

scripts/scil_compute_todi.py
StongeEtienne: @arnaudbore is it common to add default to "action='store_true',"

mdesco: I want the user to know it is a boolean option. Right now impossible to know with --smooth_todi needs to be called with a float or a true or false. Just find a way for the user to call it with 
--smooth_todi true 
and not --smooth_todi 0.25 or whatever

StongeEtienne: done

mdesco (approved):

GuillaumeTh (approved):

PR #419, opened 2021-02-25
[FIX] MultiVoxelFit is badly constructed after a multiprocess fit

boiteaclou:
When fitting using multiple processes with `fit_from_model`, if some data points included in the mask are empty, the `fit_from_model_parallel` will return a 0 instead of a Fit object. Subsequent calls to attributes of the MultiVoxelFit object constructed with those fits will fail, since those zeros are not excluded from the mask fed to it.

Added a masking routine to exclude those voxels from the MultiVoxelFit mask.


arnaudbore:
Can you provide some data to test this ?

arnaudbore:
Closing, it has already been fixed.

PR #420, opened 2021-02-25
[ENH] Extend extract_b0 capabilities

boiteaclou:
This PR contains modifications to :

- Extract b0 : added more extraction options, such as extracting b0s from clusters of b0s in the data (mean or first b0 of each cluster). Also reformatted the code to enable lazyloading. (Does not modify previous behavior of the script, fully back-compatible cli)
- Prepare Topup : TBD
- Prepare Eddy : TBD

### DIFF
--- a/scilpy/image/utils.py
+++ b/scilpy/image/utils.py
@@ -43,6 +43,11 @@ def volume_iterator(img, blocksize=1, start=0, end=0):
         Image of a 4D volume with shape X,Y,Z,N
     blocksize : int, optional
         Number of volumes to return in a single batch
+    start : int, optional
+        Starting iteration index in the 4D volume
+    end : int, optional
+        Stopping iteration index in the 4D volume
+        (the volume at this index is excluded)
 
     Yields
     -------
--- a/scilpy/image/utils.py
+++ b/scilpy/image/utils.py
@@ -34,7 +34,7 @@ def count_non_zero_voxels(image):
     return nb_voxels
 
 
-def volume_iterator(img, blocksize=1):
+def volume_iterator(img, blocksize=1, start=0, end=0):
     """Generator that iterates on volumes of data.
 
     Parameters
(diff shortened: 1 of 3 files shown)


boiteaclou:
First question about the prepare_topup cli. Right now, we ask for the full forward phase encoded dwi file/bval/bvec. Two options for the new script :
- Keep this behavior and add an argument to allow passing only b0s files. This however make the cli non back-compatible, but makes the most sense since the b0s to use should have been previously extracted using `scil_extract_b0`
- Replace passage of full forward phase encoded dwi data and only allow by passage of b0s. This also makes the cli non back-compatible
- Keep it as is and add the options of `scil_extract_b0` to the cli to handle b0 extraction from the full forward phase encoded dwi

arnaudbore:
Quick note: scil_prepare_topup extracts the b0s from the dwi and compute the mean and does the same for the rev_b0. My first guess would be to allow either dwi input or b0 for each Phase Encoding direction and the rest of the code should be as is. I find your modifications pretty heavy compare to what I had in mind but maybe this is mandatory. We should have a discussion with @GuillaumeTh so everybody is on the same page here before moving forward.

scil-admin:

scil-admin:

arnaudbore (approved):

scilpy/image/utils.py
```diff
@@ -34,7 +34,7 @@ def count_non_zero_voxels(image):
     return nb_voxels
 
 
-def volume_iterator(img, blocksize=1):
+def volume_iterator(img, blocksize=1, start=0, end=0):
```
GuillaumeTh, line 37: Please add documentation for start and end.

scripts/scil_extract_b0.py
```diff
@@ -39,46 +40,45 @@ def _build_arg_parser():
...
-                        'diffusion weighting.')
+                        'diffusion weighting. [%(default)s]')
 
     group = p.add_mutually_exclusive_group()
     group.add_argument('--all', action='store_true',
                        help='Extract all b0. Index number will be appended to '
                             'the output file.')
     group.add_argument('--mean', action='store_true', help='Extract mean b0.')
+    group.add_argument('--cluster-mean', action='store_true',
+                       help='Extracts mean of each continuous cluster of b0s')
```
GuillaumeTh, line 51: Extracts -> Extract and . at the end.

scripts/scil_extract_b0.py
```diff
@@ -39,46 +40,45 @@ def _build_arg_parser():
...
+    group.add_argument('--cluster-first', action='store_true',
+                       help='Extracts first b0 of each '
```
GuillaumeTh, line 53: Extracts -> Extract and . at the end.

scripts/scil_extract_b0.py
```diff
@@ -39,46 +40,45 @@ def _build_arg_parser():
...
+                            'continuous cluster of b0s')
+
+    p.add_argument('--block-size', '-s',
+                   metavar='INT', type=int,
+                   help='Loads the data using this block size. '
```
GuillaumeTh, line 58: Loads -> Load

GuillaumeTh (approved):
I just checked the code. A small comment otherwise looks good.

scil-admin:

scil-admin:

PR #427, opened 2021-03-04
Add compute_sf_from_sh.py script

ppoulin91:
- Compute SF on a sphere from an SH signal (ODF or fODF)
- Can provide the sphere as argument (currently symetric724 or repulsion724)
- Original b0s are concatenated to have a coherent volume for further analysis

### DIFF
--- a/scripts/scil_compute_sf_from_sh.py
+++ b/scripts/scil_compute_sf_from_sh.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Script to sample SF values from a Spherical Harmonics signal.
+"""
+
+import argparse
+
+from dipy.core.gradients import gradient_table
+from dipy.data import get_sphere
+from dipy.io import read_bvals_bvecs
+from dipy.reconst.shm import sh_to_sf
+import nibabel as nib
+import numpy as np
+
+from scilpy.gradientsampling.save_gradient_sampling import \
+    save_gradient_sampling_fsl
+from scilpy.io.image import get_data_as_mask
+from scilpy.io.utils import (add_overwrite_arg,
+                             add_sh_basis_args, assert_inputs_exist,
+                             assert_outputs_exist)
+
+ORDER_FROM_NCOEFFS = {1: 0, 6: 2, 15: 4, 28: 6, 45: 8}
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_sh',
+                   help='Path of the SH volume.')
+    p.add_argument('in_dwi',
+                   help='Path of the original DWI volume (to fetch b0 images).')
+    p.add_argument('in_bval',
+                   help='Path of the b-value file, in FSL format.')
+    p.add_argument('in_bvec',
+                   help='Path of the b-vector file, in FSL format.')
+
+    p.add_argument('out_sf',
+                   help='Name of the output SF file to save (bvals/bvecs will '
+                        'be automatically named).')
+
+    p.add_argument('--sphere', choices=['symmetric724', 'repulsion724'],
+                   default='symmetric724',
+                   help='Sphere to use for sampling SF. [%(default)s]')
+    add_sh_basis_args(p)
+    p.add_argument('--mask',
+                   help='Path to a binary mask.\nOnly data inside the mask '
+                        'will be used for computations and reconstruction ')
+    add_overwrite_arg(p)
+
+    return p
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, [args.in_sh])
+
+    out_bvecs = args.out_sf.replace(".nii.gz", ".bvec")
+    out_bvals = args.out_sf.replace(".nii.gz", ".bval")
+
+    assert_outputs_exist(parser, args, [args.out_sf, out_bvals, out_bvecs])
+
+    # Load SH
+    vol_sh = nib.load(args.in_sh)
+    data_sh = vol_sh.get_fdata(dtype=np.float32)
+
+    # Load DWI + gradients
+    vol_dwi = nib.load(args.in_dwi)
+    data_dwi = vol_dwi.get_fdata(dtype=np.float32)
+    bvals, bvecs = read_bvals_bvecs(args.in_bval, args.in_bvec)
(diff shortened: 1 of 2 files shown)


scripts/tests/test_compute_sf_from_sh.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_help_option(script_runner):
+    ret = script_runner.run('scil_compute_sf_from_sh.py', '--help')
+    assert ret.success
+
+
+def test_execution_processing(script_runner):
+    os.chdir(os.path.expanduser(tmp_dir.name))
+    in_sh = os.path.join(get_home(), 'processing', 'sh_1000.nii.gz')
+    in_b0 = os.path.join(get_home(), 'processing', 'fa.nii.gz')
```
ppoulin91, line 22: I'm using the FA here since there is no b0 in the test files, shouldn't matter much. Should we add it and change this?

mdesco (commented):
I recommend @CHrlS98 to review this PR. It is in close spirit to his latest: scil_execute_asymmetric_filtering.py

@ppoulin91 I would "steal" his --sphere option and mimic the options/formats please.

CHrlS98 (commented):
Alright, two small comments regarding the code!

I'm also questioning myself on the nature of the input? Is it the raw DWI signal converted to SH coefficients or some ODF functions expressed as SH coefficients?

ppoulin91:
@CHrlS98 
> I'm also questioning myself on the nature of the input? Is it the raw DWI signal converted to SH coefficients or some ODF functions expressed as SH coefficients?

In my case, a deep learning model will generate new SH images (corresponding to DWI ODFs), and we want to extract DTI/fODF metrics to compare against the original DWI images. Although it could also be used to extract SF from any SH image, so your comment about the optional b0 is totally valid.

scripts/scil_compute_sf_from_sh.py
```diff
@@ -0,0 +1,96 @@
...
+import nibabel as nib
+import numpy as np
+
+from scilpy.gradientsampling.save_gradient_sampling import \
+    save_gradient_sampling_fsl
+from scilpy.io.utils import (add_overwrite_arg,
+                             add_sh_basis_args, assert_inputs_exist,
+                             assert_outputs_exist)
+
+ORDER_FROM_NCOEFFS = {1: 0, 6: 2, 15: 4, 28: 6, 45: 8}
```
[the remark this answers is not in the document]
ppoulin91: Yes, I missed that, thanks!

scripts/scil_compute_sf_from_sh.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_sh',
+                   help='Path of the SH volume.')
+    p.add_argument('in_b0',
```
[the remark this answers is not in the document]
ppoulin91: You're right (see my other reply)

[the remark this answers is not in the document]
ppoulin91: In that case however, `in_bval` could also be optional, to be used with `in_b0` to generate a DWI like volume

mdesco (approved):
Looks and works good to me.

scripts/scil_compute_sf_from_sh.py
```diff
@@ -0,0 +1,122 @@
...
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, [args.in_sh])
+
+    out_bvecs = args.out_sf.replace(".nii.gz", ".bvec")
+    assert_outputs_exist(parser, args, [args.out_sf, out_bvecs])
+
+    # Either both --in_bval and --in_b0 are provided, or none
+    if xor(bool(args.in_bval), bool(args.in_b0)):
```
arnaudbore, line 68: is there something from argparse that can be use as a group instead of these lines

scripts/scil_compute_sf_from_sh.py
```diff
@@ -0,0 +1,122 @@
...
+        parser.error("If one of --in_bval or --in_b0 is provided, both "
+                     "are required.")
+
+    out_bvals = args.out_sf.replace(".nii.gz", ".bval")
+    if args.in_bval:
+        assert_outputs_exist(parser, args, [out_bvals])
```
arnaudbore, line 74: don't need this you can use optional input in assert_inputs_exist

scripts/scil_compute_sf_from_sh.py
```diff
@@ -0,0 +1,122 @@
...
+    add_force_b0_arg(p)
+
+    return p
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, [args.in_sh])
```
arnaudbore, line 62:
```suggestion
    assert_inputs_exist(parser, args.in_sh)
```

arnaudbore (commented):
Some minor comments.

scripts/scil_compute_sf_from_sh.py
ppoulin91: I could put a subparser, which would be used like this : `scil_compute_sf_from_sh.py sh.nii.gz sf.nii.gz dwi_like --bval dwi.bval --b0 b0.nii.gz`

CHrlS98 (approved):
Code looks good to me. So I'll go ahead and approve the PR.

However, if I'm not mistaken, the `bvec` file is saved as an `(N, 3)` matrix. Do we want to save it in either mrtrix or FSL format? I don't think it matters much since it is most likely not gonna be used with third parties software but @arnaudbore might have a different opinion on this?

ppoulin91:
> Code looks good to me. So I'll go ahead and approve the PR.
> 
> However, if I'm not mistaken, the `bvec` file is saved as an `(N, 3)` matrix. Do we want to save it in either mrtrix or FSL format? I don't think it matters much since it is most likely not gonna be used with third parties software but @arnaudbore might have a different opinion on this?

Good catch, it should be the same as FSL. Since it never had a problem loading, I never noticed the change, I'll fix it

arnaudbore:
After a long discussion with myself, please remove the mode option. I'm not really found of the hidden option with the subparser. Add an option like `--extract_as_dwi` and if this is true you check for bval and b0. For not mandatory options like these you don't need `in_` you can just put `--bval` and `--b0`.

ppoulin91:
@arnaudbore Done, let me know if you need anything else! (Also: I agree, it looks much better now with a flag + 2 args than with a subparser for so few arguments)

arnaudbore (approved):

PR #428, opened 2021-03-05
Add bundle measurements (span, curl, diameter, elongation)

yuzheww:
I added span, curl, diameter, elongation to scil_evaluate_bundles_individual_measurements.py.


scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -57,34 +56,49 @@ def compute_measures(filename_tuple):
...
-                         'mean_curvature'], [0, 0, 0, 0, 0, 0, 0, 0]))
+                         'span', 'curl', 'diameter', 'elongation', 'mean_curvature'],
+                        [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]))
 
     length_list = list(length(list(sft.streamlines)))
     length_avg = float(np.average(length_list))
     length_std = float(np.std(length_list))
     length_min = float(np.min(length_list))
     length_max = float(np.max(length_list))
+    length_sum = float(np.sum(length_list))
```
frheault, line 69: I would remove the sum, it is too dependent on streamlines count to be interpretatble.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -118,7 +132,15 @@ def main():
                     output_measures_dict[measure_name] = []
                 output_measures_dict[measure_name].append(
                     measure_dict[measure_name])
-
+    # add set stats if there are more than one bundle
+    if len(bundles_references_tuple_extended) > 1:
```
frheault, line 136: I like the idea of having an average across all input. I would like it for every metrics.
Average avg_len, min/max length, streamlines count, span, etc...

BUT I would add an optional boolean argument `--group_statistics` or `--average_measures`. This way if the user does not want it it can wont be compute and the output file is just like before.

frheault (commented):
This is great, I just asked Kurt for confirmation data/values so soon you will be able to confirm the results. In the meantime, I gave you comments to work on.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -4,10 +4,10 @@
 """
 Evaluate basic measurements of bundles, all at once.
 All tractograms must be in the same space (aligned to one reference)
-
 The computed measures are:
 volume, volume_endpoints, streamlines_count, avg_length, std_length,
-min_length, max_length, mean_curvature
+min_length, max_length, span, curl, diameter, elongation, mean_curvature
+Note: volume is calculated differently in scilpy with density
```
frheault, line 10: We determined this was not the case, right?

Also I would add a paragraph explaining what are the `set` stats and how they are computed. They are not just an average of the average, but they are weigthed by streamlines count. Explain why you think this make more sense and how you think we can interpret this.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -41,7 +43,9 @@ def _build_arg_parser():
                    help='Path of the input bundles.')
     p.add_argument('out_json',
                    help='Path of the output file.')
-
+    p.add_argument('--group_statistics', action='store_true',
+                   help='Show average measures \n'
```
frheault, line 47: A bit more detail is needed. Less than 2 sentences.

scilpy/utils/streamlines.py
```diff
@@ -82,6 +83,28 @@ def union(left, right):
     return result
 
 
+def uniformize_bundle_sft(sft, axis=None, swap=False):
+    """Uniformize a bundle"""
```
frheault, line 87: Add function docstring. Look for stuff that look like this in other module.
```
    """
    Associate the bundle to their reference (if they require a reference).
    Parameters
    ----------
    parser: argparse.ArgumentParser object
        Parser as created by argparse.
    args: argparse namespace
        Args as created by argparse.
    input_tractogram_list: list
        List of tractogram paths.
    Returns
    -------
    list: List of tuples, each matching one tractogram to a reference file.
    """
```

scripts/scil_evaluate_bundles_individual_measures.py
frheault: Also, maybe reference the paper I sent you from Frank Yeh (or the DSI studio website) for 'his' measures.
Look for `EPILOG` in other script to see how to cite publication in a script.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -118,7 +141,32 @@ def main():
                     output_measures_dict[measure_name] = []
                 output_measures_dict[measure_name].append(
                     measure_dict[measure_name])
-
+    # add set stats if there are more than one bundle
+    if args.group_statistics:
```
frheault, line 145: I like the idea of average but kind of weigthed by streamlines count.
But it is confusing (a bit) that min/max length are truly just min/max.

Also where are elongation, curl, curvature? Is that volontary?
You should document this part more.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -118,7 +141,32 @@ def main():
...
+                        output_measures_dict['streamlines_count']))
+        set_streamlines_count = \
+            np.sum(output_measures_dict['streamlines_count'])
+        output_measures_dict['set_stats'] = {}
+        output_measures_dict['set_stats']['set_streamlines_count'] = \
+            float(set_streamlines_count)
+        output_measures_dict['set_stats']['set_avg_length'] = \
+            set_total_length / set_streamlines_count
+        output_measures_dict['set_stats']['set_max_length'] = \
+            float(np.min(output_measures_dict['max_length']))
```
frheault, line 163: Max?

frheault (commented):
Last iteration, then I will test. Make sure it works ;)

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -73,18 +96,34 @@ def compute_measures(filename_tuple):
...
-                     nbr_streamlines, length_avg, length_std, length_min,
-                     length_max, float(np.mean(curvature_list))]))
+                    [volume, np.count_nonzero(endpoints_density) *
+                     np.product(voxel_size), nbr_streamlines,
+                     length_avg, length_std, length_min, length_max,
+                     span, curl, diameter, elon,
+                     float(np.mean(curvature_list))]))
+
+
+def compute_span(streamline_cords):
```
frheault, line 121: coords?

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -73,18 +96,34 @@ def compute_measures(filename_tuple):
...
+    xyz = np.asarray(streamline_cords)
+    if xyz.shape[0] < 2:
```
frheault, line 123: What is this if for?
If the streamlines has 1 point?
Just to be sure

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -118,7 +157,39 @@ def main():
...
+        # length and span are weighted by streamline count
+        set_total_length = np.sum(
+            np.multiply(output_measures_dict['avg_length'],
+                        output_measures_dict['streamlines_count']))
+        set_total_span = np.sum(
+            np.multiply(output_measures_dict['span'],
+                        output_measures_dict['streamlines_count']))
+        set_streamlines_count = \
+            np.sum(output_measures_dict['streamlines_count'])
+        set_avg_length = set_total_length / set_streamlines_count
```
frheault, line 172: I just realize that `set` could be replace by `group` and it would be clearer from an user perspective

frheault (approved):
I tested it, seems to work well. I think the measures makes sense.

scripts/scil_evaluate_bundles_individual_measures.py
yuzheww: Yes, if a streamline only has one point, then 0 is returned.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -29,19 +38,32 @@
...
+[1] Fang-Cheng Yeh. 2020.
+    Shape analysis of the human association pathways. NeuroImage.
+"""
 
 
 def _build_arg_parser():
     p = argparse.ArgumentParser(
         description=__doc__,
         formatter_class=argparse.RawTextHelpFormatter)
+    p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
```
arnaudbore, line 57: need to remove previous lines before merging

arnaudbore (approved):

GuillaumeTh (approved):

GuillaumeTh:
@arnaudbore you can merge if no more comments

PR #430, opened 2021-03-13
BF - Remove Invalid streamlines directly in decompose

frheault:


mdesco (approved):
Good to me

AntoineTheb:
Should it be an option ? (with default to True) I've had issues with scripts doing more than they should in the past, and it can be frustrating. Atleast with an option you can tell the user that streamlines may be removed.

frheault:
@AntoineTheb By design this script will remove/filter out the vast majority of streamlines. Invalid streamlines are not in the parcellation, so removing invalid prevent crashing and make sense.
It would be a lot simpler if at the end of tractoflow, right after scil_compute_local/pft. There was a call to scil_remove_invalid_streamlines.py directly...
@arnaudbore @GuillaumeTh (Its an extra 10 min, but we just have to remove it anyway immediatly for tractometry, rbx, connectoflow)

arnaudbore (approved):

PR #431, opened 2021-03-14
Update parameters of commit1/commit2

frheault:
Following a discussion with Ale, COMMIT2 parameters had to be fine-tuned as well as the output normalized.
@mdesco This should cover what we talked about.

Overall, this simply reduce the tolerance of COMMIT1 if COMMIT2 is about to be run.
Also, the raw commit weights per streamline had to be weighted by the streamline length.


mdesco (approved):
Looks good!

arnaudbore (approved):

PR #433, opened 2021-03-18
SSD and Correlation Coefficient

hermela11:
I added ssd and correlation coefficient calculations to scil_evaluate_connectivity_pairwise_agreement_measures.py

### DIFF
--- a/scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
+++ b/scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
@@ -61,7 +61,8 @@ def main():
 
     if args.single_compare:
         if args.single_compare in args.in_matrices:
-            all_matrices.remove(all_matrices[-1])
+            id = args.in_matrices.index(args.single_compare)
+            all_matrices.pop(id)
         pairs = list(itertools.product(all_matrices[:-1], [all_matrices[-1]]))
     else:
         pairs = list(itertools.combinations(all_matrices, r=2))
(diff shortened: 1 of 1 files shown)


scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
```diff
@@ -0,0 +1,72 @@
...
+        tmp_mat = load_matrix_in_any_format(filename)
+        all_matrices.append(tmp_mat / np.max(tmp_mat))
+
+    output_measures_dict = {'SSD': [], 'Correlation': []}
+    pairs = list(itertools.combinations(all_matrices, r=2))
+    for i in pairs:
+        ssd = np.sum((pairs[0] - pairs[1]) ** 2)
+        output_measures_dict['SSD'].append(ssd)
+        # only calculates for the first two arrays
+        corrcoef = np.corrcoef(all_matrices[0], all_matrices[1])
```
frheault, line 63: This should be pairs, not all_matrices.
Also np.corrcoef expects a 1D array (everything we have is 2D) and returns a covariance matrix not a scalar.
(I think this code does not execute)

scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
```diff
@@ -0,0 +1,72 @@
...
+
+    return p
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+
+    assert_inputs_exist(parser, args.in_matrices)
+    assert_outputs_exist(parser, args, [args.out_json])
```
arnaudbore, line 50:
```suggestion
    assert_outputs_exist(parser, args, args.out_json)
```

arnaudbore (commented):
Please add a test.

scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
```diff
@@ -0,0 +1,85 @@
...
+
+    assert_inputs_exist(parser, args.in_matrices)
+    assert_outputs_exist(parser, args, args.out_json)
+
+    all_matrices = []
+    for filename in args.in_matrices:
+        tmp_mat = load_matrix_in_any_format(filename)
+        all_matrices.append(tmp_mat / np.max(tmp_mat))
+
+    tmp_mat = load_matrix_in_any_format(filename)
```
frheault, line 57: There is a `if args.single_compare:` missing here
(and replace filename by args.single_compare)

So IF single compare is used, it is loaded last.

frheault (approved):
After that comments its ready to merge.
@arnaudbore I will do a PR right after to add the tests since its quite 'advanced'. Is that ok?
(Right after, meaning the same day)

frheault:
LGTM

arnaudbore (approved):

PR #436, opened 2021-03-29
Add bundle measures (area features)

yuzheww:
bundle surface area, irregularity
endpoints-related: surface area, radius, irregularity for each end surface

### DIFF
--- a/scilpy/tractanalysis/reproducibility_measures.py
+++ b/scilpy/tractanalysis/reproducibility_measures.py
@@ -1,15 +1,14 @@
 # -*- coding: utf-8 -*-
 
-import itertools
 
 from dipy.segment.clustering import qbx_and_merge
 from dipy.tracking.distances import bundles_distances_mdf
 from dipy.tracking.streamline import length, set_number_of_points
-from nibabel.streamlines.array_sequence import ArraySequence
 import numpy as np
 from numpy.random import RandomState
 from scipy.spatial import cKDTree
 from sklearn.metrics import cohen_kappa_score
+from sklearn.neighbors import KDTree
 
 from scilpy.utils.streamlines import (difference_robust,
                                       intersection_robust,
@@ -47,9 +46,9 @@ def binary_classification(segmentation_indices,
     tp = len(np.intersect1d(segmentation_indices, gold_standard_indices))
     fp = len(np.setdiff1d(segmentation_indices, gold_standard_indices))
     fn = len(np.setdiff1d(gold_standard_indices, segmentation_indices))
-    tn = len(np.setdiff1d(range(original_count),
-                          np.union1d(segmentation_indices,
-                                            gold_standard_indices)))
+    tn = len(np.setdiff1d(
+        range(original_count), np.union1d(
+            segmentation_indices, gold_standard_indices)))
     if mask_count > 0:
         tn = tn - original_count + mask_count
     # Extreme that is not covered, all indices are in the gold standard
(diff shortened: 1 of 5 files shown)


scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -124,6 +123,59 @@ def get_endpoints_density_map(streamlines, dimensions, point_to_select=1):
     return endpoints_map
 
 
+def get_head_tail_density_maps(streamlines, dimensions):
```
frheault, line 126: I think we had a very similar function doing that (both extremities at the same time).
Can you try to find it (get endpoints map I think) and maybe you could call this function inside of the other to save duplicated code (then you probably would have to merge head and tail into one)

scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -124,6 +123,59 @@ def get_endpoints_density_map(streamlines, dimensions, point_to_select=1):
...
+        xyz = streamline[0, :].astype(int)
+        endpoints_map_head[xyz[0], xyz[1], xyz[2]] += 1
+
+        xyz = streamline[-1, :].astype(int)
+        endpoints_map_tail[xyz[0], xyz[1], xyz[2]] += 1
+
+    return endpoints_map_head, endpoints_map_tail
+
+
+def approximate_surface_node(roi):
```
frheault, line 157: This is shown as new code, is this a copy paste from elsewhere?
Did I sent you code or what it already in scilpy?

EDIT: Oh no it was in a script, forget it

frheault (commented):
Before I test, can you test it on:
- Bundles from the test data and report if it makes senses
- Can you try it on a bundle that does not have any streamlines, does it work? If it crash, is the error message clear.
Can you try it on a bundle that is in fact many many bundles fused together and see if something goes wrong

scilpy/tractanalysis/reproducibility_measures.py
yuzheww: Yes, it was originally in scripts/scil_normalize_connectivity.py. I have extracted this method from that script and added the method to scilpy/tractanalysis/reproducibility_measures.py.

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -94,25 +102,58 @@ def compute_measures(filename_tuple):
...
+        np.where(endpoints_map_head != 0, 1, endpoints_map_head)
+    endpoints_map_tail_roi = \
+        np.where(endpoints_map_tail != 0, 1, endpoints_map_tail)
+    end_sur_area1 = \
+        approximate_surface_node(endpoints_map_head_roi) * (voxel_size[0] ** 2)
+    end_sur_area2 = \
+        approximate_surface_node(endpoints_map_tail_roi) * (voxel_size[0] ** 2)
+
+    endpoints_coords_head, endpoints_coords_tail = \
+        compute_endpoints_coordinates(sft)
```
frheault, line 128: This function can be removed, since you already computed both endpoints mask earlier, you can do:
`endpoints_coords_head = np.array(np.where(endpoints_map_head_roi)).T`

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -94,25 +102,58 @@ def compute_measures(filename_tuple):
...
     for i in range(nbr_streamlines):
         curvature_list[i] = mean_curvature(sft.streamlines[i])
 
     return dict(zip(['volume', 'volume_endpoints', 'streamlines_count',
                      'avg_length', 'std_length', 'min_length', 'max_length',
-                     'span', 'curl', 'diameter', 'elongation',
-                     'mean_curvature'],
+                     'span', 'curl', 'diameter', 'elongation', 'surface_area',
+                     'end_surface_area1', 'end_surface_area2',
+                     'radius1', 'radius2',
```
frheault, line 148: To stay in line with convention elsewhere, the 1 and 2 should be head and tail

frheault (commented):
(Sent some more comments over Slack)
Everything seems to work.

If you test it with the play data I original sent, and you test it with bundle with asymetric endpoints (PYT, POPT, FPT) vs very similar endpoints (ILF, CC_Pr_Po) does it make sense?

What about very long bundles (IFOF) vs very short (UF) ? If I asked you to explain to me these 4 cases and describe a bit the scalar measure, are they intuitive?

frheault (approved):

arnaudbore (approved):

PR #439, opened 2021-04-01
New tractogram scoring script - Update

AntoineTheb:
Several fixes to stale PR #359 

\- Rebase onto master
\- Fix problem with overlapping ROIs causing problems with, amongst others, scoring the FiberCup ground truth onto itself
\- Fix problem with more streamlines being segmented than total streamlines
\- Speed up by iteratively pruning streamlines
\- Handle no-connections as the leftovers fromm true and false connections
\- Handle mask dilation
\- Code clean up
\- Handle config file
\- Tests (kinda)

PR test data: https://drive.google.com/file/d/1Rh9cs59-9nlSJ8feHzJYP2R6dh5Yxuj8/view?usp=sharing

Test script example:
```
$ python scripts/scil_score_tractogram.py pr_test_data/fibercup/csd_det.tck \
pr_test_data/fibercup/[0-6]/FiberCupGroundTruth_filtered_bundle_[0-6].nii.gz \
--gt_heads pr_test_data/fibercup/[0-6]/FiberCupGroundTruth_filtered_bundle_[0-6]_head.nii.gz \
--gt_tails pr_test_data/fibercup/[0-6]/FiberCupGroundTruth_filtered_bundle_[0-6]_tail.nii.gz \
--out_dir score_fibercup_csd -f \
--reference pr_test_data/fibercup/0/FiberCupGroundTruth_filtered_bundle_0_head.nii.gz \
-v --dilate_endpoints 1 --wrong_path_as_separate --gt_config pr_test_data/fibercup/config.json
```

@frheault @GuillaumeTh @arnaudbore @mdesco Ready for review. There's probably some stuff in there that could be taken from scilpy instead of being reimplemented, please point it out if you find it. Don't be shy to comment on the code quality, this is somewhat of a complicated script and so I don't want the bus factor to be too high. Also, maybe it would be better to handle streamline indices instead of actual streamlines, but because the initial SFT gets iteratively pruned, the total number of streamlines in memory stays the same.


AntoineTheb:
@arnaudbore I haven't found actual bundle masks in the test data so I couldn't add a proper test. We'll have to add new data.

arnaudbore:
In the test data we already use, is there one which can be updated with what you need to test this ? or should we add a new one ? If you can provide me an updated one or a new one I will upload it so you can use it for your tests.

AntoineTheb:
@arnaudbore I managed to find a suitable tractogram and masks in the test data after all, no need to add data.

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,359 @@
...
+        truth mask, joining a correct combination of ROIs.
+    Bundle overlap : ground truth voxels containing tc streamline(s).
+
+Either input gt_endpoints or gt_heads and gt_tails. Ground truth and ROIs
+must be in the same order i.e. groundTruth1.nii.gz .... groundTruthN.nii.gz \
+    --gt_tails tail1.nii.gz ... tailN.nii.gz \
+    --gt_heads head1.nii.gz ... headN.nii.gz
+
+Masks can be dilated with --dilate_endpoints for bundle recognition.
+
```
mdesco, line 27: Can you provide a sample command line example please. A bit like in your PR conversation. I know it is a bit long but because the script support "complicated" input options, your example is great and will help users in the future.

mdesco: Otherwise, super clear help and description! Good job @AntoineTheb

mdesco (approved):
Super cool stuff!

This PR is related to the databases with everything needed on braindata for people to score tractograms. It's great timing actually.

@jhlegarreta can you test this PR as well please?

scripts/scil_score_tractogram.py
AntoineTheb: Thanks ! I have updated the doc to include an example in 51feb57

scilpy/tractanalysis/scoring.py
```diff
@@ -0,0 +1,266 @@
...
+from scilpy.tractanalysis.reproducibility_measures import \
+    get_endpoints_density_map
+from scilpy.tractanalysis.streamlines_metrics import compute_tract_counts_map
+from scilpy.utils.filenames import split_name_with_nii
+from scilpy.utils.streamlines import \
+    (difference, perform_streamlines_operation)
+
+
+def extract_streamlines(mask_1, mask_2, sft):
+
```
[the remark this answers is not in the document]
AntoineTheb: Donesoes in 51feb57

AntoineTheb:
Discussed offline: @frheault found a corner case when two false connections have overlapping rois. Will fix.

scilpy/tractanalysis/scoring.py
```diff
@@ -0,0 +1,389 @@
...
+    extracted_sft, mask_2_ids = filter_grid_roi(
+        tmp_sft, mask_2, 'either_end', False)
+    remaining_ids = np.setdiff1d(range(len(sft.streamlines)),
+                                 mask_1_ids[list(mask_2_ids)])
+    remaining_sft = sft[list(remaining_ids)]
+
+    return extracted_sft, remaining_sft
+
+
+def get_binary_maps(streamlines, dimensions, sft, invalid):
```
frheault, line 53: dimensions is available in the sft with `sft.dimensions`

scilpy/tractanalysis/scoring.py
```diff
@@ -0,0 +1,389 @@
...
+    tc_sft = StatefulTractogram.from_sft(tc_streamlines, sft)
+    wpc_sft = StatefulTractogram.from_sft([], sft)
+    fc_sft = StatefulTractogram.from_sft(fc_streamlines, sft)
+    if wrong_path_as_separate and len(wpc_streamlines):
+        wpc_sft = StatefulTractogram.from_sft(wpc_streamlines, sft)
+
+    return tc_sft, wpc_sft, fc_sft, nc_streamlines, sft
+
+
+def extract_false_connections(
```
frheault, line 349: Dont know if its a coding standard, but we often put as much parameters on the line, then skip and repeat

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help="Input tractogram to score")
+    p.add_argument('gt_bundles', nargs='+',
+                   help="Bundles ground truth(.[trk|tck], .nii or .nii.gz).")
```
frheault, line 82: Space before the parenthesis?

Why not (.[trk|tck], [.nii|nii.gz])

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+    p.add_argument('--gt_config', metavar='FILE',
+                   help=".json dict to specify bundles streamlines min, \
+                    max length and max angles.")
+    p.add_argument('--out_dir', default='gt_out/',
+                   help="Output directory")
+    p.add_argument('--wrong_path_as_separate', action='store_true',
+                   help="Separates streamlines that go outside of the ground \
+                        truth mask from true connections, outputs as \
+                        *_wpc.[tck|trk].")
+    p.add_argument('--remove_duplicate', action='store_true',
```
frheault, line 103: Is this use somewhere?

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+        parser.error('Can only provide --gt_endpoints or --gt_tails/gt_heads')
+    if not args.gt_endpoints and (not args.gt_tails and not args.gt_heads):
+        parser.error(
+            'Either input --gt_endpoints or --gt_heads and --gt_tails.')
+
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+
+    _, ext = os.path.splitext(args.in_tractogram)
+    sft = load_tractogram_with_reference(
```
frheault, line 144: Would it be easier to remove invalid right after loading, rather than passing the argument to a function and just do it for fc/tc/fp/tp?

scilpy/tractanalysis/scoring.py
```diff
@@ -0,0 +1,389 @@
...
+        _, ext = split_name_with_nii(gt_bundle)
+        if ext in ['.gz', '.nii.gz']:
+            gt_img = nib.load(gt_bundle)
+            gt_mask = get_data_as_mask(gt_img)
+            affine = gt_img.affine
+            dimensions = gt_mask.shape
+        else:
+            gt_sft = load_tractogram_with_reference(
+                parser, args, gt_bundle, bbox_check=False)
+            if args.remove_invalid:
```
frheault, line 139: Move this if two lines down, this function goes to vox/corner and back to its initial state.
So if you call remove_invalid_streamlines already in vox/corner you save a transformation

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+        endpoints_overreach = np.zeros(dimensions)
+        endpoints_overreach[np.where(
+            (gt_bundle_masks[i] == 0) &
+            (current_tc_endpoints_voxels >= 1))] = 1
+        tmp_dict['tc_endpoints_overlap'] = np.count_nonzero(
+            endpoints_overlap)
+        tmp_dict['tc_endpoints_overreach'] = np.count_nonzero(
+            endpoints_overreach)
+
+        final_results["bundle_wise"]["true_connections"][str(filename)] = \
```
frheault, line 348: Minor: There is a mix of `"` and `'` for string here, a few line above you use `'` for all str

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+        final_results['wpc_streamlines'] = wpc_streamlines_count
+        final_results['wpc_streamlines_ratio'] = \
+            wpc_streamlines_count / total_count
+        final_results['wpc_bundle'] = len(
+            [x for x in wpc_streamlines_list if x])
+
+    final_results['total_streamlines'] = total_count
+    final_results["bundle_wise"] = {}
+    final_results["bundle_wise"]["true_connections"] = {}
+    final_results["bundle_wise"]["false_connections"] = {}
```
frheault, line 291: Minor: There is a mix of " and ' for string here, a few line above you use ' for all str

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+    if args.verbose:
+        logging.basicConfig(level=logging.INFO)
+
+    _, ext = os.path.splitext(args.in_tractogram)
+    sft = load_tractogram_with_reference(
+        parser, args, args.in_tractogram, bbox_check=False)
+    initial_count = len(sft)
+
+    logging.info('Computing ground-truth masks')
+    gt_bundle_masks, gt_bundle_inv_masks, affine, dimensions,  = \
```
frheault, line 149: Considering the risks for it (especially for fibercup or white matter challenge data):
I would import `from dipy.io.utils import is_header_compatible` and you compare the sft to each args.gt_bundles (The function takes care of lazy_loading the data and compare header)

With a for loop, if it never return false, it means they can work together.

scripts/scil_score_tractogram.py
```diff
@@ -0,0 +1,376 @@
...
+        compute_gt_masks(args.gt_bundles, parser, args)
+
+    # If endpoints without heads/tails are loaded, split them and continue
+    # normally after. Q/C of the output is important
+    if args.gt_endpoints:
+        logging.info('Extracting ground-truth end and tail masks')
+        gt_tails, gt_heads, affine, dimensions = \
```
frheault, line 156: Same comment here:
Considering the risks for it (especially for fibercup or white matter challenge data):
I would import `from dipy.io.utils import is_header_compatible` and you compare the sft to each args.gt_*.

This could help to stop people from comparing 1) the wrong thing and 2) the right thing but in the wrong space.
Ground truth can be in high res, vs data is low res (e.g white matter challenge) or ground truth is validation data from another subject to a training subject, etc.

frheault (approved):
Nothing related to computing, mostly minor.
The only real suggestion is the thing about header compatibility.

scripts/scil_score_tractogram.py
AntoineTheb: I know most people use `"`, but it's an extra keystroke (shift) compared to `'` and I'm all about saving keystrokes lol

scripts/scil_score_tractogram.py
AntoineTheb: Good point, fixed in 0434e4e

scilpy/tractanalysis/scoring.py
AntoineTheb: Removed as per previous comment

scripts/scil_score_tractogram.py
AntoineTheb: Good point, fixed

scripts/scil_score_tractogram.py
AntoineTheb: Donesies

scil-admin:

PR #442, opened 2021-04-11
Add bundle measure (fractal dimension)

yuzheww:


scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -179,6 +179,57 @@ def approximate_surface_node(roi):
...
+                              box_size_max=None, base=2.0,
+                              threshold=0.0, box_size=None):
+    """
+    Compute the fractal dimension of a bundle to measure the roughness.
+    The code is extracted from https://github.com/FBK-NILab/fractal_dimension
+    Parameters
+    ----------
+    density: ndarray
+        A ndarray where voxel values represent the density of a bundle.
+    n_steps: int
```
frheault, line 192: No description

scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -179,6 +179,57 @@ def approximate_surface_node(roi):
...
+    box_size_min: float
+        The minimum size of a box.
```
frheault, line 194: This description does not help to understand the parameters.
What is its influence, what is a good range, how to choose it, when/why would it be better to fix it manually?
This apply to all parameters below

frheault (approved):
LGTM!

Last thing to add, could you add a similar warning to the docstring of the script itself too (scil_evaluate_bundles_individual_measures.py). Since the end user will likely read only this part.

arnaudbore:
@yuzheww please add this new option in the test of the script.

yuzheww:
> @yuzheww please add this new option in the test of the script.

Do you mean I should add a new argument for this measure in this script?

PR #444, opened 2021-04-16
Port snr in roi bitbucket

arnaudbore:
Here is some data to test: [pr444](https://box.criugm.qc.ca/f/70fb6ffe20c2427bb6dc/?dl=1)

Example: 
```
scil_snr_in_roi.py dwi.nii.gz dwi.bval dwi.bvec cc.nii.gz --split_shells --noise_mask cc.nii.gz -v
scil_snr_in_roi.py dwi.nii.gz dwi.bval dwi.bvec cc.nii.gz --noise_map noise_map.nii.gz -v
```

Completely different from the bitbucket repo: dict structure saved as json file + a lot of figures and text output only if verbose


arnaudbore:
Still needs a proper test

karanphil:
Would it be nice if the script could take a noisemap as input (optional)? @arnaudbore

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Script to compute signal to noise ratio (SNR) in a region of interest (ROI)
+of a DWI volume.
+
+It will compute the SNR for all DWI volumes of the input image seperately.
+The output will contain the SNR
```
mdesco, line 9: SNR.

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,76 @@
...
+The mean of the signal is computed inside the mask.
+The standard deviation of the noise is estimated inside noise_mask.
```
mdesco, line 11: inside the noise_mask.

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,76 @@
...
+If it's not supplied, it will be estimated using the data outside medotsu.
```
mdesco, line 12: using the data outside the brain, computed with Dipy medotsu.

mdesco (commented):
Agree with @karanphil 's suggestion. @karanphil do you have a properly scaled noise_map somewhere in your stuff or stuff from Alexis?

karanphil:
> Agree with @karanphil 's suggestion. @karanphil do you have a properly scaled noise_map somewhere in your stuff or stuff from Alexis?

Here is a link to data with a noise_map. @arnaudbore @mdesco you only have access with your udes address.

https://usherbrooke-my.sharepoint.com/:f:/g/personal/karp2601_usherbrooke_ca/Ep0kXP5OyMdBtQ_zWCQBwwABBgvvUmg-XXLRonVxFuX8pg?e=ujJneb

There is a noise map at the end of __WIP_Linear_run1_20200123125839_601_ph_ADC.nii and __WIP_Linear_run1_20200123125839_601_ADC.nii. They don't have the same scaling, I don't which one makes sense. The data to use is __WIP_Linear_run1_20200123125839_601.nii.

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,160 @@
...
+# -*- coding: utf-8 -*-
+
+"""
+Script to compute signal to noise ratio (SNR) in a region of interest (ROI)
+of a DWI volume.
+
+It will compute the SNR for all DWI volumes of the input image seperately.
+The output will contain the SNR.
+The mean of the signal is computed inside the mask.
+The standard deviation of the noise is estimated inside the noise_mask.
```
karanphil, line 11: or inside the same mask if a noise_map is provided.

karanphil (approved):

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,161 @@
...
+                   help='Path of the bvecs file, in FSL format.')
+
+    p.add_argument('in_mask',
+                   help='Binary mask of the region used to estimate SNR.')
+
+    g1 = p.add_argument_group(title='Masks options')
+    mask = g1.add_mutually_exclusive_group()
+    mask.add_argument('--noise_mask',
+                      help='Binary mask used to estimate the noise'
+                           'from the DWI.')
```
mdesco, line 57: space missing before 'from'

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Script to compute signal to noise ratio (SNR) in a region of interest (ROI)
+of a DWI volume.
+
+It will compute the SNR for all DWI volumes of the input image seperately.
+The output will contain the SNR.
```
mdesco, line 9: the SNR, which is the ratio of mean(signal) / std(noise).

scripts/scil_snr_in_roi.py
```diff
@@ -0,0 +1,161 @@
...
+The mean of the signal is computed inside the mask.
+The standard deviation of the noise is estimated inside the noise_mask
+or inside the same mask if a noise_map is provided.
+If it's not supplied, it will be estimated using the data outside the brain,
+computed with Dipy medotsu
+
+If verbose is True, the SNR for every DWI volume will be outputed.
+
+This works best in a well-defined ROI such as the corpus callosum.
+It is heavily dependent on the ROI and its quality.
```
mdesco, line 19: (nouveau paragraphe)
We highly recommend using a noise_map if you can acquire one. See refs [1, 2] that describe the noise map acquisition. 
[1] St-Jean, S., Coupé, P., & Descoteaux, M. (2016). Non Local Spatial and Angular Matching : Enabling higher spatial resolution diffusion MRI datasets through adaptive denoising. Medical Image Analysis, 32, 115–130. https://doi.org/10.1016/j.media.2016.02.010
[2]  Reymbaut, A., Caron, A. V., Gilbert, G., Szczepankiewicz, F., Nilsson, M., Warfield, S. K., Descoteaux, M., & Scherrer, B. (2021). Magic DIAMOND: Multi-fascicle diffusion compartment imaging with tensor distribution modeling and tensor-valued diffusion encoding. Medical Image Analysis, 70. https://doi.org/10.1016/j.media.2021.101988

mdesco (approved):
> I just tested the script on my data and it works well. Love the output.
> 
> The results seem good when using the `noise_map` option, as the highest SNR comes from the b0 and it decreases with respect to the b-values. However, it does not follow this pattern when using the `noise_mask` option. I found higher SNRs for b-values like b-700 or b-1200 then for the b0. And the b-2400 is super close to the b0 in terms of SNR. I think that the method of using the `noise_mask` instead of the `noise_map` is not very robust. I would very much suggest to use the `noise_map` method. We could still let the option of using `noise_mask` if someone doesn't have a `noise_map`, but I would add this in the doc. What do you think @mdesco ?

I would leave as is but like I suggest highly recommend the use of noise_map. The script was truly developed to take ratios from images. Moreover, noisemaps are not that common. But, your experience is noted and I will recommend all users to use a noisemap.

@arnaudbore  thoughts?

Ready to go after my minor edits.

PR #448, opened 2021-04-19
Refactor scil_compute_sf_from_sh.py

ppoulin91:
Using Dipy's `sh_to_sf` caused memory problems because everything is computed in float64. I refactored the code to work like `convert_sh_basis`, so it can run in parallel, and I use float32 to make it fit in memory, especially when working with 1x1x1 volumes with > 200 SF directions.

### DIFF
--- a/scilpy/reconst/multi_processes.py
+++ b/scilpy/reconst/multi_processes.py
@@ -12,6 +12,7 @@ import numpy as np
 
 cvx, have_cvxpy, _ = optional_package("cvxpy")
 
+
 def fit_from_model_parallel(args):
     model = args[0]
     data = args[1]
@@ -59,7 +60,7 @@ def fit_from_model(model, data, mask=None, nbr_processes=None):
         mask *= mask_any
 
     nbr_processes = multiprocessing.cpu_count() if nbr_processes is None \
-        or nbr_processes <= 0 else nbr_processes
+                                                   or nbr_processes <= 0 else nbr_processes
 
     # Ravel the first 3 dimensions while keeping the 4th intact, like a list of
     # 1D time series voxels. Then separate it in chunks of len(nbr_processes).
@@ -79,7 +80,7 @@ def fit_from_model(model, data, mask=None, nbr_processes=None):
     fit_array = np.zeros(data_shape[0:3], dtype='object')
     tmp_fit_array = np.zeros((np.count_nonzero(mask)), dtype='object')
     for i, fit in results:
-        tmp_fit_array[chunk_len[i]:chunk_len[i+1]] = fit
+        tmp_fit_array[chunk_len[i]:chunk_len[i + 1]] = fit
 
     fit_array[mask] = tmp_fit_array
     fit_array = MultiVoxelFit(model, fit_array, mask)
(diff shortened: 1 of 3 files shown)


scilpy/reconst/multi_processes.py
```diff
@@ -472,3 +472,83 @@ def convert_sh_basis(shm_coeff, sphere, mask=None,
...
+    sf = np.zeros((sh.shape[0], new_output_dim), dtype=np.float32)
+
+    for idx in range(sh.shape[0]):
+        if sh[idx].any():
+            sf[idx] = np.dot(sh[idx], B_in)
+
+    return chunk_id, sf
+
+
+def convert_sh_to_sf(shm_coeff, sphere, mask=None,
```
CHrlS98, line 491: Since DIPY 's `sh_to_sf_matrix` supports it, maybe we could have an argument for `full_basis`?

CHrlS98 (commented):
Alright!
A small comment, but otherwise I like it.

Also I'm wondering if the type could also be an argument to `convert_sh_to_sf` if ever we'd want to use `float64`?

scripts/scil_compute_sf_from_sh.py
```diff
@@ -43,14 +44,15 @@ def _build_arg_parser():
     p.add_argument("--extract_as_dwi", action="store_true",
                    help="Generate a DWI-like output, including a `.bval` file "
                         "and b0 images in the sf file.")
-    p.add_argument('--bval', required=True,
+    p.add_argument('--bval',
```
frheault, line 47: I really don't undestand why bvecs is automatically saved (with a default filename we can't change), but bval is optional and we can choose the name.

scripts/scil_compute_sf_from_sh.py
```diff
@@ -43,14 +44,15 @@ def _build_arg_parser():
     p.add_argument("--extract_as_dwi", action="store_true",
```
frheault, line 44: Also, I dont understand why extract_as_dwi is a parameters (it confused me at first).
Why not remove this parameters and simply say in --in_bval that the resulting SF are actually interpretable as DWI.
(It would be important to say that this is a 'related' script to scil_compute_sh_from_signal.py)

Also after that you can keep a warning for --b0 if --in_bval isn't there
Then --out_bvecs and --out_bvals could be both optional if the person want it or not.

frheault (commented):
I have a question, why only accept a sphere ?
It would be nice to have a mutually exclusive option to give either a sphere name or a bvecs so you can resample from 20 directions to 64 and not just 128 or 362 or 724.

(This is related to my suggestion that --out_bvecs and --out_bvals should all be paramaters)

ppoulin91:
@CHrlS98 : I added a dtype option (limited to float32/float64), and a --full_basis option to support asym. SH. Let me know if this fits your needs and if anything needs changing!

@frheault : You're right, it was getting confusing. I removed the --extract_as_dwi option, and added --in_* and --out_* options which are all handled separately, and it's the user's responsibility to ask for the right outputs as needed. I also added the --in_bvec option (mutually exclusive with --sphere) to use arbitrary SF directions. Let me know what you think of the new version.

CHrlS98 (approved):
Alright, 
I just tried it on full SH basis and it works nicely.
Thanks!

ppoulin91:
@frheault Some of your comments apply to functions out of the scope of the PR (I think the code was auto-formatted and I didn't notice) which are not in the script directly, so no access to `parser` and `args`. I'll fix it for my function+script, the rest should be done in another PR if we want to refactor stuff (and the `validate_nbr_processes()` function needs a touch up too, it has an unused argument and the docstring is not standard)

scil-admin:

PR #449, opened 2021-04-20
added normalize

hermela11:


scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
```diff
@@ -35,6 +35,8 @@ def _build_arg_parser():
                    help='Path of the output json file.')
     p.add_argument('--single_compare',
                    help='Compare inputs to this single file.')
+    p.add_argument('--normalize', action='store_true',
+                   help='If set, will normalize all matricies zero to one.')
```
frheault, line 39: Not zeros, the maximum values will be one
*matrices

frheault (approved):
After that small fix, ready to go

scripts/scil_evaluate_connectivity_pairwise_agreement_measures.py
hermela11: I meant the range of 0-1. Should I reword to add that in?

frheault: Ahhhh! I understand.
Maybe '**from** zero to one.' is enough then.

But right now you are not changing the smaller value. Can you add something for that
something like `tmp_mat -= np.min(tmp_mat)`

PR #450, opened 2021-04-22
Lower bound for cutting voxel label map

frheault:


scripts/scil_compute_bundle_voxel_label_map.py
```diff
@@ -127,19 +127,21 @@ def main():
                                                  np.bool)
 
     structure = ndi.generate_binary_structure(3, 1)
-    if np.count_nonzero(binary_bundle) > 10000:
+    if np.count_nonzero(binary_bundle) > 10000 and len(sft_bundle) > 100:
```
arnaudbore, line 130: Please use an argument instead of the arbitrary 100 streamlines. default 100.

arnaudbore (commented):
Small comment otherwise LGTM

arnaudbore (approved):

PR #458, opened 2021-05-28
fix_dsi_studio_trk

frheault:
After some time I gathered enough DSI-Studio corner case data to find exactly how to fix them.

Its working perfectly on 1 datasets from Leon C., 1 from Kurt S., HCP, old surgery case from Sherbrooke and the TractEM (BLSA) data
There is at least 2 thing wrong with the TRK header, 2 thing wrong with the TRK data, 1 thing wrong with the NIFTI and the NIFTI is moved to AC-PC.
That 4 things to fix to get it aligned with itself in MI-Brain and 2 more step to get to "native space'' (including rigid registration).

I can't garantee its perfect, but I think its close to it. If the data does not have too much skull and voxels smaller than 2mm iso, its pretty robust.

In summary:

1. Read the tractogram with nibabel API
2. Shift the origin by 1 voxel in X and Y
3. Flip tractogram in X/Y around its center
4. Multiply this by the affine of the DSI nifti, now this is a valid RASMM trk that you can save and it will be lined up with the DSI nifti.

If you want it on top of the original

1. If the native was already LPS, fine. If not, re-flip the TRK & the NIFTI itself around its center for each axis not being L/P/S
2. If the data has a lot of skull, crop a medium cube of from the center and register it (rigid) to the native image
3. Apply transfo to tractogram, cut invalid streamlines and save

### DIFF
--- a/scripts/scil_fix_dsi_studio_trk.py
+++ b/scripts/scil_fix_dsi_studio_trk.py
@@ -23,6 +23,9 @@ removing entire streamlines.
 This script was tested on various datasets and worked on all of them. However,
 always verify the results and if a specific case does not work. Open an issue
 on the Scilpy GitHub repository.
+
+WARNING: This script is still experimental, DSI-Studio evolves quickly and
+results may vary depending on the data itself as well as DSI-studio version.
 """
 
 import argparse
--- a/scripts/scil_fix_dsi_studio_trk.py
+++ b/scripts/scil_fix_dsi_studio_trk.py
@@ -2,34 +2,48 @@
 # -*- coding: utf-8 -*-
 
 """
+This script is made to fix DSI-Studio TRK file (unknown space/convention) to
+make it compatible with TrackVis, MI-Brain, Dipy Horizon (Stateful Tractogram).
 
+The script either make it match with an anatomy from DSI-Studio (AC-PC aligned,
+sometimes flipped) or if --in_native_fa is provided it moves it back to native
+DWI space (this involved registration).
+
+Since DSI-Studio sometimes leaves some skull around the brain, the --auto_crop
+aims to stabilize registration. If this option fails, manually BET both FA.
+Registration is more robust at resolution above 2mm (iso), be careful.
+
+If you are fixing bundles, use this script once with --save_transfo and verify
+results. Once satisfied, call the scripts on bundles using a bash for loop with
+--load_transfo to save computation.
+
+We recommand the --cut_invalid to remove invalid points of streamlines rather
+removing entire streamlines.
+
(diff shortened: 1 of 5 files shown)


scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scripts/scil_fix_dsi_studio_trk.py
```diff
@@ -0,0 +1,205 @@
...
+                              'bounding box.')
+    invalid.add_argument('--keep_invalid', action='store_true',
+                         help='Keep the streamlines landing out of the '
+                              'bounding box.')
+    add_overwrite_arg(p)
+
+    return p
+
+
+def get_axis_shift_vector(flip_axes):
```
arnaudbore, line 90: why do you have this here instead of transformation.py

scripts/tests/test_fix_dsi_studio_trk.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import os
+import tempfile
+
+from scilpy.io.fetcher import fetch_data, get_home, get_testing_files_dict
+
+# If they already exist, this only takes 5 seconds (check md5sum)
+fetch_data(get_testing_files_dict(), keys=['tracking.zip'])
```
arnaudbore, line 10: don't need to fetch anything or maybe you could add some dsi data in this package

scripts/scil_fix_dsi_studio_trk.py
arnaudbore: get_axis_flip_vector

arnaudbore (approved):
LGTM but I'm not really the one who could test this.

scripts/tests/test_fix_dsi_studio_trk.py
frheault: I deleted the line about the fetch, this script is (for now) for case-by-case basis and DSI data is basically unique so I think for now I would prefer just to keep it like that until someone actually test it in real life condition.

scripts/scil_fix_dsi_studio_trk.py
frheault: It is a piece of code unique to DSI data correction and there is no use for it outside (for now). Since this is risky spatial transform I would leave it there.

Also it use the flip_axes variable, but it is used as a shift vector (in a -= statement below)

scil-admin:

arnaudbore (approved):

PR #465, opened 2021-06-11
Simona's comments about COMMIT

frheault:
Adressing comments from Simona:

Regarding the information in the help of scil_run_commit.py I would change
**1)**
```
- compartment_EC.nii.gz (Extra-Cellular)
- compartment_IC.nii.gz (Intra-Cellular)
- compartment_ISO.nii.gz (isotropic volume fraction (freewater comportment))
```
in
```
- compartment_EC.nii.gz (estimated Extra-Cellular signal fraction)
- compartment_IC.nii.gz (estimated Intra-Cellular signal fraction)
- compartment_ISO.nii.gz (estimated isotropic signal fraction (freewater comportment))
```
**2) The lines**
```
- commit_weights.txt
    Text file containing the commit weights for each streamline of the
    input tractogram.
```
appear 2 times. Also are you sure you want to modify them from the name in the orignal code (which is streamline_weights.txt)? This would mean that running the code in your wrapper will return different outputs which in my opinion might create confusion for non-expert going back and forth from original code and your implementation.
**3) Also I think you should be careful** because with the recent introduction of T2/axon/blur/etc the streamline_weights.txt returned by COMMIT might be either with multiple components per streamline either a composed value accounting already the blur.
**4) I would change this**
`  --nbr_iter NBR_ITER   Maximum number of iterations [500].`
in
`  --nbr_iter NBR_ITER   Maximum number of iterations [1000].`
**5)  In my opinion from this**
```
  --in_peaks IN_PEAKS   Peaks file representing principal direction(s) locally,
                         typically coming from fODFs. This file is mandatory for the default
                         stick-zeppelin-ball model, when used with multi-shell data.
```
one could understand that it is always mandatory when having multi-shell data but it is not true. You can use simply ball&stick for multishell as well.
**6) Why this choice?**
```
  --commit2             Run commit2, requires .h5 as input and will force
                        ball&stick model.
```
I think I understand why you want to force a default model but then the users should be able to change that is needed. To me in this way you are limiting a lot the “main potential” of COMMIT/COMMIT2 which is being able to change models
**7) As before**
```
  --lambda_commit_2 LAMBDA_COMMIT_2
                        Specify the clustering prior strength [0.001].
```
it should be allowed to set different lambda per compartment. Also this is not a clustering prior strength but a “group prior strength” or something like that. I do not like clustering there because it might be confusing.
**8 ) Again here I’d change**
```
  --ball_stick          Use the ball&Stick model.
                        Disable the zeppelin compartment for single-shell data.
```
in something like
```
  --ball_stick          Use the ball&Stick model.
                        Disable the zeppelin compartment. Only model suitable for single-shell data.
```
or you see what I mean. To avoid that @Ale probably we should simply add an internal check and even if the user tries to use SZB for one shell data we force it to BS only
**9) I would change**
```
  --perp_diff PERP_DIFF [PERP_DIFF ...]
                        Perpendicular diffusivity in mm^2/s.
                        Default for ball_stick: None
                        Default for stick_zeppelin_ball: [0.85E-3, 0.51E-3]
```
in
```
  --perp_diff PERP_DIFF [PERP_DIFF ...]
                        Perpendicular diffusivity in mm^2/s.
                        Default for ball_stick: None
                        Default for stick_zeppelin_ball: [0.51E-3]
```
**10) for this**
```
  --iso_diff ISO_DIFF [ISO_DIFF ...]
                        Istropic diffusivity in mm^2/s.
                        Default for ball_stick: [2.0E-3]
                        Default for stick_zeppelin_ball: [1.7E-3, 3.0E-3]
```
you can actually keep [1.7E-3, 3.0E-3] also for ball_stick because you can’t distinguish properly but you’d get a linear combination as results.
**11) this looks like a very dangerous option**
```
  --threshold_weights THRESHOLD
                        Split the tractogram in two; essential and
                        nonessential, based on the provided threshold [0.0].
                         Use None to skip this step.
```

### DIFF
--- a/scripts/scil_run_commit.py
+++ b/scripts/scil_run_commit.py
@@ -10,8 +10,10 @@ The default forward microstructure model is stick-zeppelin-ball, which requires
 multi-shell data and a peak file (principal fiber directions in each voxel,
 typically from a field of fODFs).
 
-It is possible to use the ball-and-stick model for single-shell data. In this
-case, the peak file is not mandatory.
+It is possible to use the ball-and-stick model for single-shell and multi-shell
+data. In this case, the peak file is not mandatory. Multi-shell should follow a
+"NODDI protocol" (low and high b-values), multiple shells with similar b-values
+should not be used with COMMIT.
 
 The output from COMMIT is:
 - fit_NRMSE.nii.gz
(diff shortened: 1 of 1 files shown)


scil-admin:

frheault:
@mdesco This could be looked over

scil-admin:

mdesco (approved):
Good for me. Can me merged.

arnaudbore (approved):

PR #484, opened 2021-07-28
Swap x/y

frheault:
scil_visualize_connectivity.py was wrong for non-squre matrices with --name_axis option (transpose)


scil-admin:

GuillaumeTh (approved):

PR #490, opened 2021-08-12
Tracking python

FredericMassi:
Adding functionalities to local_tracking.py, utils.py and tracker.py to allow the branching streamlines algorithm. scil_compute_tracking_branching.py calls them accordingly to generate streamlines with branching. local_tracking_branching.py is a version of local_tracking that only supports the branching algorithm, not usual tracking -> it currently isn't used anywhere.


scil-admin:

scil-admin:

scil-admin:

scil-admin:

mdesco (approved):
@FredericMassi hope you have the time to perform these small edits. Let me know if you do not. @arnaudbore @CHrlS98 and I can finish it...

scilpy/tracking/local_tracking_branching.py
```diff
@@ -0,0 +1,396 @@
+# -*- coding: utf-8 -*-
```
CHrlS98, line 1: Looks like this file is an edited copy of `scilpy/tracking/local_tracking` where branching is assumed by default. Either we undo changes to `scilpy/tracking/local_tracking` and use only this file for branchings, or we keep changes to `local_tracking.py` and delete this current file. I don't know what is better, but in any case we do not need both.

scilpy/tracking/tracker.py
```diff
@@ -302,11 +302,42 @@ def get_direction(self, pos, v_in):
...
+        """
+        maxima_direction = self.tracking_field.get_tracking_maxima(pos, v_in)
+        v_out = []
+        for d in maxima_direction:
+            v_out.append(d)
+        return v_out
+
+    line_map = []
+    tree_map = []
+    _tree = -1
```
CHrlS98, line 318: These class members should be declared in a constructor (`__init__` function).

scilpy/tracking/utils.py
```diff
@@ -40,3 +40,9 @@ def __init__(self):
         self.is_all = None
         self.is_keep_single_pts = None
         self.mmap_mode = None
+        self.branching_mask = None
+        self.save_type = None
+
```
CHrlS98, line 45: Remove line break.

CHrlS98 (commented):
Ok, so I read the code and made some suggestions. Nothing critical, but I think they could increase code readability. If you think some suggestions are bad/useless, let me know @mdesco  @arnaudbore.

It will probably end on my stack anyway.

scil-admin:

CHrlS98:
@mdesco @arnaudbore @FredericMassi

I'm currently working on a tracking algorithm using the tracking modules we are using here, and I get the feeling it will get really messy if all our *proof-of-concept* tracking algorithms require new cases in `local_tracking.py`. I think we might want to keep `local_tracking.py` really general and have callbacks for method-specific behaviours. Those callbacks could be passed as arguments to the `local_tracking.track` method directly from the algorithm's script.

However this is not a simple task, especially when in the middle of a school term, so maybe we should just address the remaining comments (https://github.com/scilus/scilpy/pull/490#discussion_r705359547, https://github.com/scilus/scilpy/pull/490#discussion_r705350981 and https://github.com/scilus/scilpy/pull/490#discussion_r705377467) and refactor in a future PR.

What do you think?

mdesco:
@CHrlS98 I suggest a call with @arnaudbore and I to discuss what you suggest above. I'm a bit disconnected with that code... Maybe a good idea to include the Machine Learners working on tracking for their insights...?
@ppoulin91 @AntoineTheb @EmmaRenauld

PR #491, opened 2021-08-13
Remove rd from empty return

frheault:
Crash on empty node since the rd was removed


scil-admin:

arnaudbore (approved):

PR #494, opened 2021-08-30
Add a soft distance to facilitate manipulation

frheault:
Enhancement to avoid having to dilate mask and atlas manually before filtering.

To facilitate usage it dilate everything the same way.
The enlargement is made in term of voxel, so the provided value is relative to voxel size

### DIFF
--- a/scripts/scil_filter_tractogram.py
+++ b/scripts/scil_filter_tractogram.py
@@ -9,16 +9,16 @@ For example, --atlas_roi ROI_NAME ID MODE CRITERIA
 - MODE must be one of these values: ['any', 'all', 'either_end', 'both_ends']
 - CRITERIA must be one of these values: ['include', 'exclude']
 
-If any meant any part of the streamline must be in the mask, all means that 
+If any meant any part of the streamline must be in the mask, all means that
 all part of the streamline must be in the mask.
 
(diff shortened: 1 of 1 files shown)


scil-admin:

scripts/scil_filter_tractogram.py
```diff
@@ -96,6 +103,13 @@ def _build_arg_parser():
 def prepare_filtering_list(parser, args):
     roi_opt_list = []
     only_filtering_list = True
+
+    if args.soft_distance is not None:
+        if args.soft_distance < 1:
+            parser.error('The minimum soft distance is 1mm.')
```
arnaudbore, line 109: In the help you said it was in voxel not mm

scripts/scil_filter_tractogram.py
```diff
@@ -96,6 +103,13 @@ def _build_arg_parser():
...
+        elif args.soft_distance > 5:
+            logging.warning('Soft distance above 5mm leads to weird results.')
```
arnaudbore, line 111: Same thing voxel or mm

arnaudbore (approved):
Small comment + it needs an update for the test_filter_tractogram.py

mdesco (approved):
All good after minor edits.

PR #497, opened 2021-09-09
WIP : Added the first version of scil_make_priors_functionnectome

NotaCS:
First version of the Pythonized script to create the anatomical priors.
Pull request suggested by Arnaud.

### DIFF
--- a/scripts/scil_make_priors_functionnectome.py
+++ b/scripts/scil_make_priors_functionnectome.py
@@ -31,6 +31,7 @@ from scilpy.segment.streamlines import filter_grid_roi
 from scilpy.tracking.tools import filter_streamlines_by_length
 from scilpy.tractanalysis.streamlines_metrics import compute_tract_counts_map
 from scilpy.utils.streamlines import hack_invalid_streamlines
+from scilpy.tracking.tools import resample_streamlines_step_size
 
 from dipy.io.streamline import load_tractogram
 from dipy.io.stateful_tractogram import StatefulTractogram
@@ -84,18 +85,34 @@ def voxelize_tractogram(sft):
     '''
     Floor the points of the streamlines to the voxel they belong to and
     removes duplicate points.
-
-    To avoid missing voxels, the tractogram step-size should be smaller than
-    the voxels. (no auto-correction for that implemented yet)  # TODO
     '''
     sft.to_vox()
     sft.to_corner()
     voxed_strms = []
+    resamp = False
     for strm in sft.streamlines:
-        # strm_r = np.rint(strm).astype(np.int32)  # rounds, which is not correct here
         strm_r = strm.astype(np.int32)  # Floors, which should be correct
         _, ind = np.unique(strm_r, return_index=True, axis=0)
-        voxed_strms.append(strm_r[np.sort(ind)])
+        strm_v = strm_r[np.sort(ind)]
+        voxed_strms.append(strm_v)
+        steps = [np.linalg.norm(strm_v[i]-strm_v[i+1]) for i in range(len(strm_v)-1)]
+        if max(steps) > 1.8:  # Max distance between neighbouring voxels should be 1.73
+            print('Step size bigger than a voxel (tractogram probably compressed).'
+                  'Resampling the tractogram.')
+            resamp = True
+            break
+    if resamp:
+        stpsize = sft.voxel_sizes.min()/4  # Should be small enough
+        sft = resample_streamlines_step_size(sft, stpsize)
+        print('Resampling done.')
+        sft.to_vox()
+        sft.to_corner()
+        voxed_strms = []
+        for strm in sft.streamlines:
+            strm_r = strm.astype(np.int32)  # Floors, which should be correct
+            _, ind = np.unique(strm_r, return_index=True, axis=0)
+            strm_v = strm_r[np.sort(ind)]
+            voxed_strms.append(strm_v)
     return StatefulTractogram.from_sft(voxed_strms, sft)
 
 
--- a/scripts/scil_make_priors_functionnectome.py
+++ b/scripts/scil_make_priors_functionnectome.py
@@ -163,9 +163,7 @@ def prep_streamlines(track_Files, ref=None, from_endpoints=False, distEnd=None):
     print("Resampling at the voxel level")
     stream_tpled = voxelize_tractogram(sft)  # Not actually tupled anymore
     if from_endpoints:
-        stream_tpled = nib.streamlines.array_sequence.ArraySequence(
-                            get_endpoints(stream_tpled,
-                                          distEnd))
+        stream_tpled = AS(get_endpoints(stream_tpled, distEnd))
     if tmp_File:
         stream_tpled.save(tmp_File)
         print("Preparation for next subject done. Waiting for current subject process to end.")
@@ -276,6 +274,7 @@ def strm_multi(vlist, vdict=None, stream_tupled=None, out_dir=None, templ_i=None
             log.write(logtxt)
     else:  # Normal case
         for vox in vdict:
+            vox = tuple(vox)
             out_name = 'probaMap_{:02d}_{:02d}_{:02d}_tmp.nii.gz'.format(*vox)
             out_Ftmp = os.path.join(out_dir, out_name)
             vox_vol = np.zeros(templ_i.shape, dtype=np.float32)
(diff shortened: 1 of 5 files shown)


scil-admin:

NotaCS:
Ideally, this script should be able to run without actually installing the whole scilpy library (because I want it to be incorporated with the Functionnectome, which will be installed by non-developers, who might mess-up their python environment?

scripts/scil_make_priors_functionnectome.py
```diff
@@ -0,0 +1,108 @@
...
+
+    trk_list = glob.glob(os.path.join(args.in_dir_tractogram, '*.trk'))
+    for trk_F in trk_list:
+        print(os.path.basename(trk_F))
+        trk_sft = load_tractogram(trk_F, templ_i, bbox_valid_check=False)
+        _ = trk_sft.remove_invalid_streamlines()  # Ideally, only remove the sections outsite of the boundingbox
+        trk_sft = filter_streamlines_by_length(trk_sft, 25, 250)
+        trk_sft.to_vox()
+        for vox in vox_list:  # TODO Parallelize
+            vox_mask = base_mask.copy()
```
frheault, line 68: I think the next few lines could be replaced by
filter_grid_roi(sft, mask, filter_type, is_exclude): in segment/streamlines.py of scilpy

scripts/scil_make_priors_functionnectome.py
```diff
@@ -0,0 +1,108 @@
...
+            vox_mask[tuple(vox)] = True
+
+            vox_strm = trk_sft.get_streamlines_copy()
+            vox_strm = Streamlines(utils.target(vox_strm, templ_i.affine, vox_mask))
+            vox_strm = vox_strm.get_streamlines_copy()
+
+            vol_strm = np.zeros(templ_i.shape)
```
frheault, line 75: The next few lines should be replaced by compute_tract_counts_map() in tractanalysis/streamlines_metrics

scripts/scil_make_priors_functionnectome.py
NotaCS: Hi François, thanks for pointing out that function, I think it's exactly what I need! It can even filter end-points, which is perfect!

NotaCS:
Is there a function to crop the streamlines that have a portion outside of the bounding-box? For now, it's completely removing the said voxels, but I just want to remove the bad sections (typically, it's sections that are in the lower brain-stem / upper spinal cord).

scripts/scil_make_priors_functionnectome.py
NotaCS: C'est étrange, `target` et `filter_grid_roi` ne me donnent pas le même nombre de streamlines avec un même tractogram et mask (filter_grid_roi en donne plus). Est-ce que filter_grid_roi est plus liberal dans sa définition de ce qui est dans le mask ?

frheault: Je crois que la difference est dû à un shift, habituellement il faut un sft.to_corner() (pour définir ce qu'est le centre d'un voxel). Tu ne le faisais pas, mais la fonction le fait. Peut-être que c'est ça.

scripts/scil_make_priors_functionnectome.py
frheault: This function computes a density map (that you can then binarize). But it is robust to large step size. Doing it by hand will skip a few voxel if you step sizes are above 0.5mm, while the cython function does a line integration.

image

frheault:
@NotaCS from scilpy.utils.streamlines import cut_invalid_streamlines

That will cut the streamlines that are out. It is somewhat slow, but as good as it gets in python.

scripts/scil_make_priors_functionnectome.py
NotaCS: OK, I just tested it, no problem. Though it dosn't give the exact same voxels as my script... so I think I'll have to check how it count the streamlines

NotaCS:
> @NotaCS from scilpy.utils.streamlines import cut_invalid_streamlines
> 
> That will cut the streamlines that are out. It is somewhat slow, but as good as it gets in python.

@frheault Thank you! Indeed, it's quite slow. But in the end, with the other functions you gave me, it seems that we don't even need it anymore. the streamlines sections outside of the bounding-box will just be ignored when creating the final map.

scripts/scil_make_priors_functionnectome.py
NotaCS: The tractogram I'm using has a step size of 0.5mm, but voxels of 2mm in the final map, so it shouldn't be a problem. It might be due to the shift you mentioned in the other comment. I'll investigate, thanks!

NotaCS:
Alright, so actually, I still needed to remove the bad streamline points... So I made my own function!
It's rougher but much much faster than cut_invalid_streamlines (6 sec instead of 6 min).

frheault:
@NotaCS The code is cryptic, what is the difference?

sft.remove_invalid_streamlines() remove the entire streamline if it contains a single invalid point.
cut_invalid_streamlines() keeps the longest possible segment that is made only of valid streamlines.

What is yours doing? From what I understand it just crops out invalid points and reconnects the streamlines after, which is not a behavior I trust fully (But I doubt that in human data it's a problem, so it's fair).

I saw the test 'strm[0] > epsilon'. This is a check for the lower corner of the grid. You also need a test for the upper corner of the grid: Which is strm[0] < grid.shape-epsilon (It is a 3D test since the shape is not always the same value in x/y/z)

Also that test seems to remove the streamlines that have their first point and last out of the bounding box. Which is the vast majority of case (in the CST or top of the CC) It is very rare to enter and exit back a volume. At this point you should simply use sft.remove_invalid_streamlines()? No?

We can talk on Teams if thats faster.

scil-admin:

NotaCS:
New, full version available, with a parallel processing option and region-wise priors creation.
For now, it is just way too slow, as filter_grid_roi takes 8 sec to generate filter the whole tractogram, and we have to use it on all the voxels (about 230,000 times), for every subjet (100), so that's not possible.
I'm working on a function to reduce the size of the tractogram to speed this up.

scil-admin:

NotaCS:
With the new version, one the map for one voxel (only for one subject) takes now less than 2sec. With parallelisation of the process (say, no 32 gpu), it should take: (time per voxel) * (nb of voxel) * (nb of subjects) / (nb of cpu) = 1.5*228,336*100/32 = 1070325 sec = 300h = about 12 days....
Unless I'm missing something, it's still way too long (but it's much better than before).
Any suggestion?

scil-admin:

NotaCS:
I just pushed the new code, much faster, but still a bit too slow for my test (it would take 3-4 days on 32cpu to run it on the 100 subjects). I'm trying to implement another approach proposed by Guillaume that may speed things up quite a bit

NotaCS:
New version, iterating over streamlines instead of voxels (in a first step at least), as per Guillaume advice. Much (much!) faster. But gives a "Broken pipe" error if I run it on too many (>=6) CPUs at the same time on my local machine.

scil-admin:

scil-admin:

NotaCS:
Mmmmhh... It doesn't work well. I don't know why, but it only spawned 3 workers (when I asked 6), one got stuck quickly, and my RAM usage went through the roof (although each processes should only have to work with about 3 or 4 GB of data, max)...

scripts/scil_make_priors_functionnectome.py
```diff
@@ -0,0 +1,377 @@
...
+            if nb_proc == 1:
+                print('Starting process')
+                strm_multi(vox_dict, stream_tupled, args.out_dir, templ_i)
+            else:  # Parallelize
+                vox_batchs = np.array_split(list(vox_dict), nb_proc)
+                vox_dict_batchs = [{tuple(v): vox_dict[tuple(v)] for v in vb} for vb in vox_batchs]
+                print('Starting parallel process')
+                with Pool(processes=nb_proc) as pool:
+                    poolCheck = pool.starmap_async(strm_multi, zip(vox_dict_batchs,
+                                                                   [stream_tupled]*nb_proc,
```
GuillaumeTh, line 216: Here instead of duplicate n times the data, prepare your batches before and give an array of batches with 1 batch per process. In the end you will just have 1 time the data in RAM instead of n times.

NotaCS: normally, [stream_tupled]*nb_proc does not make copy of the obejct (each item of the list points to the same original object). And I can't split stream_tupled into batches, because each process may need any streamline in it. Or am I missing something?

NotaCS:
Alright, so it was indeed probably the starmap that was messing things up. Now it works perfectly. I could optimise the RAM a bit more (as each worker has a copy of the data right now), but it will do for now, it's not that hungry.

scil-admin:

NotaCS:
Ran well: on a local machine, with 8 CPUs, it took a bit less than 17 hours and about 13GB of RAM. Perfect :)

NotaCS:
Mmmh... I compared a map with one from the original priors. It's quite similar, but really not exactly the same. A quick correlation between the two (on the brain voxels only) gives me R = 0.9, which is quite low considering that they are supposed to be the same. Some voxels have a difference of up to 0.66 (knowing that a map ranges from 0 to 1, it's a huge discrepancy).

scil-admin:

NotaCS:
When I put a tractogram in vox space, and align it to the corner, what exactly mean the coordinates? 
For example, does a streamline point at (15, 25, 14) means that I am looking at a point in the center of voxel (15, 25, 14), or at one of its corner?

scil-admin:

scil-admin:

scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-497/11/

scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-497/13/

NotaCS:
I'll close the PR for now, to stop the automatic messages that must be sent to everyone each time I update something ;)

PR #500, opened 2021-09-29
Support empty bundle in tractometry

frheault:
RBx-Flow does not output empty bundle, so it works well with Tractometry-Flow
However, bundles from other sources could be empty.

These minimal changes allow the pipeline to execute the same way if a bundle is just not there or if it is there, but empty.


scil-admin:

arnaudbore (approved):
Go !

PR #501, opened 2021-09-30
Add inverse operation flag

frheault:
Following an issue raised by @StongeEtienne (Sorry I forgot about it), this allows for a more intuitive usage of the transform script.
TestingData:
https://drive.google.com/file/d/1VcTaBS_eCfze4tmM7cpliTYyJ6S6Bp0i/view?usp=sharing

It is still very hard to know what it's doing for anyone not familiar with the convention.
```
Example:
To apply transform from ANTS to tractogram. If the ANTS commands was
MOVING->REFERENCE, this will bring a tractogram from MOVING->REFERENCE
scil_apply_transform_to_tractogram.py ${MOVING_FILE} ${REFERENCE_FILE}
                                        0GenericAffine.mat ${OUTPUT_NAME}
                                        --inverse
                                        --in_deformation 1InverseWarp.nii.gz

If the ANTS commands was MOVING->REFERENCE, this will bring a tractogram
from REFERENCE->MOVING
scil_apply_transform_to_tractogram.py ${MOVING_FILE} ${REFERENCE_FILE}
                                        0GenericAffine.mat ${OUTPUT_NAME}
                                        --in_deformation 1Warp.nii.gz
                                        --reverse_operation
```
Overall the idea is that a registration from ANTS has ''a direction'' SOURCE->TARGET, but ANTs nomenclature is made for image, so it has to be the inverse transformations (so --inverse + using InverseWarp).
When going against the initial direction of the ANTS command, the inverse transformations are no needed anymore BUT the order of the operation has to be inverted.

I use the flag "--_reverse_operation_", I am open to using a better name (--inverse cannot change, this is too common to change)

### DIFF
--- a/scilpy/utils/streamlines.py
+++ b/scilpy/utils/streamlines.py
@@ -410,10 +410,9 @@ def concatenate_sft(sft_list, erase_metadata=False, metadata_fake_init=False):
     return fused_sft
 
 
-def transform_warp_streamlines(sft, linear_transfo, target, inverse=False,
-                               deformation_data=None,
-                               remove_invalid=True, cut_invalid=False):
-    # TODO rename transform_warp_sft
+def transform_warp_sft(sft, linear_transfo, target, inverse=False,
+                       reverse_op=False, deformation_data=None,
+                       remove_invalid=True, cut_invalid=False):
     """ Transform tractogram using a affine Subsequently apply a warp from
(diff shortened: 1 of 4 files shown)


scil-admin:

GuillaumeTh (approved):

PR #506, opened 2021-10-08
WIP: Encapsulating parts in the tracking scripts and classes

EmmaRenauld:
I started checking which parts can be incorporated in dwi_ml by encapsulating the parts of code that can be re-used in our case.

WIP, but I am already placing the pull request so that people can now what I have been testing in our future tractography discussions.

Questions that I have so far:
 - in general: a lot of suggestions from pycodestyle and others (ex: camelcase, attributes set outside of init, imports not used, etc. ). Would a cleaning be good?
 - in scil_compute_local_tracking
        1. in _get_direction_getter(args, mask_data), mask_data is not used!
        2. in load_mask_and_verify_anisotropy: the comment says anisotropy is verified in the mask, but it is actually verified on the sh file. (see other comment 1 below)
 
 - In scilpy.tracking.local_tracking:
       3. In the function 'track': there is a tracker and a pft_tracker. I wanted to find the difference between the two. Following the code: pft_tracker is used : 
          ......in get_streamlines:
          ..............in get_line_from_seed:
          ....................in _get_line: It is not used!!!
          Why is is passed all the way down not to be used? Code's help could be clarified.

- In scilpy.tracking.tracking_field:
           4. `self.maxima_neighbours = self.get_direction_neighbours(np.pi / 16.)` Why is this not theta? Or another variable? Why is this hardcoded?
           5. SphericalHarmonicField.get_maxima(): I can't find any usage for this in all of scilpy. You seem to be using reconst.utils.get_maximas though.

Ideas on how to make integration with dwi_ml as smooth as possible: 
- in scil_compute_local_tracking:
        1. in load_mask_and_verify_anisotropy : Considering that in our case, we may not use the sh file, could this be done on the mask and suppose that the input data will have the same voxel size? This way we could use the same function. Also, currently, sh_data is loaded twice. When verifying the anisotropy, and in _get_direction_getter.

### DIFF
--- a/scilpy/tracking/local_tracking.py
+++ b/scilpy/tracking/local_tracking.py
@@ -12,17 +12,21 @@ import numpy as np
 
 from dipy.tracking.streamlinespeed import compress_streamlines
 
+from scilpy.tracking.tracker import AbstractTracker
+
 data_file_info = None
 
 
-def track(tracker, mask, seed, param, compression_th=0.1, nbr_processes=1,
-          max_tries=None, save_seeds=False):
+def track(tracker: AbstractTracker, mask, seed, param, compression_th=0.1,
+          nbr_processes=1, max_tries=None, save_seeds=False):
     """
     Generate a set of streamline from seed, mask and odf files.
 
     Parameters
     ----------
     tracker : Tracker, tracking object.
+        See scilpy.tracking.tracker.abstractTracker. It contains a
+        tracking_field (input data) and a propagator.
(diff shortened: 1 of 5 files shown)


scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-506/2/

scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-506/4/

PR #511, opened 2021-10-26
Modifications in the (bitbucket) tracking scripts - part 2

EmmaRenauld:
**Bigger changes:** 

- **Merging the tracker and the propagator** into propagator to avoid circular reference. Old file tracker.py (containing the two classes) becomes propagator.py
- Adding a main **script** scil_compute_local_tracking_2.py (other names suggested?)
- New "Tracker" class
    - Englobes the track() function that was previously alone in local_tracking.py. This will allow creating a child version in dwi_ml.     File renamed tracker.py
    - Now the main script is: 1) Initialize the tracker. 2)  Tracker.track(). 
    - So to be clear, this new "Tracker" is absolutely not the same as the old, it is one step above. See new diagram:

https://docs.google.com/presentation/d/1mvvBYjKZ-IVmkl1EYWBTQ_8T8H0R3_Y-/edit?usp=sharing&ouid=101323301275554058305&rtpof=true&sd=true 

**Smaller changes:** 

- Encapsulating parts in the scil_compute_local_tracking that are common in scilpy.tracking.utils
    - Similiar to functions like add_processes_arg in scilpy.io.utils.
    - It makes the proof-reading a bit less natural but at least the args are the same in both scripts. Do we like that? Separated in a few sub-functions so I can use only some of them in dwi_ml. If we prefer, I can copy all of this back into each script.
- Now the seedGenerator and DataVolume take receive the data instead of the img. In dwi_ml, we have the data and affine, not the image, it would be a little weird to create the image back just to re-divide it.
- In the DataVolume: renaming methods as in the SFT nomenclature. Ex: using "voxmm"  instead of "position", to clarify. Adding "origin" option (center or corner)
- AbstractTrackingField created above the ODFTrackingField.

### DIFF
--- a/scilpy/image/datasets.py
+++ b/scilpy/image/datasets.py
@@ -2,7 +2,7 @@
 import numpy as np
 
 from dipy.core.interpolation import trilinear_interpolate4d, \
-                                    nearestneighbor_interpolate
+    nearestneighbor_interpolate
 
 
 class DataVolume(object):
@@ -46,9 +46,8 @@ class DataVolume(object):
 
     def get_voxel_value(self, i, j, k):
         """
-        Get the voxel value at x, y, z in the dataset
-        if the coordinates are out of bound, the nearest voxel
-        value is taken.
+        Get the voxel value at x, y, z in the dataset.
+        If the coordinates are out of bound, the nearest voxel value is taken.
 
         Parameters
         ----------
@@ -84,7 +83,7 @@ class DataVolume(object):
         return (0 <= i < self.dim[0] and 0 <= j < self.dim[1] and
                 0 <= k < self.dim[2])
 
-    def get_voxel_at_position(self, x, y, z):
+    def voxmm_to_idx(self, x, y, z, origin='center'):
         """
         Get the 3D indice of the closest voxel at position x, y, z expressed
         in mm.
@@ -93,17 +92,29 @@ class DataVolume(object):
         ----------
         x, y, z: floats
             Position coordinate (mm) along x, y, z axis.
+        origin: str
+            'Center': Voxel 0,0,0 goes from [-resx/2, -resy/2, -resz/2] to
+                [resx/2, resy/2, resz/2].
+            'Corner': Voxel 0,0,0 goes from [0,0,0] to [resx, resy, resz].
 
         Return
         ------
         out: list
             3D indice of voxel at position x, y, z.
         """
-        return [(x + self.pixdim[0] / 2) // self.pixdim[0],
-                (y + self.pixdim[1] / 2) // self.pixdim[1],
-                (z + self.pixdim[2] / 2) // self.pixdim[2]]
+        if origin == 'center':
+            return np.asarray([(x + self.pixdim[0] / 2) // self.pixdim[0],
+                               (y + self.pixdim[1] / 2) // self.pixdim[1],
+                               (z + self.pixdim[2] / 2) // self.pixdim[2]],
+                              dtype=int)
+        elif origin == 'corner':
+            return np.asarray([x // self.pixdim[0],
+                               y // self.pixdim[1],
+                               z // self.pixdim[2]], dtype=int)
+        else:
+            raise ValueError("Origin must be one of 'center' or 'corner'.")
 
-    def get_voxel_coordinate(self, x, y, z):
+    def voxmm_to_vox(self, x, y, z):
         """
         Get voxel space coordinates at position x, y, z (mm).
 
@@ -119,24 +130,7 @@ class DataVolume(object):
         """
         return [x / self.pixdim[0], y / self.pixdim[1], z / self.pixdim[2]]
 
-    def get_voxel_value_at_position(self, x, y, z):
-        """
-        Get value of the voxel closest to position x, y, z (mm) in the dataset.
-        No interpolation is done.
-
-        Parameters
-        ----------
-        x, y, z: floats
-            Position coordinate (mm) along x, y, z axis.
-
-        Return
-        ------
-        value: ndarray (self.dim[-1],)
-            The value evaluated at position x, y, z.
-        """
-        return self.get_voxel_value(*self.get_voxel_at_position(x, y, z))
-
-    def get_position_value(self, x, y, z):
+    def voxmm_to_value(self, x, y, z):
         """
         Get the voxel value at voxel position x, y, z (mm) in the dataset.
         If the coordinates are out of bound, the nearest voxel value is taken.
@@ -154,7 +148,7 @@ class DataVolume(object):
             is of length 1, return a scalar value.
         """
         if self.interpolation is not None:
-            if not self.is_position_in_bound(x, y, z):
+            if not self.is_voxmm_in_bound(x, y, z):
                 eps = float(1e-8)  # Epsilon to exclude upper borders
                 x = max(-self.pixdim[0] / 2,
                         min(self.pixdim[0] * (self.dim[0] - 0.5 - eps), x))
@@ -162,8 +156,7 @@ class DataVolume(object):
                         min(self.pixdim[1] * (self.dim[1] - 0.5 - eps), y))
                 z = max(-self.pixdim[2] / 2,
                         min(self.pixdim[2] * (self.dim[2] - 0.5 - eps), z))
-            coord = np.array(self.get_voxel_coordinate(x, y, z),
-                             dtype=np.float64)
+            coord = np.array(self.voxmm_to_vox(x, y, z), dtype=np.float64)
 
             if self.interpolation == 'nearest':
                 result = nearestneighbor_interpolate(self.data, coord)
@@ -174,12 +167,11 @@ class DataVolume(object):
             # Squeezing returns only value instead of array of length 1 if 3D
             # data
             return np.squeeze(result)
-
         else:
             raise Exception("No interpolation method was given, cannot run "
                             "this method..")
 
-    def is_position_in_bound(self, x, y, z):
+    def is_voxmm_in_bound(self, x, y, z, origin='center'):
         """
         Test if the position x, y, z mm is in the dataset range.
 
@@ -187,10 +179,14 @@ class DataVolume(object):
         ----------
         x, y, z: floats
             Position coordinate (mm) along x, y, z axis.
+        origin: str
+            'Center': Voxel 0,0,0 goes from [-resx/2, -resy/2, -resz/2] to
+                [resx/2, resy/2, resz/2].
+            'Corner': Voxel 0,0,0 goes from [0,0,0] to [resx, resy, resz].
 
         Return
         ------
         value: bool
             True if position is in dataset range and false otherwise.
         """
-        return self.is_voxel_in_bound(*self.get_voxel_at_position(x, y, z))
+        return self.is_voxel_in_bound(*self.voxmm_to_idx(x, y, z, origin))
--- a/scilpy/image/datasets.py
+++ b/scilpy/image/datasets.py
@@ -74,18 +74,20 @@ class DataVolume(object):
 
         Parameters
         ----------
-        i, j, k: ints
-            Voxel indice along each axis.
+        i, j, k: ints or floats
+            Voxel indice along each axis (as ints) or voxel coordinates in
+            voxel world (as floats).
 
         Return
         ------
         out: bool
             True if voxel is in dataset range, False otherwise.
         """
-        return (0 <= i < self.dim[0] and 0 <= j < self.dim[1] and
-                0 <= k < self.dim[2])
+        return (0 <= i <= (self.dim[0] - 1) and
+                0 <= j <= (self.dim[1] - 1) and
+                0 <= k <= (self.dim[2] - 1))
 
-    def voxmm_to_idx(self, x, y, z, origin='center'):
+    def voxmm_to_idx(self, x, y, z, origin):
         """
         Get the 3D indice of the closest voxel at position x, y, z expressed
         in mm.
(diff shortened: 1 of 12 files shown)


scilpy/tracking/tracking_field.py
```diff
@@ -1,82 +1,156 @@
...
+            self.dirs = np.zeros(len(self.sphere.vertices), dtype=np.ndarray)
+            for i in range(len(self.sphere.vertices)):
+                self.dirs[i] = TrackingDirection(self.sphere.vertices[i], i)
+            self.tracking_neighbours = self._get_sphere_neighbours(self.theta)
+        else:
+            self.sphere = None
+            self.dirs = None
+            self.tracking_neighbours = None
+
+    def _get_sphere_neighbours(self, max_angle):
```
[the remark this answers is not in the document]
EmmaRenauld: In the end, it is easier from integration with dwi_ml to avoid a child class and to have this as an option.

I add a check in this method, and get_opposite_direction_sphere won't crash now if self.dirs = [] instead of None.

scilpy/tracking/propagator.py
```diff
@@ -0,0 +1,280 @@
...
+            A valid tracking direction. None if no valid direction is found.
+        """
+        sf, directions = self.tracking_field.get_next_direction(
+            pos, v_in, 'prob')
+        if np.sum(sf) > 0:
+            return directions[sample_distribution(sf)]
+        return None
+
+
+class DeterministicMaximaSHPropagator(AbstractPropagator):
```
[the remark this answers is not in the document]
EmmaRenauld: (Answer after talking to Maxime: DeterministicSH is enough, but we could have one day DeterministicPeak (not on the SH). Modified.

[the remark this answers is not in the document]
CHrlS98: Super!

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scripts/scil_compute_local_tracking_2.py
```diff
@@ -0,0 +1,227 @@
+#! /usr/bin/env python3
```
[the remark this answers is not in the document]
mdesco: I like scil_compute_local_tracking_dev more than "in-house". "dev" means it is wip and research... in-house could be misleading.

scripts/scil_compute_local_tracking_2.py
```diff
@@ -0,0 +1,228 @@
...
+# -*- coding: utf-8 -*-
+
+"""
+Local streamline HARDI tractography using only scilpy methods -- no dipy (i.e
```
mdesco, line 5: using scilpy-only methods

scripts/scil_compute_local_tracking_2.py
```diff
@@ -0,0 +1,228 @@
...
+
+As in scil_compute_local_tracking:
+
+    The tracking direction is chosen in the aperture cone defined by the
+    previous tracking direction and the angular constraint.
+    - Algo 'det': the maxima of the spherical function (SF) the most closely
+    aligned to the previous direction.
+    - Algo 'prob': a direction drawn from the empirical distribution function
+    defined from the SF.
+    - Algo 'eudx' is not yet available!
```
mdesco, line 19: just remove that line. We will most likely never make eudx available, no?

mdesco (commented):
Testing the script, I get the following error from fODF coming from TractoFlow on HCP subject.
scil_compute_local_tracking_2.py sub-197348__fodf.nii.gz seeds_fa015.nii.gz seeds_fa015.nii.gz test.trk --processes 8
  File "/Users/desm2239/my_env/scilpy_master/bin/scil_compute_local_tracking_2.py", line 7, in <module>
  File "/Users/desm2239/Research/Source/scilpy/scripts/scil_compute_local_tracking_2.py", line 228, in <module>
    main()
  File "/Users/desm2239/Research/Source/scilpy/scripts/scil_compute_local_tracking_2.py", line 181, in main
    dataset = DataVolume(odf_sh_img, args.sh_interp)
  File "/Users/desm2239/Research/Source/scilpy/scilpy/image/datasets.py", line 47, in __init__
    self.nbr_voxel = self.data.size
AttributeError: 'Nifti1Image' object has no attribute 'size'

arnaudbore (commented):
Really good job. Some missing docstring. More explainations and we are almost good to go. Didn't test.

scripts/scil_compute_local_tracking_2.py
EmmaRenauld: I wasn't sure. Ok.

scilpy/tracking/propagator.py
```diff
@@ -0,0 +1,266 @@
...
+        Should use self.tracking_field.get_possible_next_directions.
+
+        Parameters
+        ----------
+        pos: ndarray (3,)
+            Current tracking position.
+        v_in: ndarray (3,)
+            Previous tracking direction.
+        """
+        pass
```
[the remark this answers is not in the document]
EmmaRenauld: Changed to NotImplementedError. Better, or still unclear?

scilpy/tracking/seed.py
```diff
@@ -1,23 +1,40 @@
...
+        voxres: np.array(3,)
+            The pixel resolution, ex, using img.header.get_zooms()[:3].
         """
-        self.pixdim = img.header.get_zooms()[:3]
+        self.data = data
+        self.voxres = voxres
 
-        data = img.get_fdata(caching='unchanged', dtype=np.float64)
-        self.seeds = np.array(np.where(np.squeeze(data) > 0)).transpose()
+        # self.seed_voxels are all the voxels where a seed could be placed
```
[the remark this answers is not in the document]
EmmaRenauld: oups, thanks,.

scripts/scil_compute_local_tracking_2.py
```diff
@@ -0,0 +1,228 @@
...
+    seed_res = seed_img.header.get_zooms()[:3]
+    seed_generator = SeedGenerator(seed_data, seed_res)
+    if args.npv:
+        nbr_seeds = len(seed_generator.seeds) * args.npv
+    elif args.nt:
+        nbr_seeds = args.nt
+    else:
+        # Setting npv = 1.
+        nbr_seeds = len(seed_generator.seeds)
+    if len(seed_generator.seeds) == 0:
```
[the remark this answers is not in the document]
EmmaRenauld: Did you mean load the seeding mask before the tracking mask? If so, ok, changed.

scripts/scil_compute_local_tracking.py
```diff
@@ -119,12 +74,12 @@ def _build_arg_parser():
 
 
 def _get_direction_getter(args):
-    sh_data = nib.load(args.in_sh).get_fdata(dtype=np.float32)
+    odf_data = nib.load(args.in_odf).get_fdata(dtype=np.float32)
```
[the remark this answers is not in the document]
EmmaRenauld: Yeah, I thought some people would be unhappy. But with Philippe using SH fitted to the raw signal, I thought it was a necessary move.

scilpy/tracking/utils.py
```diff
@@ -13,30 +13,87 @@ def __init__(self, cartesian, index=None):
...
+                   help='Seeding mask (.nii.gz).')
+    p.add_argument('in_mask',
+                   help='Tracking mask (.nii.gz).\n'
+                        'Tracking will stop outside this mask.')
+    p.add_argument('out_tractogram',
+                   help='Tractogram output file (must be .trk or .tck).')
+
+
+def add_tracking_options(p):
+    track_g = p.add_argument_group('  Tracking options')
```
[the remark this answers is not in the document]
EmmaRenauld: I thought it was maybe easier to perceive that they were still under "optional arguments" and not "positional arguments", but with the '--var' we know it anyway so I don't mind.

EmmaRenauld:
> Testing the script, I get the following error from fODF coming from TractoFlow on HCP subject. scil_compute_local_tracking_2.py sub-197348__fodf.nii.gz seeds_fa015.nii.gz seeds_fa015.nii.gz test.trk --processes 8 Traceback (most recent call last): File "/Users/desm2239/my_env/scilpy_master/bin/scil_compute_local_tracking_2.py", line 7, in exec(compile(f.read(), **file**, 'exec')) File "/Users/desm2239/Research/Source/scilpy/scripts/scil_compute_local_tracking_2.py", line 228, in main() File "/Users/desm2239/Research/Source/scilpy/scripts/scil_compute_local_tracking_2.py", line 181, in main dataset = DataVolume(odf_sh_img, args.sh_interp) File "/Users/desm2239/Research/Source/scilpy/scilpy/image/datasets.py", line 47, in **init** self.nbr_voxel = self.data.size AttributeError: 'Nifti1Image' object has no attribute 'size'

Sorry. Forgot to test after the last commit. Should be good now.

scil-admin:

arnaudbore (commented):
LGTM I'm waiting for @mdesco and @CHrlS98 approvals.

EmmaRenauld:
There is a subject that should still be discussed I think. This version works, but in the long term, I don't know if it's always going to fit with any model. I think that eventually, the tracking field should also be merged in the propagator. 

Currently the tracking field: 1) gets the data (ex, fODF-sh), 2) returns the directions/maxima that are in a cone of angle theta. Then, the propagator sample from these directions.

But let's take the example of DTI. Sampling would be done direction on the tensor. I don't think we can really take a list of directions that are in a cone theta and sample on these directions, can we? So  sampling should entirely be left to the propagator, and sampling should mean either, 'sampling from model' or 'sampling from a list of direction' based on the type of data. And it the case of DTI, theta is a stopping criterion, not a restreint on the tracking field. 

Should we merge the tracking field and propagator? (Possibly in a 3rd PR)

scilpy/image/datasets.py
```diff
@@ -84,7 +85,7 @@ def is_voxel_in_bound(self, i, j, k):
         return (0 <= i < self.dim[0] and 0 <= j < self.dim[1] and
                 0 <= k < self.dim[2])
 
-    def get_voxel_at_position(self, x, y, z):
+    def voxmm_to_idx(self, x, y, z, origin='center'):
```
CHrlS98, line 88: Is the new `origin` argument useful? Why would we need this kind of flexibility? Could we not just set it to one or the other and take it into account when writing our tracking scripts?

scilpy/tracking/seed.py
```diff
@@ -1,27 +1,43 @@
...
+        # (voxel space, int numbers).
+        self.seeds = np.array(np.where(np.squeeze(data) > 0),
+                              dtype=float).transpose()
+        if len(self.seeds) == 0:
+            logging.warning("There are positive voxels in the seeding mask!")
 
     def get_next_pos(self, random_generator, indices, which_seed):
         """
-        Generate the next seed position.
+        Generate the next seed position (Space=voxmm, origin=corner).
```
CHrlS98, line 40: We need to be careful with origin. It is `corner` here, but default is `center` in `DataVolume`.

CHrlS98 (commented):
Ok, my last comments.

scil-admin:

scilpy/image/datasets.py
EmmaRenauld: That dataset is not necessarily only associated with our traking script, though. It could eventually be used elsewhere, so I tought it was better to be thorough, like in the sft. But I won't object if others agree.

scilpy/tracking/seed.py
EmmaRenauld: I will just remove the default in DataVolume. Nice catch.

CHrlS98:
> There is a subject that should still be discussed I think. This version works, but in the long term, I don't know if it's always going to fit with any model. I think that eventually, the tracking field should also be merged in the propagator.
> 
> Currently the tracking field: 1) gets the data (ex, fODF-sh), 2) returns the directions/maxima that are in a cone of angle theta. Then, the propagator sample from these directions.
> 
> But let's take the example of DTI. Sampling would be done direction on the tensor. I don't think we can really take a list of directions that are in a cone theta and sample on these directions, can we? So sampling should entirely be left to the propagator, and sampling should mean either, 'sampling from model' or 'sampling from a list of direction' based on the type of data. And it the case of DTI, theta is a stopping criterion, not a restreint on the tracking field.
>
> Should we merge the tracking field and propagator? (Possibly in a 3rd PR)

Maybe we can wait for a specific situation where this refactoring is needed and do it if and when we need it, so that the new design really suits our requirements. Also I think deterministic tracking with DTI would work, because peaks are extracted on the whole sphere before selecting the direction inside a cone (if I'm not mistaken).

EmmaRenauld:
> Maybe we can wait for a specific situation where this refactoring is needed and do it if and when we need it, so that the new design really suits our requirements. Also I think deterministic tracking with DTI would work, because peaks are extracted on the whole sphere before selecting the direction inside a cone (if I'm not mistaken).

Well I actually need now to decide if I will try to force dwi_ml to fit with all this or not. In our case, the tracking_field gets the inputs at current position and runs it through the model. The output from the model is not necessarily a list of directions. It could be, ex, a learned tensor, which is the same as my DTI example above.

scilpy/image/datasets.py
CHrlS98: Yes ok. But in this case we need to be extra careful when writing our tracking scripts, because the `SeedGenerator` always returns positions with origin `corner`. Should the `SeedGenerator` have an origin argument too then?

CHrlS98:
Will `dwi_ml` be incorporated to `scilpy` (if so the changes can be done in its associated PR), or will the `scilpy` modules be imported in another project?

I think it would be better to keep the PR as is and merge it before making new changes, cause this is already a nice refactor that could benefit others right away.

scil-admin:

scil-admin:

scil-admin:

scil-admin:

mdesco (commented):
It is still crashing when I run:
scil_compute_local_tracking_dev.py sub-197348__fodf.nii.gz seeds_fa015.nii.gz seeds_fa015.nii.gz test.trk --processes 8

Data to test:
https://www.dropbox.com/s/a4s8gvkbro4pqz3/PR_511.zip?dl=0

scil-admin:

scilpy/tracking/tracking_field.py
```diff
@@ -88,33 +179,67 @@ def _get_sf(self, pos):
...
+
+        Parameters
+        ----------
+        pos: ndarray (3,)
+            Position in trackable dataset, expressed in mm.
+        previous_direction: TrackingDirection
+            Incoming direction. Outcoming direction won't be further than an
+            angle theta.
+        tracking_choice: str
+            Either "prob" or "det"
```
CHrlS98, line 220: Just noticed documentation on returned values is missing here!

scil-admin:

scilpy/image/datasets.py
EmmaRenauld: I added one. For now it is not used-defined (self.origin='corner') but at least a bit more obvious?

CHrlS98 (approved):
Alright nice!

I will open a PR in the future to clarify the `DataVolume`, `SeedGenerator` coordinate system extravaganza.

scil-admin:

scil-admin:

PR #517, opened 2021-11-02
Fix reslice issue

GuillaumeTh:
Finally I fixed the resampling issue that we had. Now the offset is always ok. I tested it with iso and aniso voxels and worked on my side. Please test it with your own data.

The fix comes from mrtrix resampling script.


GuillaumeTh:
@AntoineTheb  could you test it too ?

scil-admin:

arnaudbore (approved):
Perfect !
Thank you @GuillaumeTh  !

PR #518, opened 2021-11-02
ENH: Avoid moving the original sft when cutting invalid streamlines

jhlegarreta:
Avoid moving the original sft when cutting invalid streamlines: save the
original space/origin, and move the streamlines back to the original
space/origin.

### DIFF
--- a/scilpy/utils/streamlines.py
+++ b/scilpy/utils/streamlines.py
@@ -623,6 +623,11 @@ def cut_invalid_streamlines(sft):
     """
     if not len(sft):
         return sft, 0
+
+    # Keep track of the streamlines' original space/origin
+    space = sft.space
+    origin = sft.origin
+
     sft.to_vox()
     sft.to_corner()
 
@@ -674,5 +679,12 @@ def cut_invalid_streamlines(sft):
     new_sft = StatefulTractogram.from_sft(new_streamlines, sft,
                                           data_per_streamline=new_data_per_streamline,
                                           data_per_point=new_data_per_point)
+
+    # Move the streamlines back to the original space/origin
+    sft.to_space(space)
+    sft.to_origin(origin)
+
+    new_sft.to_space(space)
+    new_sft.to_origin(origin)
 
     return new_sft, cutting_counter


scil-admin:

scilpy/utils/streamlines.py
```diff
@@ -675,4 +681,8 @@ def cut_invalid_streamlines(sft):
                                           data_per_streamline=new_data_per_streamline,
```
GuillaumeTh, line 681: In the from_sft, the sft must be copy_sft.

scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-518/2/

scil-admin:

scilpy/utils/streamlines.py
```diff
@@ -675,4 +680,11 @@ def cut_invalid_streamlines(sft):
...
                                           data_per_point=new_data_per_point)
 
+    # Move the streamlines back to the original space/origin
+    sft.to_space(space)
+    sft.to_origin(origin)
+
```
arnaudbore, line 686: do we need sft here ?

jhlegarreta: Yes, we do. Changing the original space/origin to a new space/origin inside the method and not changing it back before returning leaves the sft in the new space/origin. That same sft object might be used for further processing (even if it is to use its header properties) back at the caller script, which will lead to unexpected results.

GuillaumeTh: I agree with Jon. The only question that I have is why do you revert changing the sft in the from_sft call ?

jhlegarreta: > The only question that I have is why do you revert changing the sft in the from_sft call ?

So that is linked to [this comment](https://github.com/scilus/scilpy/pull/518/files#r741209633). The suggestion was tried in commit 09dcc55 and tests were failing: https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-518/2/. I did not dig further.

Let me know if you think something else other is required to avoid the failing tests.

GuillaumeTh (approved):

PR #529, opened 2021-12-13
Optional uniformize

frheault:
Following Laurent's problem with the script, this is a simple fix.

Laurent was manually uniformizing his bundle and doing QC on the results. He was then using scil_compute_bundle_voxel_label_map.py and also doing QC.

Then when he was using this script, it was automatically flipping head/tail (originally for simplicity) rendering all his QC useless.

Now, turning off the uniformization/flipping is possible.

### DIFF
--- a/scripts/scil_evaluate_bundles_individual_measures.py
+++ b/scripts/scil_evaluate_bundles_individual_measures.py
@@ -25,6 +25,7 @@ If data comparison is performed, the bundles MUST be in same resolution.
 """
 
 import argparse
+import itertools
 import json
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -76,10 +81,12 @@ def _build_arg_parser():
     return p
 
 
-def compute_measures(filename_tuple):
+def compute_measures(args):
+    filename_tuple, no_uniformize = args
     sft = load_tractogram(filename_tuple[0], filename_tuple[1])
     _, dimensions, voxel_size, _ = sft.space_attributes
-    uniformize_bundle_sft(sft)
+    if no_uniformize:
```
arnaudbore, line 88: Opposite non ?

scripts/scil_evaluate_bundles_individual_measures.py
```diff
@@ -68,6 +69,10 @@ def _build_arg_parser():
     p.add_argument('--group_statistics', action='store_true',
                    help='Show average measures \n'
                         '[%(default)s].')
+
+    p.add_argument('--no_uniformize', action='store_false',
```
arnaudbore, line 73: don't like the double negation here...

scil-admin:

PR #530, opened 2021-12-13
ENH: Remove unused default process count parameter

jhlegarreta:
Remove unused default process count parameter from the process count
validation method.

Take advantage of the commit to:
- Document the method behavior if the process count is not contained in
the arguments.
- Fix the returned value docstring.
- Fix the logged process count in script.

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -206,9 +206,12 @@ def add_sh_basis_args(parser, mandatory=False):
                         help=help_msg)
 
 
-def validate_nbr_processes(parser, args, default_nbr_cpu=None):
+def validate_nbr_processes(parser, args):
     """ Check if the passed number of processes arg is valid.
-    If not valid (0 < nbr_cpu_to_use <= cpu_count), raise parser.error.
+    Valid values are considered to be in the [0, CPU count] range:
+        - Raises a parser.error if an invalid value is provided.
+        - Returns the maximum number of cores retrieved if no value (or a value
+        of 0) is provided.
 
     Parameters
     ----------
@@ -216,12 +219,10 @@ def validate_nbr_processes(parser, args, default_nbr_cpu=None):
         Parser as created by argparse.
     args: argparse namespace
         Args as created by argparse.
-    default_nbr_cpu: int (or None)
-        Number of cpu to use, default is cpu_count (all).
 
-    Results
-    ------
-    nbr_cpu
+    Returns
+    -------
+    nbr_cpu: int
         The number of CPU to be used.
     """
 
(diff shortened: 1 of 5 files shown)


scilpy/io/utils.py
```diff
@@ -206,22 +206,21 @@ def add_sh_basis_args(parser, mandatory=False):
                         help=help_msg)
 
 
-def validate_nbr_processes(parser, args, default_nbr_cpu=None):
+def validate_nbr_processes(parser, args):
```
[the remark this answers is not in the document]
jhlegarreta: Added in e35494a.

scilpy/io/utils.py
```diff
@@ -206,22 +206,22 @@ def add_sh_basis_args(parser, mandatory=False):
...
     """ Check if the passed number of processes arg is valid.
-    If not valid (0 < nbr_cpu_to_use <= cpu_count), raise parser.error.
+    If not valid (0 < nbr_cpu_to_use <= cpu_count), raise parser.error. If no
```
CHrlS98, line 211:
```suggestion
    If not valid (0 <= args.nbr_processes <= cpu_count), raise parser.error. If no
```

CHrlS98 (commented):
Should we also make this change?

scilpy/io/utils.py
jhlegarreta: Done in 8fa6548.

scilpy/io/utils.py
```diff
@@ -206,22 +206,22 @@ def add_sh_basis_args(parser, mandatory=False):
                         help=help_msg)
 
 
-def validate_nbr_processes(parser, args, default_nbr_cpu=None):
+def validate_nbr_processes(parser, args):
     """ Check if the passed number of processes arg is valid.
-    If not valid (0 < nbr_cpu_to_use <= cpu_count), raise parser.error.
+    If not valid (0 < args.nbr_processes <= cpu_count), raise parser.error. If
```
CHrlS98, line 211: Should we not replace `<` by `<=`, since `args.nbr_processes = 0` does not raise a parser.error ?

CHrlS98 (commented):
One last comment. Otherwise, all looks good.

scilpy/io/utils.py
jhlegarreta: Not sure if I get it:  as you say, `args.nbr_processes = 0` does NOT raise any error, and thus it is a valid value, isn't it?

CHrlS98: Yes, but the range in parentheses seems to refers to valid values, and as it as right now 0 is excluded from it.

jhlegarreta: OK, I see. Looks like the wording could be improved: I grant that the condition within the brackets is not right if mean the valid values, but reading "If not valid", I'd assume that the values within the brackets refer to not valid/invalid values, which is not the case. Re-worked the thing in e733d7d. Also realized that the `nbr_cpu == 0` will never happen.

CHrlS98 (approved):
Cool, I like the improved docstring!

scilpy/io/utils.py
jhlegarreta: Resolving following https://github.com/scilus/scilpy/pull/530#pullrequestreview-833433073.

Thanks for the suggestions @CHrlS98.

arnaudbore (approved):

PR #541, opened 2022-01-24
compute_connectivity fix average

frheault:
Switch the computation of the average metrics in a bundle, use the same as in scil_compute_mean_std.py for identical results.

The previous method gave weird results when outliers were present in the maps.


scil-admin:

arnaudbore (approved):

PR #557, opened 2022-03-09
Fixing doc in io.utils

EmmaRenauld:
Fixing some doc:

1) `args: list` raises a warning in editors such as pycharm, as it is actually a Namespace.
2) missing parameters


scil-admin:

PR #559, opened 2022-03-09
Fix vf stuff

karanphil:
There was a minor bug with the `vf_rgb` when nan values occured in `shm_coeff`. In the script `scil_compute_msmt_fodf.py`, we changed any nan for 0 in `shm_coeff`, but forgot to do so in `vf`, so that when calculating `vf_rgb`, we were dividing by nan, resulting in a `vf_rgb` full of nan. This has been fixed by a simple 2 lines change.

Also, the parameter `bg_range` in `scil_visualize_fodf.py` was not working for rgb background. The value_range parameter was removed when calling `actor.slice` and instead, we now apply the value_range manually on the texture before sending it to `actor.slice`. This works for both scalar and rgb data.

### DIFF
--- a/scilpy/viz/scene_utils.py
+++ b/scilpy/viz/scene_utils.py
@@ -154,6 +154,11 @@ def create_texture_slicer(texture, mask, slice_index, value_range=None,
     else:
         masked_texture = texture
 
+    if value_range:
+        masked_texture = (masked_texture - value_range[0])
+                         / value_range[1] * 255
+        masked_texture = np.clip(masked_texture, 0 , 255)
+
     slicer_actor = actor.slicer(masked_texture, affine=affine,
                                 value_range=value_range,
                                 opacity=opacity,
--- a/scilpy/viz/scene_utils.py
+++ b/scilpy/viz/scene_utils.py
@@ -28,8 +28,6 @@ def initialize_camera(orientation, slice_index, volume_shape):
 
     Parameters
     ----------
-    slicer_actor : actor
-        Slicer actor from Fury
     orientation : str
         Name of the axis to visualize. Choices are axial, coronal and sagittal.
     slice_index : int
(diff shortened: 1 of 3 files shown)


scil-admin:

scilpy/viz/scene_utils.py
```diff
@@ -154,10 +154,12 @@ def create_texture_slicer(texture, mask, slice_index, value_range=None,
     else:
         masked_texture = texture
```
CHrlS98, line 155: Can we mask the texture in-place instead of copying the data to a new variable `masked_texture` while you're at it?

scilpy/viz/scene_utils.py
```diff
@@ -25,6 +25,22 @@ class CamParams(Enum):
 def initialize_camera(orientation, slice_index, volume_shape):
     """
     Initialize a camera for a given orientation.
+
+    Parameters
+    ----------
+    slicer_actor : actor
```
arnaudbore, line 31: No slice actor

scilpy/viz/scene_utils.py
```diff
@@ -93,14 +120,50 @@ def set_display_extent(slicer_actor, orientation, volume_shape, slice_index):
...
+                      scale, radial_scale, norm, colormap):
     """
     Create a ODF slicer actor displaying a fODF slice. The input volume is a
     3-dimensional grid containing the SH coefficients of the fODF for each
     voxel at each voxel, with the grid dimension having a size of 1 along the
     axis corresponding to the selected orientation.
+
+    Parameters
+    ----------
+    texture : np.ndarray (3d or 4d)
```
arnaudbore, line 134: sh_odf not a texture

arnaudbore (approved):
Few comments otherwise LGTM

CHrlS98 (approved):
I'm okay with these changes!

PR #574, opened 2022-04-08
Improve DOC on scil_resample_tractogram.py

frheault:
Make it clearer on the default behavior.
New parameters make it easier to subsample a lot of tractogram using bash but without upsampling the small ones.


scil-admin:

PR #581, opened 2022-05-17
Fixing coordinate type + decreasing test size in dev tracking

EmmaRenauld:
Small fix: coordinate type was a tuple when expected to be an array. Script was still running, but streamline contained one tuple and n-1 arrays. 

Also changing test from --nt 1000 to --nt 10. Decreasing this test time from 46 seconds to 2 seconds on my computer.


EmmaRenauld:
@CHrlS98 , easy review!

scil-admin:

CHrlS98 (approved):
Not tested, but I'm okay with the changes :+1:

PR #585, opened 2022-06-01
Small changes in the dev tracker

EmmaRenauld:
- Encapsulated parts for easier management with dwi_ml.
 (Multiprocessing does not need to save data to tmpdir and load it back because we use hdf5 files with lazy use, which deal well with multiprocessing).
 
- Made the processing frequency param (this way, I can have more prints in dwi_ml).
 
- Added docstring.
 
- Removed unused method in the propagator.
 
- Removed the comment #todo must be better designed for dipy: I don't thing we intend to change this.

- Sending line instead of line[-1] to propagator. In our case, propagator sometimes needs to know where you are coming from.

### DIFF
--- a/scilpy/tracking/tracker.py
+++ b/scilpy/tracking/tracker.py
@@ -370,7 +370,7 @@ class Tracker(object):
         propagation_can_continue = True
         while len(line) < self.max_nbr_pts and propagation_can_continue:
             new_pos, new_tracking_info, is_direction_valid = \
-                self.propagator.propagate(line[-1], tracking_info)
+                self.propagator.propagate(line, tracking_info)
 
             # Verifying and appending
             if is_direction_valid:
--- a/scilpy/tracking/tracker.py
+++ b/scilpy/tracking/tracker.py
@@ -16,9 +16,9 @@ from scilpy.tracking.propagator import AbstractPropagator, PropagationStatus
 from scilpy.tracking.seed import SeedGenerator
 
 # For the multi-processing:
-# Iterable. Will contain all parameters necessary for a sub-process
+# Dictionary. Will contain all parameters necessary for a sub-process
 # initialization.
-multiprocess_init_args = []
+multiprocess_init_args = {}
 
 
 class Tracker(object):
@@ -184,7 +184,10 @@ class Tracker(object):
         pool = multiprocessing.Pool(
             self.nbr_processes,
             initializer=self._send_multiprocess_args_to_global,
-            initargs=(data_file_name, self.mmap_mode))
+            initargs={
+                'data_file_name': data_file_name,
+                'mmap_mode': self.mmap_mode
+            })
 
         return pool
 
@@ -236,7 +239,7 @@ class Tracker(object):
             (file where the data is saved, mmap_mode).
         """
         self.propagator.reset_data(np.load(
-            init_args[0], mmap_mode=init_args[1]))
+            init_args['data_file_name'], mmap_mode=init_args['mmap_mode']))
 
     def _get_streamlines(self, chunk_id):
         """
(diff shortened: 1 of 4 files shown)


EmmaRenauld:
@CHrlS98 No real changes. Should be easy to review. Thanks!

scil-admin:

scilpy/tracking/tracker.py
```diff
@@ -16,10 +16,14 @@
...
 
 # For the multi-processing:
-data_file_info = None
+# Iterable. Will contain all parameters necessary for a sub-process
+# initialization.
+multiprocess_init_args = []
 
 
 class Tracker(object):
+    printing_frequency = 1000
```
CHrlS98, line 25: Is there a reason why this attribute isn't under `__init__`?

scilpy/tracking/tracker.py
```diff
@@ -158,15 +151,59 @@ def _set_nbr_processes(self, nbr_processes):
...
+        -------
+        pool: The multiprocessing pool.
+        """
+        # Using pool with a class method will serialize all parameters
+        # in the class, which can be heavy, but it is what we would be
+        # doing manually with a static class.
+        # Be careful however, parameter changes inside the method will
+        # not be kept.
+        # toDo
+        # must be better designed for dipy
```
CHrlS98, line 177: I think this todo is legacy, because the port to `dipy` has already been made.

scilpy/tracking/tracker.py
```diff
@@ -16,10 +16,14 @@
 from scilpy.tracking.seed import SeedGenerator
 
 # For the multi-processing:
-data_file_info = None
+# Iterable. Will contain all parameters necessary for a sub-process
+# initialization.
+multiprocess_init_args = []
```
CHrlS98, line 21: Couldn't it be a dictionary? Like, set to `None` here but pass a `dict` to the pool `init_args`?

CHrlS98 (commented):
Looks good. Indeed not many changes. I haven't tested but here are my comments.

scil-admin:

scilpy/tracking/tracker.py
CHrlS98: Just to increase the code readability. I just feel like having `np.load(init_args['data'], init_args['mmap_mode'])` would be clearer. But I won't defend my point with my life either, your call.

scil-admin:

scilpy/tracking/utils.py
```diff
@@ -79,12 +79,13 @@ def add_out_options(p):
 
 
 def verify_streamline_length_options(parser, args):
-    if not args.min_length > 0:
-        parser.error('min_length must be > 0, {}mm was provided.'
+    if not args.min_length >= 0:
```
CHrlS98, line 82: Do we enforce minimum number of points of 2 somewhere? (Less than 2 points is hardly a streamline.)

EmmaRenauld: No, not yet. But the default parameter is 10 mm. So 1-point streamlines would only be accepted if the user manually specifies --min_length 0 (which can be useful when debugging something. You can even verify the number of streamlines of only one point that were created after).

CHrlS98: I'm just worried about the saving/loading of streamlines with 0 point (empty list) or 1 point. Do we know if our tools support these? If not, should we include some sort of warning when `min_length / step_size < 1`?

scil-admin:

scilpy/tracking/utils.py
CHrlS98: oh that's right! Ok it might not matter. It's the user's problem in the end. If he does want it to happen he should set `min_length = step_size`.

arnaudbore (approved):
LGTM

PR #586, opened 2022-06-01
Adding check for ints

karanphil:
Really small PR. Quick fix for the function `transform_anatomy` from utils/image.py. Previously, it was forcing the input file to be loaded in float32 with `get_fdata(dtype=np.float32)`, and later the output was saved using the dtype of the loaded file, which was always float32 regardless of the "real" input dtype. I added a check for uint or int, which is followed by `moving_data = np.asanyarray(nib_file.dataobj).astype(curr_type)` if True, and `moving_data = nib_file.get_fdata(dtype=curr_type)` if False, where `curr_type = nib_file.get_data_dtype()`.

I was getting this problem when using `scil_reshape_to_reference.py` with labels (.mgz) in int32. The output (.nii.gz) ended up in float32 and `scil_split_volume_by_ids.py` didn't work.

You can test this either directly with the function `transform_anatomy` or by calling `scil_reshape_to_reference.py` with `in_file` being an int or a float (testing both) file.

### DIFF
--- a/scilpy/utils/image.py
+++ b/scilpy/utils/image.py
@@ -42,7 +42,12 @@ def transform_anatomy(transfo, reference, moving, filename_to_save,
     static_data = nib.load(reference).get_fdata(dtype=np.float32)
 
     nib_file = nib.load(moving)
-    moving_data = nib_file.get_fdata(dtype=np.float32)
+    curr_type = nib_file.get_data_dtype()
+    if np.issubdtype(curr_type, np.signedinteger) or \
+       np.issubdtype(curr_type, np.unsignedinteger):
+        moving_data = np.asanyarray(nib_file.dataobj).astype(curr_type)
+    else:
+        moving_data = nib_file.get_fdata(dtype=curr_type)
     moving_affine = nib_file.affine
 
     if moving_data.ndim == 3 and isinstance(moving_data[0, 0, 0],


scil-admin:

frheault:
I would put that as a parameter. Most transformation tool switch to float to accommodate interpolation.

It shouldn't automatically revert to int, only if required.
Or vice versa, if it does it automatically it should be possible to turn off.

karanphil:
> I would put that as a parameter. Most transformation tool switch to float to accommodate interpolation.
> 
> It shouldn't automatically revert to int, only if required. Or vice versa, if it does it automatically it should be possible to turn off.

Then I think I would let it as it was before and use scil_image_math convert to change the dtype of my data.

I just felt like it was weird to enforce the dtype and save the data with said dtype after, like if it was the "original dtype", even if it's not at all.

scil-admin:

scil-admin:

scilpy/utils/image.py
```diff
@@ -37,12 +37,19 @@ def transform_anatomy(transfo, reference, moving, filename_to_save,
...
+    keep_dtype : bool
+        If True, keeps the data_type of the input moving image when saving
+        the output image
     """
     grid2world, dim, _, _ = get_reference_info(reference)
     static_data = nib.load(reference).get_fdata(dtype=np.float32)
 
     nib_file = nib.load(moving)
-    moving_data = nib_file.get_fdata(dtype=np.float32)
+    curr_type = nib_file.get_data_dtype()
```
arnaudbore, line 48: put curr_type in the if.

arnaudbore (commented):
Quick comment.
Can you check if one of the scripts that uses transform_anatomy uses a dtype different from float32 in the tests and if it's the case can you add a test with keep_dtype=True.

PR #589, opened 2022-06-23
[BF] Fix arg in invalid streamlines + warning in todi

arnaudbore:


scilpy/tractanalysis/todi_util.py
```diff
@@ -85,7 +85,13 @@ def streamlines_to_pts_dir_norm(streamlines, n_steps=1, asymmetric=False):
     seg_dir, seg_norm = get_segments_dir_and_norm(segments,
                                                   seg_mid,
                                                   asymmetric)
-    return seg_mid, seg_dir, seg_norm
+
+    mask = seg_norm > 1.0e-20
```
CHrlS98, line 89: is this value big enough to catch anything?

CHrlS98 (commented):
Just the one question

scil-admin:

scil-admin:

scil-admin:

scilpy/tractanalysis/todi_util.py
arnaudbore: Here the only purpose is to make sure norm !=0 otherwise people should use remove_invalid_streamlines with a threshold they can set.

scripts/scil_remove_invalid_streamlines.py
```diff
@@ -40,6 +40,9 @@ def _build_arg_parser():
                    help='Consider single point streamlines invalid.')
     p.add_argument('--remove_overlapping_points', action='store_true',
                    help='Consider streamlines with overlapping points invalid.')
+    p.add_argument('--threshold', type=float, default=0.001,
```
arnaudbore, line 43: @frheault does this default threshold makes sense to you ? 
Should we add an option to just remove overlapping points ?

scil-admin:

CHrlS98 (approved):

scilpy/tractanalysis/todi_util.py
frheault: Did it at least solve the use case you had (i.e. is it big enough to solve one know case) ?

scripts/scil_remove_invalid_streamlines.py
frheault: Since our smallest ever use step-size is 0.1, I could accept 0.01 (1/10 the smallest step-size we use).

But since almost all instance of overlapping point are perfect zero due to algorithmic error, this should catch everything anyway.

frheault (approved):

PR #590, opened 2022-06-27
ENH: Adding QBx option in subsampling + cleaning

EmmaRenauld:
I am suggesting a new option in both scripts scil_resample_tractogram and scil_split_tractogram: to use quickbundles to ensure that the random subsampling gets a bit of every cluster present, proportionnally to their initial size.

Other changes: 
- cleaning, in particular, in many places, using `sft[indices]` instead of `from_sft(...)`.
- CHANGING DEFAULT BEHAVIOUR of split_tractogram. By default, it was splitting sequentially, not randomly. Changed that. Do you agree? Sequential splitting is still offered, as option.
- more verbose.

Other ideas that I didn't change, in case the scripts are included somewhere:
- the name split_tractograms should be split_tractogram. It only splits one tractogram. Do you want me to change that?
- the option `--nb_chunk` should be `--nb_chunks`, in my opinion. Do you want me to change that?


scil-admin:

arnaudbore:
> I am suggesting a new option in both scripts scil_resample_tractogram and scil_split_tractogram: to use quickbundles to ensure that the random subsampling gets a bit of every cluster present, proportionnally to their initial size.
> 
> Other changes:
> 
> * cleaning, in particular, in many places, using `sft[indices]` instead of `from_sft(...)`.
> * CHANGING DEFAULT BEHAVIOUR of split_tractogram. By default, it was splitting sequentially, not randomly. Changed that. Do you agree? Sequential splitting is still offered, as option.

I would need some input from others @CHrlS98  @frheault . I don't see any issue here.

> * more verbose.
> 
> Other ideas that I didn't change, in case the scripts are included somewhere:
> 
> * the name split_tractograms should be split_tractogram. It only splits one tractogram. Do you want me to change that?
Yes
> * the option `--nb_chunk` should be `--nb_chunks`, in my opinion. Do you want me to change that?
Yes

scil-admin:

arnaudbore (approved):
A couple of comments otherwise LGTM

scil-admin:

scripts/tests/test_split_tractogram.py
```diff
@@ -0,0 +1,35 @@
...
+    ret = script_runner.run('scil_split_tractogram.py', '--help')
+    assert ret.success
+
+
+def test_execution_tracking(script_runner):
+    os.chdir(os.path.expanduser(tmp_dir.name))
+    in_tracto = os.path.join(get_home(), 'tracking',
+                             'local.trk')
+    ret = script_runner.run('scil_split_tractogram.py', in_tracto,
+                            'local_split_', '--nb_chunks', '3', '-f')
```
arnaudbore, line 24: Please remove the `_` here and for the other defs. It explains my confusion about the `_`

scil-admin:

scil-admin:

CHrlS98:
@arnaudbore @EmmaRenauld 
About: 
> CHANGING DEFAULT BEHAVIOUR of split_tractogram. By default, it was splitting sequentially, not randomly. Changed that. Do you agree? Sequential splitting is still offered, as option.

I don't think I have ever used `scil_split_tractogram`, but I don't suppose any flow relies on the fact that the split is done sequentially (I'd hope that's not the case at least!). Having random split as default behaviour makes perfect sense to me!

scilpy/tracking/tools.py
```diff
@@ -147,28 +149,187 @@ def get_subset_streamlines(sft, max_streamlines, rng_seed=None):
...
+    else:
+        sft_left = orig_sft[:]  # Equivalent of deep copy
+
+    for i in range(nb_chunks):
+        nb_to_keep = chunk_sizes[i]
+
+        sub_ind = get_subset_streamlines(
+            sft_left, nb_to_keep, seed, return_indices_only=True)
+
+        final_indices[i] = [orig_indices_left[ind] for ind in sub_ind]
```
CHrlS98, line 245: why not append? It would remove the list initialization bit at line 232.

scilpy/tracking/tools.py
```diff
@@ -147,28 +149,187 @@ def get_subset_streamlines(sft, max_streamlines, rng_seed=None):
...
+        sfts.append(orig_sft[curr:curr + nb_str])
+        curr += chunk_sizes[i]
+
+    return sfts
+
+
+def get_n_subsets_streamlines_random(orig_sft, chunk_sizes, seed,
+                                     return_indices_only=False):
+    """
+    Get n chunks of subsets of streamlines.
```
CHrlS98, line 207: `n` is `len(chunk_sizes)`? Maybe we should think of a better name if that's the case? Like `random_split_streamlines` or something? Also isn't `get_subset_streamlines` the `n=1` case? Could we merge both methods?

scilpy/tracking/tools.py
```diff
@@ -147,28 +149,187 @@ def get_subset_streamlines(sft, max_streamlines, rng_seed=None):
...
+    orig_indices_left = np.arange(len(orig_sft))
+    if return_indices_only:
+        sft_left = orig_sft  # No need to keep orig_sft intact.
+    else:
+        sft_left = orig_sft[:]  # Equivalent of deep copy
+
+    for i in range(nb_chunks):
+        nb_to_keep = chunk_sizes[i]
+
+        sub_ind = get_subset_streamlines(
```
CHrlS98, line 242: you could shuffle ids just once no? and create your chunks from these indices?

scripts/scil_resample_tractogram.py
```diff
@@ -116,7 +148,14 @@ def main():
             args.point_wise_std, args.streamline_wise_std,
             args.gaussian, args.spline, args.seed)
     elif args.nb_streamlines < original_number:
-        sft = get_subset_streamlines(sft, args.nb_streamlines, args.seed)
+        if args.downsample_per_cluster:
+            # output contains rejected streamlines, we don't use them.
+            sft, _ = get_n_subsets_streamlines_per_cluster(
+                sft, [args.nb_streamlines], args.seed, args.qbx_thresholds)
+            logging.debug("Kept {} out of expected {} streamlines."
```
CHrlS98, line 155:
```suggestion
            logging.debug("Kept {} out of {} expected streamlines."
```

scilpy/tracking/tools.py
EmmaRenauld: > n is len(chunk_sizes)? Maybe we should think of a better name if that's the case? Like random_split_streamlines or something?

I tried updating the doc and names. Tell me if it's clearer.

> Also isn't get_subset_streamlines the n=1 case? Could we merge both methods?

Yes it is; `get_n_subsets _random` called `get_subset_streamlines` n times. But also, it is used in scil_resample_tractogram, so I could merge it but it would return a list of one subset instead of only the subset, and I should change scil_resample_tractogram to get the first result. Do you want me to do that?

CHrlS98: Maybe we should get a second opinion before you start refactoring ? @arnaudbore @frheault

scilpy/tracking/tools.py
EmmaRenauld: Yes I could. I was just re-using the already-existing method.  I'll change that.

scil-admin:

PR #593, opened 2022-07-05
Clarify vox space

EmmaRenauld:
So far in our script, we coded everything in voxmm, and use the `voxmm_to_...` methods of the DatasetVolume. I needed the equivalent methods starting from vox space for dwi_ml.  So, base idea of this PR: to add methods in the DatasetVolume such as `vox_to_voxmm`, `is_vox_in_bound`, etc.

But, looking back at the code made me realise there were some incoherencies We called 'vox' space something that did not depend on the origin, which does not fit with dipy's usage. Clarified those methods names. 

Another change:  Other scripts such as scil_compute_seed_density_maps use seeds in vox space, center origin. Modified.

Note. Summary:  
- dipy uses voxmm, center. 
- scil_compute_local_tracking uses voxmm, corner
- scripts using seeds use vox, center.
- dwi_ml uses vox, corner (because torch's trilinear interpolation method uses vox space).
Could be dangerous to get mixed up. But so far, doc is added, and  self.space, self.origin variables have been added for clarity.

### DIFF
--- a/scilpy/image/datasets.py
+++ b/scilpy/image/datasets.py
@@ -46,7 +46,7 @@ class DataVolume(object):
         self.dim = self.data.shape[0:4]
         self.nbr_voxel = self.data.size
 
-    def get_voxel_value(self, i, j, k):
+    def get_value_at_idx(self, i, j, k):
         """
         Get the voxel value at x, y, z in the dataset.
         If the coordinates are out of bound, the nearest voxel value is taken.
@@ -61,22 +61,17 @@ class DataVolume(object):
         value: ndarray (self.dim[-1],)
             The value evaluated at voxel x, y, z.
         """
-        if not self.is_voxel_in_bound(i, j, k):
-            i = max(0, min(self.dim[0] - 1, i))
-            j = max(0, min(self.dim[1] - 1, j))
-            k = max(0, min(self.dim[2] - 1, k))
-
+        i, j, k = self._idx_to_closest_coord(i, j, k)
         return self.data[i][j][k]
 
-    def is_voxel_in_bound(self, i, j, k):
+    def is_idx_in_bound(self, i, j, k):
         """
         Test if voxel is in dataset range.
 
         Parameters
         ----------
-        i, j, k: ints or floats
-            Voxel indice along each axis (as ints) or voxel coordinates in
-            voxel world (as floats).
+        i, j, k: ints
+            Voxel indice along each axis (as ints).
 
         Return
         ------
@@ -86,6 +81,139 @@ class DataVolume(object):
         return (0 <= i < (self.dim[0]) and
                 0 <= j < (self.dim[1]) and
                 0 <= k < (self.dim[2]))
+
+    def _idx_to_closest_coord(self, i, j, k):
+        if not self.is_idx_in_bound(i, j, k):
+            i = max(0, min(self.dim[0] - 1, i))
+            j = max(0, min(self.dim[1] - 1, j))
+            k = max(0, min(self.dim[2] - 1, k))
+        return i, j, k
+
+    def _vox_to_closest_coord(self, x, y, z, origin):
+        """
+        In voxel space, get closest coordinate in the voxel in out of bout.
+        Else, return initial coordinates.
+
+        Parameters
+        ----------
+        x, y, z: floats
+            Current coordinates (vox).
+        origin: str
+            'center' or 'corner'.
+
+        Return
+        ------
+        x, y, z: floats
+            Closest valid coordinates.
+        """
+        if not self.is_vox_in_bound(x, y, z, origin):
+            eps = float(1e-8)  # Epsilon to exclude upper borders
+            if origin == 'corner':
+                # Min = 0
+                # Max = 0.9999 in first voxel.
+                x = max(0, min(self.dim[0] - eps, x))
+                y = max(0, min(self.dim[1] - eps, y))
+                z = max(0, min(self.dim[2] - eps, z))
+            elif origin == 'center':
+                # Min = -0.5
+                # Max = 0.499999 in first voxel.
+                x = max(-0.5, min(self.dim[0] - 0.5 - eps, x))
+                y = max(-0.5, min(self.dim[1] - 0.5 - eps, y))
+                z = max(-0.5, min(self.dim[2] - 0.5 - eps, z))
+            else:
+                raise ValueError("Origin should be 'center' or 'corner'.")
+        return x, y, z
+
+    @staticmethod
+    def vox_to_idx(x, y, z, origin):
+        """
+        Get the 3D indice of the closest voxel at position x, y, z expressed
+        in mm.
+
+        Parameters
+        ----------
+        x, y, z: floats
+            Position coordinate in voxel space along x, y, z axis.
+        origin: str
+            'center' or 'corner'.
+
+        Return
+        ------
+        out: list
+            3D indice of voxel at position x, y, z.
+        """
+        if origin == 'corner':
+            return np.floor((x, y, z))
+        elif origin == 'center':
+            return np.floor((x + 0.5, y + 0.5, z + 0.5))
+        else:
+            raise ValueError("Origin must be 'center' or 'corner'.")
+
+    def vox_to_value(self, x, y, z, origin):
+        """
+        Get the voxel value at voxel position x, y, z (vox) in the dataset.
+        If the coordinates are out of bound, the nearest voxel value is taken.
+        Value is interpolated based on the value of self.interpolation.
+
+        Parameters
+        ----------
+        x, y, z: floats
+            Position coordinate (vox) along x, y, z axis.
+        origin: str
+            'center' or 'corner'.
+
+        Return
+        ------
+        value: ndarray (self.dims[-1],) or float
+            Interpolated value at position x, y, z (mm). If the last dimension
+            is of length 1, return a scalar value.
+        """
+        if self.interpolation is not None:
+            # Checking if out of bound.
+            x, y, z = self._vox_to_closest_coord(x, y, z, origin)
+
+            # Interpolation: Using dipy's pyx methods. The doc can be found in
+            # the file dipy.core.interpolation.pxd. Dipy works with origin
+            # center.
+            coord = np.array((x, y, z), dtype=np.float64)
+            if origin == 'corner':
+                coord -= 0.5
+
+            if self.interpolation == 'nearest':
+                # They use round(point), not floor. This is the equivalent of
+                # origin = 'center'.
+                result = nearestneighbor_interpolate(self.data, coord)
+            else:
+                # Trilinear
+                # They do not say it explicitly but they verify if
+                # point[i] < -.5 or point[i] >= (data.shape[i] - .5),
+                # meaning that they work with origin='center'.
+                result = trilinear_interpolate4d(self.data, coord)
+
+            # Squeezing returns only value instead of array of length 1 if 3D
+            # data
+            return np.squeeze(result)
+        else:
+            raise Exception("No interpolation method was given, cannot run "
+                            "this method..")
+
+    def is_vox_in_bound(self, x, y, z, origin):
+        """
+        Test if voxel is in dataset range.
+
+        Parameters
+        ----------
+        x, y, z: floats
+            Voxel coordinates along each axis in voxel space.
+        origin: str
+            Origin ('center' or 'corner').
+
+        Return
+        ------
+        out: bool
+            True if voxel is in dataset range, False otherwise.
+        """
+        return self.is_idx_in_bound(*self.vox_to_idx(x, y, z, origin))
 
     def voxmm_to_idx(self, x, y, z, origin):
         """
@@ -97,18 +225,16 @@ class DataVolume(object):
         x, y, z: floats
             Position coordinate (mm) along x, y, z axis.
         origin: str
-            'center': Voxel 0,0,0 goes from [-resx/2, -resy/2, -resz/2] to
-                [resx/2, resy/2, resz/2].
-            'corner': Voxel 0,0,0 goes from [0,0,0] to [resx, resy, resz].
+            'center' or 'corner'.
 
         Return
         ------
-        out: list
+        x, y, z: ints
             3D indice of voxel at position x, y, z.
         """
-        return np.floor(self.voxmm_to_vox(x, y, z, origin))
+        return self.vox_to_idx(*self.voxmm_to_vox(x, y, z), origin)
 
-    def voxmm_to_vox(self, x, y, z, origin):
+    def voxmm_to_vox(self, x, y, z):
         """
         Get voxel space coordinates at position x, y, z (mm).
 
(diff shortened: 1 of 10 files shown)


scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scil-admin:

scilpy/image/datasets.py
```diff
@@ -87,71 +82,85 @@ def is_voxel_in_bound(self, i, j, k):
                 0 <= j < (self.dim[1]) and
                 0 <= k < (self.dim[2]))
 
-    def voxmm_to_idx(self, x, y, z, origin):
+    def _idx_to_closest_coord(self, i, j, k):
```
CHrlS98, line 85: Not sure about the name, gives the impression there is a rounding operation or something... Maybe `_clip_idx_to_bounds` ?

scilpy/image/datasets.py
```diff
@@ -87,71 +82,85 @@ def is_voxel_in_bound(self, i, j, k):
...
+            i = max(0, min(self.dim[0] - 1, i))
+            j = max(0, min(self.dim[1] - 1, j))
+            k = max(0, min(self.dim[2] - 1, k))
+        return i, j, k
+
+    def _vox_to_closest_coord(self, x, y, z, origin):
         """
-        Get the 3D indice of the closest voxel at position x, y, z expressed
-        in mm.
+        In voxel space, get closest coordinate in the voxel in out of bout.
```
CHrlS98, line 94: "in out of bout" ?

scilpy/image/datasets.py
```diff
@@ -87,71 +82,85 @@ def is_voxel_in_bound(self, i, j, k):
...
 
-    def voxmm_to_idx(self, x, y, z, origin):
+    def _idx_to_closest_coord(self, i, j, k):
+        if not self.is_idx_in_bound(i, j, k):
+            i = max(0, min(self.dim[0] - 1, i))
+            j = max(0, min(self.dim[1] - 1, j))
+            k = max(0, min(self.dim[2] - 1, k))
+        return i, j, k
+
+    def _vox_to_closest_coord(self, x, y, z, origin):
```
CHrlS98, line 92: `_clip_vox_to_bounds` ?

scilpy/tracking/seed.py
```diff
@@ -57,23 +59,32 @@ def get_next_pos(self, random_generator, indices, which_seed):
...
+        r_x = random_generator.uniform(0, 1)
+        r_y = random_generator.uniform(0, 1)
+        r_z = random_generator.uniform(0, 1)
+
+        # Moving inside the voxel
+        x += r_x
+        y += r_y
+        z += r_z
+
+        assert self.origin == 'corner'
```
CHrlS98, line 81: Unlikely it would change since it is hardcoded. Maybe just make it extremely clear that the seed generator always returns seeds with origin corner? And remove the attribute `self.origin`?

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -186,6 +187,19 @@ def main():
     odf_sh_res = odf_sh_img.header.get_zooms()[:3]
     dataset = DataVolume(odf_sh_data, odf_sh_res, args.sh_interp)
 
+    if not np.array_equal(mask_res, odf_sh_res):
```
CHrlS98, line 190: Can we use `scilpy.io.image.assert_same_resolution` here and at L.193?

scripts/scil_compute_seed_density_map.py
```diff
@@ -58,10 +60,18 @@ def main():
...
     sft = load_tractogram(args.tractogram_filename, 'same',
                           bbox_valid_check=False)
-    # Streamlines are saved in RASMM but seeds are saved in VOX
-    # This might produce weird behavior with non-iso
+    # Streamlines are saved in RASMM by nibabel but seeds are saved in VOX,
+    # the default space in dipy.
+    # Also, origin should be center when creating the seeds (wee below, we
+    # are using round, not floor; works with center origin), which is the
+    # default in dipy.
+    # Changing sft here just to make sure, but it will not have consequences on
```
CHrlS98, line 68: If you can validate that this indeed doesn't change anything I think we should just remove to_vox and to_center below.

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -211,7 +225,15 @@ def main():
                   .format(len(streamlines), nbr_seeds, str_time))
 
     # save seeds if args.save_seeds is given
-    data_per_streamline = {'seeds': seeds} if args.save_seeds else {}
+    # We seeded (and tracked) in voxmm, corner, but in dipy, they use
+    # vox, center. In other scripts using the seeds (ex,
+    # scil_compute_density_map), this is what they will expect.
+    if args.save_seeds:
+        seeds = [seed / seed_res for seed in seeds]  # to_vox
+        seeds = [seed - 0.5 for seed in seeds]  # to_center
```
CHrlS98, line 233: you can scale and translate inside the same for loop to avoid iterating twice through all seeds.

scilpy/tracking/tracker.py
```diff
@@ -401,17 +401,28 @@ def _propagate_line(self, line, tracking_info):
...
+        # Space in voxmm in this class but verifying in case a child class
+        # allows a different space usage.
+        if self.space == 'voxmm':
+            if not self.mask.is_voxmm_in_bound(*last_pos, origin=self.origin):
+                return False
+
+            # Checking if out of mask
+            if self.mask.voxmm_to_value(*last_pos, origin=self.origin) <= 0:
+                return False
+        elif self.space == 'vox':
```
CHrlS98, line 419: should we make `space` a configurable parameter inside `__init__` function?

scilpy/image/datasets.py
```diff
@@ -160,40 +169,23 @@ def voxmm_to_value(self, x, y, z, origin):
...
-
-            coord = np.array(self.voxmm_to_vox(x, y, z, origin),
-                             dtype=np.float64)
+            # Checking if out of bound.
+            x, y, z = self._vox_to_closest_coord(x, y, z, origin)
 
             # Interpolation: Using dipy's pyx methods. The doc can be found in
             # the file dipy.core.interpolation.pxd. Dipy works with origin
             # center.
+            coord = np.array((x, y, z), dtype=np.float64)
```
CHrlS98, line 178: do we really need the precision of a float64?

CHrlS98 (commented):
Here are some comments on the code! I didn't test.

Also, do you plan on bringing your ML tracking code into scilpy at some point? It could help understanding how you use some of the methods/attributes in the tracking code.

EmmaRenauld:
> do you plan on bringing your ML tracking code into scilpy at some point? It could help understanding how you use some of the methods/attributes in the tracking code.

I'm not sure yet. Maybe some parts of the tracker for multiple tracking, but considering you already have your own organization, I didn't want to add a second version of GPU tracking on scilpy. We could discuss it one day.

scilpy/image/datasets.py
EmmaRenauld: Well there is a rounding operation; a floor option. (In vox_to_idx). How could the doc be clearer?

scilpy/image/datasets.py
EmmaRenauld: lol. oups

scilpy/image/datasets.py
EmmaRenauld: Ok, yes, I like it!

scilpy/tracking/seed.py
EmmaRenauld: I added a line to support corner instead of forcing it. Looks ok?

scripts/scil_compute_seed_density_map.py
EmmaRenauld: Ok, I'll check that soon.

scripts/scil_compute_local_tracking_dev.py
EmmaRenauld: thanks

scilpy/tracking/tracker.py
EmmaRenauld: Probably. Maybe in another PR to make sure it works now?

scilpy/image/datasets.py
EmmaRenauld: I don't think I'm the one who decided that. @arnaudbore , any opinion?

scil-admin:

scil-admin:

scil-admin:

scilpy/tracking/tracker.py
EmmaRenauld: Done!

scil-admin:

scil-admin:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-593/12/

scilpy/image/datasets.py
EmmaRenauld: Oups.... I tried, and it fails.

```
    result = trilinear_interpolate4d(self.data, coord)
  File "dipy/core/interpolation.pyx", line 357, in dipy.core.interpolation.trilinear_interpolate4d
ValueError: Buffer dtype mismatch, expected 'double' but got 'float'
```
I will revert and add a comment.

scil-admin:

scilpy/tracking/propagator.py
```diff
@@ -387,15 +413,17 @@ def _get_sf(self, pos):
         Parameters
         ----------
         pos: ndarray (3,)
-            Position in voxmm in the trackable dataset.
+            Position in the trackable dataset, either in vox or voxmm space.
```
CHrlS98, line 416: The "either in vox or voxmm space" is a little bit misleading. It gives the impression we can supply it in any space. I'd just remove it, the position is in the space the propagator is, and the user should know that.

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -225,12 +223,11 @@ def main():
...
     # save seeds if args.save_seeds is given
-    # We seeded (and tracked) in voxmm, corner, but in dipy, they use
-    # vox, center. In other scripts using the seeds (ex,
-    # scil_compute_density_map), this is what they will expect.
+    # We seeded (and tracked) in vox, center, which is what is expected for
+    # seeds.
     if args.save_seeds:
-        seeds = [seed / seed_res for seed in seeds]  # to_vox
-        seeds = [seed - 0.5 for seed in seeds]  # to_center
+        assert our_origin == Origin('center')
```
CHrlS98, line 229: since we hardcoded it the assertions are unnecessary

CHrlS98 (commented):
Two comments but otherwise looks good to me!

scil-admin:

scripts/scil_compute_seed_density_map.py
```diff
@@ -22,7 +22,9 @@ def _build_arg_parser():
                                 formatter_class=argparse.RawTextHelpFormatter)
 
     p.add_argument('tractogram_filename',
-                   help='Tracts filename. Format must be .trk.')
+                   help='Tracts filename. Format must be .trk. \nFile should '
+                        'contain a "seeds" value in the data_per_streamline.\n'
+                        'These seeds must be in voxel world, center origin.')
```
frheault, line 27: I would write it as `space: voxel, origin: center`

scripts/scil_compute_seed_density_map.py
frheault: I don't know why they are there, they do not modify anything related to seed. They should be removed.

frheault (approved):
I think the next PR ( #597 ) should have a before and after figure showing the change of both 593 & 597 do not change anything major.

I think example data with a command line example would be useful to showcase both PR did not impact results in a meaningful way

scilpy/tracking/propagator.py
EmmaRenauld: Ok, thanks

EmmaRenauld:
> I don't know why they are there, they do not modify anything related to seed. They should be removed.

@frheault Yes, that's what we thought! Thanks!

scil-admin:

scil-admin:

CHrlS98 (approved):

EmmaRenauld:
As suggested by @frheault, I tried to reproduce results in master branch vs this branch.

Needs some tweaking (random rng seed in the tracker uses the seeding position, so it varies if we use a different space so I gave it a fixed value + adding sft.to_vox, to_center in the end to compare streamlines together in the same space), but running the test (`pytest scripts/tests/test_compute_local_tracking_dev.py` **see note) and printing stuff along the way gives exactly the same results in both cases :) So I suggest this PR is ready for approval.


IMPORTANT NOTE: For this to be true, the `compress 0.1` argument must be removed in the test. Without compression, the streamline is indeed the same in both cases. COMPRESSION SEEMS TO BE VERY DIFFERENT DEPENDING ON SPACE!!! IS THAT NORMAL?? @arnaudbore ?

Initial streamline is of length 244. 
Compression in voxmm space = 83 points.
Compression in vox space = 46 points.

Should we be worried?

scil-admin:

scil-admin:

arnaudbore:
> As suggested by @frheault, I succeeded to reproduce results in master branch vs this branch.
> 
> Needs some tweaking of the rng seed (the way we made it, the random rng seed depends on space), but running the test and printing stuff gives exactly the same results in both cases :) So I suggest this PR is ready for approval.
> 
> ** CAREFUL! The space changes the compression! In dipy's compression method, they say that the compression threshold value is in mm. So I suppose that the streamlines should be in mm? But even in the "real" scil_compute_local_tracking, streamlines are not converted in mm before using the compression. Is that an error? @arnaudbore

You are right, there is something fishy here. I think we never had this issue before because personnaly I've never used it and most of the time voxels are 1mm3 so at the end it does not make any difference. Can you fix this in this PR please ?

EmmaRenauld:
Done. Can you check the last commit, @arnaudbore ? In both scripts, data was not a SFT yet so I could not use their to_voxmm method.

scil-admin:

scilpy/tracking/tracker.py
```diff
@@ -300,12 +300,20 @@ def _get_streamlines(self, chunk_id):
...
-                    streamlines.append(
-                        compress_streamlines(np.array(line, dtype='float32'),
-                                             self.compression_th))
-                else:
-                    streamlines.append((np.array(line, dtype='float32')))
+                    # Compressing. Threshold is in mm. Verifying space.
+                    if self.space == Space.VOX:
+                        # Equivalent of sft.to_voxmm:
+                        streamline *= self.seed_generator.voxres
+                        compress_streamlines(streamline, self.compression_th)
```
CHrlS98, line 310: You could just transform the compression threshold to voxel space by dividing by voxel resolution, no?

EmmaRenauld: I'm not sure how compression works internally...

CHrlS98: I think the working principle is if the shortest distance between some point and the segment connecting its two immediate neighbours is smaller than the threshold, it is removed from the compressed streamline. The space in which the compression is done should not matter, see docstring [here](https://github.com/dipy/dipy/blob/08b89939a886f0b4a38d3eebc66cb0d1aae7a4fc/dipy/tracking/streamlinespeed.pyx#L540).

EmmaRenauld: Ok. However, here the voxel resolution is a 3D vector. If the resolution is not iso, we cannot do what you suggest. So I think this should still be the best way.

CHrlS98: ouais ok, true for non-iso.

arnaudbore (approved):
Go !

PR #594, opened 2022-07-11
Removing redundant normalization

EmmaRenauld:
Just a very small change. Removing redundant normalization in the abstract class. Normalization is already managed by the child classes.


scil-admin:

PR #607, opened 2022-09-12
[WIP] Update numpy - warning bot

arnaudbore:


arnaudbore:
NOT A PRIORITY
==============
We would have to jump to python3.8 and update python-spams. @frheault I will need your help for python-spams.

PR #616, opened 2022-09-28
Add a function to verify compatibility

arnaudbore:
#615 
This may need a little bit more work to make it smooth :)

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -18,6 +18,7 @@ import six
 
 from scilpy.io.streamlines import load_tractogram_with_reference
 from scilpy.utils.bvec_bval_tools import DEFAULT_B0_THRESHOLD
+from scilpy.utils.filenames import split_name_with_nii
 
 eddy_options = ["mb", "mb_offs", "slspec", "mporder", "s2v_lambda", "field",
                 "field_mat", "flm", "slm", "fwhm", "niter", "s2v_niter",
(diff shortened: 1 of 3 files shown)


arnaudbore:

scilpy/io/utils.py
```diff
@@ -463,6 +464,35 @@ def verify_compatibility_with_reference_sft(ref_sft, files_to_verify,
                              .format(file))
 
 
+def verify_compatibility(parser, list_files):
```
EmmaRenauld, line 467: I think this title is a bit too similar to "is_header_compatible", it feels like it will do the same when looking only at the name.
verify_compatibility_multiple_files? I don't know.

EmmaRenauld: Also, this loop is probably done in many many other scripts! Could be re-used!

EmmaRenauld (commented):
Did not test but looks good!

arnaudbore:
> It would be nice to print the affine or dimensions to see the inequality or imcompatibility in the error message. @arnaudbore possible?

Once dipy logging will be fixed we will see why files are not compatible.

scilpy/io/utils.py
arnaudbore: I think I got everything cover now. I also added the pair comparison instead of using only one reference.

arnaudbore:

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -99,12 +99,13 @@ def main():
 
     # Load all metrics files.
     if args.metrics_dir:
-        verify_compatibility(parser, [args.in_mask] +
-                              os.listdir(args.metrics_dir))
+        is_header_compatible_multiple_files(parser, [args.in_mask] +
+                                            os.listdir(args.metrics_dir))
```
EmmaRenauld, line 103: This does not work. It sends the filenames without their paths.

EmmaRenauld (commented):
Tested. Appart from comment, the rest is working.

p.s. You could be checking also if their dimensions work. I had wrongly included a 4D file in my metrics, and I can only know it when it bugs in `/utils/metrics_tools.py` but it becomes difficult to know which file is faulty.

arnaudbore:

scripts/scil_compute_metrics_stats_in_ROI.py
```diff
@@ -98,9 +99,13 @@ def main():
 
     # Load all metrics files.
     if args.metrics_dir:
+        is_header_compatible_multiple_files(parser, [args.in_mask] +
+                                            list_metrics_files)
         metrics_files = [nib.load(args.metrics_dir + f)
                          for f in sorted(os.listdir(args.metrics_dir))]
```
EmmaRenauld, line 105: Tested. Your code works, but the line below, not.
`args.metrics_dir + f` should be os.path.join(args.metrics_dir, f)`.

If I use `--metrics_dir my_path` instead of `--metrics_dir my_path/'`, it fails.

Otherwise it works.

EmmaRenauld:
Shouldn't a 4D data and a 3D data be considered "compatible" if the resolution and so on fits? Right now it fails when dealing with 4D data.

scripts/scil_compute_metrics_stats_in_ROI.py
arnaudbore: Nice catch ! This was a bug in the current version.

arnaudbore:

arnaudbore:

PR #626, opened 2022-10-12
Gradient sampling ajustment

karanphil:
Very small change in the `scil_generate_gradient_sampling.py` script, in order to get a gradient table readable by a Philips scanner. Indeed, Philips scanners require that the gtab have unique pairs of bvecs and bvals for each line. By default, the script outputs the same pair of bvecs [0, 0, 0] and bvals [0] for each bvals in the gtab. The added option simply copies the bvecs from the previous line for each bvals, except the first one.

To test this, simply run the script, for example `scil_generate_gradient_sampling.py 16 32 64 dwi --eddy --duty --b0_every 18 --bvals 500 1000 2000 --mrtrix --b0_philips`, and make sure that no b0 has the same bvecs.

### DIFF
--- a/scilpy/gradientsampling/optimize_gradient_sampling.py
+++ b/scilpy/gradientsampling/optimize_gradient_sampling.py
@@ -49,7 +49,8 @@ def swap_sampling_eddy(points, shell_idx, verbose=1):
 
         # System energy matrix
         # TODO: test other energy functions such as electron repulsion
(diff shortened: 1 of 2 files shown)


arnaudbore:

arnaudbore:

EmmaRenauld (approved):
Tested with an without --b0_philips. Works, thanks!


p.s. For future work. Wouldn't we want to add a rng seed option to be able to produce twice the same result?

arnaudbore:
@EmmaRenauld can you add an issue about improving this with a rng seed option ?

PR #640, opened 2022-11-03
Project streamlines from/to wm/gm

frheault:
This new script expands on a previous version, projecting values from WM (streamlines) to the cortex (endpoints).
This version can project data from GM or WM to GM or WM (4 combos)

The data can come from metric maps (like before, FA, MD) or data_per_streamline or data_per_point (new).
The data_per_* can be from the .trk metadata or loaded from an array file.

This can be useful to project FA to the cortex, (WM->GM) or cortical thickness to the deep WM (GM->WM).
It can be used to visualize the distribution of streamlines and what they "touch" in a different way.

[test_for_kurt_project.zip](https://github.com/scilus/scilpy/files/9929743/test_for_kurt_project.zip)


scripts/scil_project_streamlines_to_map.py
```diff
@@ -0,0 +1,232 @@
...
+
+    sft = load_tractogram_with_reference(parser, args, args.in_bundle)
+    sft.to_vox()
+    sft.to_corner()
+
+    if len(sft.streamlines) == 0:
+        logging.warning('Empty bundle file {}. Skipping'.format(args.bundle))
+        return
+
+    mins, maxs, indices = _process_streamlines(sft.streamlines,
```
AntoineTheb, line 173: This and the following lines should be abstracted in a or multiple functions, which can reside in `scilpy`. The function(s) could return maps that can be saved in the main function. Overall, only IO should be in the main function.

scripts/scil_project_streamlines_to_map.py
```diff
@@ -0,0 +1,232 @@
...
+    p.add_argument('--from_wm', action='store_true',
+                   help='Project metrics from whole streamlines coverage.')
+    p.add_argument('--to_wm', action='store_true',
+                   help='Project metrics into streamlines coverage.')
+    add_reference_arg(p)
+    add_overwrite_arg(p)
+    return p
+
+
+def _compute_streamline_mean(cur_ind, cur_min, cur_max, data):
```
AntoineTheb, line 73: This probably has a place in the library

scripts/scil_project_streamlines_to_map.py
```diff
@@ -0,0 +1,232 @@
...
+                                                  cur_range)
+    streamline_data = data[cur_min[0]:cur_max[0],
+                           cur_min[1]:cur_max[1],
+                           cur_min[2]:cur_max[2]]
+    streamline_average = np.average(streamline_data,
+                                    weights=streamline_density)
+    return streamline_average
+
+
+def _process_streamlines(streamlines, just_endpoints):
```
AntoineTheb, line 87: This probably has a place in the library and/or use functions that are already in the library

scripts/scil_project_streamlines_to_map.py
```diff
@@ -0,0 +1,232 @@
...
+            s = np.stack((s[0, :], s[-1, :]), axis=0)
+        offset_streamlines.append((s - mins[-1]).astype(np.float32))
+
+    offset_streamlines = ArraySequence(offset_streamlines)
+    indices = uncompress(offset_streamlines)
+
+    return mins, maxs, indices
+
+
+def _project_metrics(curr_metric_map, count, orig_s, streamline_mean,
```
AntoineTheb, line 110: This could also go into the library. I would also add comments to tell the user that this function has side-effects instead of returns.

AntoineTheb (changes requested):
Following discussions that we had, the main should be cleaned up and functions should be moved into `scilpy` instead of mixing IO and logic.

arnaudbore:

PR #642, opened 2022-11-03
Fix rare division by 0 in f1

EmmaRenauld:
I came over a weird case, with a bundle found with 0% overlap, 100% overreach.

The f1 fails with a division by 0. In that case, by definition, the f1 score (Dice score) = how good you are, should be 0.

image

### DIFF
--- a/scilpy/tractanalysis/scoring.py
+++ b/scilpy/tractanalysis/scoring.py
@@ -60,6 +60,11 @@ def compute_f1_score(overlap, overreach):
 
     Ref: https://en.wikipedia.org/wiki/F1_score
     """
+    # In the case where overlap = 0 (found the bundle but entirely out of the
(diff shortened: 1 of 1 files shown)


arnaudbore:

AntoineTheb:
Im puzzled by how that's possible. In your screenshot, the GT bundle is in red and the streamline in blue ? How did the streamline get associated with this bundle ? Due to dilation ?

EmmaRenauld:
> I'm puzzled by how that's possible. In your screenshot, I presume the GT bundle is in red and the streamline in blue ? How did the streamline get associated with this bundle ? Due to dilation ?

Yes, exactly, dilation.  (Of the endpoints. Dilation of the GT itself to create the "all_mask" is normal; else we would never have any overreach). Dilation of the endpoints is normal too, to get the fanning that somebody would create, for instance.

AntoineTheb (approved):

PR #643, opened 2022-11-03
Fix: names for overreach

EmmaRenauld:
Oups, sorry....
I seem to have inverted the two versions of the Overreach scores in the final json (computation was ok). Fixed and clarified.

### DIFF
--- a/scilpy/tractanalysis/scoring.py
+++ b/scilpy/tractanalysis/scoring.py
@@ -25,10 +25,11 @@ Fidelity metrics:
     - OL : percentage of ground truth voxels containing VS streamline(s).
     - OR/ORn: percentage of voxels containing VS streamline(s) when it
         shouldn't. We compute two versions of the overreach:
-        OR = % of the recovered bundle. Values range between 0 and 100%. Values
-           are not defined when we recovered no streamline for a bundle, but we
-           set the OR to 0 in that case.
-        ORn = % of the ground truth bundle. Values could be higher than 100%.
+        OR_pct_vs = % of the recovered bundle. Values range between 0 and 100%.
+           Values are not defined when we recovered no streamline for a bundle,
+           but we set the OR to 0 in that case.
+        OR_pct_gt = % of the ground truth bundle. Values could be higher than
+           100%.
     - f1 score (which is the same as the Dice score).
 """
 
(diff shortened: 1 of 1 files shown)


scilpy/tractanalysis/scoring.py
```diff
@@ -25,10 +25,11 @@
     - OL : percentage of ground truth voxels containing VS streamline(s).
     - OR/ORn: percentage of voxels containing VS streamline(s) when it
```
AntoineTheb, line 26: Change this line to `OR_vs/OR_gt: percentage of voxels containing VS streamline(s) when it` ?

scilpy/tractanalysis/scoring.py
```diff
@@ -25,10 +25,11 @@
...
         shouldn't. We compute two versions of the overreach:
-        OR = % of the recovered bundle. Values range between 0 and 100%. Values
-           are not defined when we recovered no streamline for a bundle, but we
-           set the OR to 0 in that case.
-        ORn = % of the ground truth bundle. Values could be higher than 100%.
+        OR_pct_vs = % of the recovered bundle. Values range between 0 and 100%.
```
AntoineTheb, line 28: Is overreach ever not a percentage ? If so, maybe the `pct` suffix is redundant ?

scilpy/tractanalysis/scoring.py
```diff
@@ -102,11 +103,11 @@ def compute_f1_overlap_overreach(current_vb_voxels, gt_mask, dimensions):
         corresponds to the FN count (false negative).
     overlap: float
         TP divided by the ground truth count (i.e. TP + FN), in percentage.
+    overreach_pct_gt: float
+        The overreach, normalized by the ground truth area.
     overreach_pct_total: float
```
AntoineTheb, line 108: This metric is still referred as `overreach_pct_vs` everywhere else I think ?

AntoineTheb (commented):
Some comments but I'm not 100% confident about them

arnaudbore:

scilpy/tractanalysis/scoring.py
EmmaRenauld: Yes we do, the count of overreach voxels

PR #647, opened 2022-11-10
Fix logging across scilpy

frheault:
Thanks to Laurent I realized that `scil_search_keywords.py` stopped working.
It was due to the fact that setting logging to `INFO` or `DEBUG` the way we did (everywhere in Scilpy) does not work anymore.

So it also meant that around 70% of our logging was not working properly (WARNING is fine, because it was the default for the `root::logging`).
So I fixed it everywhere.

I also added a `pytest.ini` to print warnings from other libraries **only once** and keep all warnings from Scilpy (**even if many**) reducing printed line from 1M to 

PS: A few pep8 changes got in there while saving files, a few scripts didn't have permissions like everything else (777) so I changed it for uniformity.


arnaudbore:

arnaudbore:

arnaudbore:

frheault:
So the total output of tests on Travis is now less than 1000 (including the checkout, build, tests and warning)
`= 368 passed, 76 warnings in 2029.82s (0:33:49) =`

arnaudbore:

scripts/scil_run_commit.py
```diff
@@ -80,8 +79,7 @@
 import numpy as np
 import nibabel as nib
 
-from scilpy.io.streamlines import (lazy_streamlines_count,
```
AntoineTheb, line 83: Why is `lazy_streamlines_count` no longer imported ? Was it unused before ?

AntoineTheb (approved):
One comment but otherwise LGTM !

EmmaRenauld (approved):
I tested the pytest locally and, man!! Nice! So much easier! Takes a lot less time, too.

scripts/scil_run_commit.py
frheault: I am setting up my new computer and my autopep8 settings had too much power I think.

PR #648, opened 2022-11-11
Creating module 'labels'

EmmaRenauld:
One step into the cleaning of the modules. It seems an easy one to start with, not too much chances to get mixed up with something else.

I placed it in `scilpy.image.labels.py`.

I cleaned the 5 labels scripts that I found. Their main now only do imports and saving.
- scil_combine_labels
- scil_dilate_labels
- scil_remove_labels
- scil_split_volume_by_ids
- scil_split_volume_by_labels

I moved `scil.io.image.get_data_as_label` into scil.image.labels, explaining the other modifications. It could also be kept in io as it does fetch the data from the img.

### DIFF
--- a/scilpy/image/labels.py
+++ b/scilpy/image/labels.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+import os
+
+import numpy as np
+
+
+def get_data_as_labels(in_img):
+    """
+    Get data as label (force type np.uint16), check data type before casting.
+
+    Parameters
+    ----------
+    in_img: nibabel.nifti1.Nifti1Image
+        Image.
+
+    Return
+    ------
+    data: numpy.ndarray
+        Data (dtype: np.uint16).
+    """
+    curr_type = in_img.get_data_dtype()
+
+    if np.issubdtype(curr_type, np.signedinteger) or \
+       np.issubdtype(curr_type, np.unsignedinteger):
+        return np.asanyarray(in_img.dataobj).astype(np.uint16)
+    else:
+        basename = os.path.basename(in_img.get_filename())
+        raise IOError('The image {} cannot be loaded as label because '
+                      'its format {} is not compatible with a label '
+                      'image'.format(basename, curr_type))
--- a/scilpy/image/labels.py
+++ b/scilpy/image/labels.py
@@ -3,6 +3,7 @@ import logging
 import os
 
 import numpy as np
+from scipy.spatial.ckdtree import cKDTree
 
 
 def get_data_as_labels(in_img):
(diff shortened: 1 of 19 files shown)


arnaudbore:

frheault:
Great ! When 1.4 is released, and this PR is updated with master (just in case) I will test to check if the data before/after is the same.

I think all these functions could be tested with a fake 4x4 array, when I test it I will try to generate some data for unit tests.

scilpy/image/labels.py
```diff
@@ -0,0 +1,281 @@
+# -*- coding: utf-8 -*-
+import inspect
+import logging
+import os
+
+import numpy as np
+from scipy.spatial.ckdtree import cKDTree
+
+import scilpy  # ToDo. Is this the only way?
+
```
EmmaRenauld, line 10: To put line 48

scilpy/image/labels.py
```diff
@@ -0,0 +1,281 @@
...
+def get_lut_dir():
+    """
+    Return LUT directory in scilpy repository
+
+    Returns
+    -------
+    lut_dir: string
+        LUT path
+    """
+    # Get the valid LUT choices.
```
EmmaRenauld, line 47: Not the best code. Will add an issue for eventually improving.

arnaudbore:

scripts/scil_split_volume_by_ids.py
```diff
@@ -89,31 +63,25 @@ def main():
...
-    cnt_filename = 0
-    for label in label_indices:
-        if int(label) != 0:
-            split_label = np.zeros(label_img.shape,
-                                   dtype=np.uint16)
-            split_label[np.where(label_img_data == int(label))] = label
+    split_data = split_labels(label_img_data, label_indices)
 
-            split_image = nib.Nifti1Image(split_label,
+    for i in range(len(label_indices)):
```
frheault, line 79: This list is always out of range singe label_indices include the 0 index, but the filenames excludes it.

Right now labels 2 is saved as PREFIX_1.nii.gz (shifted)

frheault (approved):
I think this works well (except for one thing), if there is another correction I will do it in the next PR

(Immediately after this one I plan on doing unit tests for all new functions (replacing `main`)

arnaudbore:

scripts/scil_split_volume_by_ids.py
EmmaRenauld: I'm not sure that this was the issue. I think I fixed it nonetheless. Can you verify?

arnaudbore:

scripts/scil_split_volume_by_ids.py
EmmaRenauld: Fixed.

arnaudbore:

PR #672, opened 2023-01-31
Fixing uniform_coloring in scil_visualize_bunldes.py

karanphil:
Just a quick fix of the `uniform_coloring` option in `scil_visualize_bunldes.py`. The color was given in RGB (in range [0, 255]) but the viewer used range [0, 1], so I kept the 0 to 255 as input and divided by 255 before feeding it to the viewer. @AntoineTheb


arnaudbore:

AntoineTheb (approved):
Thanks for the fix @karanphil, LGTM

frheault:
@karanphil Can you update this code with master and then check if everything runs normally.
The master is now up to 3.10 (but should run in 3.8/3.9), so you may have to create a new venv.

arnaudbore:

karanphil:
> 

Works fine with python 3.10!

frheault (approved):

PR #683, opened 2023-03-02
skip __init__ in search_keyword.

frheault:
No matter the permission the `__init__.py` was giving that error when calling `scil_search_keywords.py `with the `--search_parser` option

`PermissionError: [Errno 13] Permission denied: '/home/rhef1902/Libraries/scilpy/scripts/__init__.py'`

The second change is actually a commit mistake, but it simply supports `.mat` files from ANTs for both single and double precision matrices.


arnaudbore:

PR #684, opened 2023-03-02
Reactivate md5sum in fetcher

frheault:
Switch from logging to raise ValueError


arnaudbore:

PR #685, opened 2023-03-02
uint8 in compute_density when possible

frheault:
Uses uint8 if args.binary ==  1

This is the default in scilpy and MI-Brain for binary mask (0/1)


arnaudbore:

PR #688, opened 2023-03-08
Adding small_dims option to scil_compute_fodf_max_in_ventricles

karanphil:
Small PR to allow the use of the `scil_compute_fodf_max_in_ventricles.py` script with 2D data or small dimensions 3D data. Just make sure it doesn't break other stuff that uses it (it should since the code is exactly the same when small_dims=False).


arnaudbore:

frheault (approved):

PR #692, opened 2023-03-10
Fix default sphere tracking

arnaudbore:


arnaudbore:

scripts/scil_compute_local_tracking.py
```diff
@@ -61,7 +61,7 @@ def _build_arg_parser():
     track_g = add_tracking_options(p)
     track_g.add_argument('--algo', default='prob',
                          choices=['det', 'prob', 'eudx'],
-                         help='Algorithm to use [%(default)s]')
+                         help='Default: [%(default)s]')
```
CHrlS98, line 64: Why is there no more description for the `--algo` argument?

arnaudbore: The only one I could find was: `Algorithm to use (must be "det" or "prob" or "eudx")`
Not very informative.

PR #704, opened 2023-04-03
Adding legacy options to scil_convert_sh_basis.py

karanphil:
Hi, this is a small addition to the `scil_convert_sh_basis.py` script to allow for the choice of not legacy versions of descoteaux07 and tournier07 SH bases. Previously, both bases were assumed to be legacy, which is usually the case for Dipy (descoteaux07). However, the current Mrtrix3 tools use a new version of tournier07, which is not the "legacy" one. So when converting from tournier07 with new data to descoteaux07, I got some weird little blobs at the center of my fODFs. Now, we can specify if we don't want the legacy version for the output, and if the input was coded with legacy or not.

I was wondering if I should check for the input/output selected bases and choose not legacy by default for tournier07. What do you think @arnaudbore @CHrlS98 ?


arnaudbore:

CHrlS98 (commented):
It looks good, but I just wonder if we shouldn't change the name of the SH bases such that we have `descoteaux07`, `descoteaux07_legacy`, `tournier07` and `tournier07_legacy`? And specify input SH basis and output SH basis instead of only input.

karanphil:
> 

Good point. Either ways look fine to me. I will let @arnaudbore choose ;)

CHrlS98:
We can start this way and then if we want to support legacy vs non-legacy everywhere we can think of a way to integrate it in the `add_sh_basis_args()` method.

scripts/scil_convert_sh_basis.py
```diff
@@ -25,9 +26,15 @@ def _build_arg_parser():
 
     p.add_argument('in_sh',
                    help='Input SH filename. (nii or nii.gz)')
-
     p.add_argument('out_sh',
                    help='Output SH filename. (nii or nii.gz)')
+    
+    p.add_argument('--is_input_not_legacy', action='store_true',
```
arnaudbore, line 32: Definitively a detail but I would like better this way
```
is_input_not_legacy -> in_sh_is_legacy (since we know it is (or not) and I'm not found of double negative)
is_output_not_legacy -> out_sh_legacy
```

scripts/scil_convert_sh_basis.py
```diff
@@ -42,13 +49,18 @@ def main():
     assert_inputs_exist(parser, args.in_sh)
     assert_outputs_exist(parser, args, args.out_sh)
 
+    print(args.is_input_not_legacy)
```
arnaudbore, line 52: Remove print and add logging to verbose some info of the sh basis/legacy/not legacy...

scilpy/reconst/multi_processes.py
```diff
@@ -446,6 +447,16 @@ def convert_sh_basis(shm_coeff, sphere, mask=None,
...
         Default: multiprocessing.cpu_count()
+    is_input_legacy: bool, optional
+        If true, this means that the input SH used a legacy basis definition
+        for backward compatibility with previous ``tournier07`` and
+        ``descoteaux07`` implementations.
+        Default: True
+    is_output_legacy: bool, optional
+        If true, this means that the output SH will use a legacy basis
+        definition for backward compatibility with previous ``tournier07`` and
+        ``descoteaux07`` implementations.
```
arnaudbore, line 458: If there is a link to mrtrix or dipy documentation about that. It would be useful here and/or in the script

arnaudbore (approved):
A couple of comments but apart from that LGTM.

scripts/scil_convert_sh_basis.py
karanphil: Not sure the verbose is necessary here... and if so, what would be the right way to do it?

arnaudbore:

arnaudbore:

arnaudbore:

CHrlS98 (approved):

PR #706, opened 2023-04-06
Unit test tractogram operations

EmmaRenauld:
[WIP] Need to check if other tests are needed.

Some fixes from testing:

- In the script scil_tractogram_math: to use a robust operation, you're supposed to send operation 'difference' and option --robust. But if you send operation 'difference_robust', it was accepted. You could end up with operation 'difference_robust_robust'. Modifying accepted operations in argparser.

- If you want to do, example, intersection of many big files: it was concatenating all files (doubling eveything in memory) and then keeping only the few accepted streamlines. Inversing: First filtering the sft, (freeing memory), then concatenating. Encapsulated that step out of the main code to be able to test it.

- Robust intersection seemed buggy to me.  Lines `if not union_mode or not difference_mode` are always true, so this was not correctly testing that we are doing intersection_mode. It should have been AND. Also changed a bit the code to make it clearer, but plase double-check me.

Unit tests passing locally, not sure why the deploy step doesn't work here.

### DIFF
--- a/scilpy/io/fetcher.py
+++ b/scilpy/io/fetcher.py
@@ -169,4 +169,5 @@ def fetch_data(files_dict, keys=None):
                                           "with a zip file.")
 
         else:
+            # toDo. Verify that data on disk is the right one.
             logging.warning("Not fetching data; already on disk.")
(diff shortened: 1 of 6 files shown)


arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-706/1/

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-706/2/

arnaudbore:

frheault (approved):

PR #709, opened 2023-04-20
Fix: Add bundle to scores if it was not found

EmmaRenauld:
When a (VB) bundle is not found: add its score to the average and to the final result.


arnaudbore:

AntoineTheb (approved):

PR #715, opened 2023-04-28
Adding absolute_peak_values to scil_compute_fodf_metrics.py

karanphil:
Hi, I added a `absolute_peak_values` option to `scil_compute_fodf_metrics.py`, because right now, we return the `peak_values` normalized per voxel and the `peaks` are also multiplied by these `peak_values`, so we kind of have the information in double. Doing so, we lose any idea of peaks scale between the voxels, since the first peak of each voxel has a `peak_value` of 1.

What I propose is to save the actual peak amplitudes as `peak_values` and to save the `peaks` (peak_dirs) as unit vectors. This way, we keep a way to compare between voxels, and we can still easily apply the `peak_values` per voxel if we want afterwards.

In order to not break any existing pipeline, I put it as an option and kept the legacy way as default.

@CHrlS98  @mdesco @frheault @arnaudbore , I think this concerns you.

### DIFF
--- a/scripts/scil_compute_fodf_metrics.py
+++ b/scripts/scil_compute_fodf_metrics.py
@@ -64,6 +64,11 @@ def _build_arg_parser():
     p.add_argument('--rt', dest='r_threshold', type=float, default='0.1',
                    help='Relative threshold on fODF amplitude in percentage  '
                         '[%(default)s].')
+    p.add_argument('--absolute_peak_values', action='store_true',
+                   help='If set, the peak_values are not normalized for each '
+                        'voxel, \nbut rather they keep the actual fODF '
+                        'amplitude of the peaks. \nAs a result, the peaks are '
+                        'also all normalized to 1. [%(default)s]')
     add_sh_basis_args(p)
     add_overwrite_arg(p)
     add_processes_arg(p)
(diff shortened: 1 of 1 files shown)


arnaudbore:

scripts/scil_compute_fodf_metrics.py
```diff
@@ -64,6 +64,11 @@ def _build_arg_parser():
     p.add_argument('--rt', dest='r_threshold', type=float, default='0.1',
                    help='Relative threshold on fODF amplitude in percentage  '
                         '[%(default)s].')
+    p.add_argument('--absolute_peak_values', action='store_true',
+                   help='If set, the peak_values are not normalized for each '
```
CHrlS98, line 68: max-normalized? I'd add explicitly.

scripts/scil_compute_fodf_metrics.py
```diff
@@ -64,6 +64,11 @@ def _build_arg_parser():
...
+                        'voxel, \nbut rather they keep the actual fODF '
+                        'amplitude of the peaks. \nAs a result, the peaks are '
```
CHrlS98, line 70: I bugged a bit on the last line. "Also, peaks are given as unit directions instead of being proportional to peak_values." or something?

scripts/scil_compute_fodf_metrics.py
```diff
@@ -64,6 +64,11 @@ def _build_arg_parser():
     p.add_argument('--rt', dest='r_threshold', type=float, default='0.1',
                    help='Relative threshold on fODF amplitude in percentage  '
                         '[%(default)s].')
+    p.add_argument('--absolute_peak_values', action='store_true',
```
CHrlS98, line 67: `--abs_peaks_and_values` maybe?

CHrlS98 (commented):
Should we also add a note in the script description?

CHrlS98 (approved):
Nice

mdesco (approved):

PR #718, opened 2023-05-10
Add angle coloring option + fix bugs

EmmaRenauld:
Added an option to color a tractogram based on the local angle of the streamline. Particularly nice on smoothed streamlines.

\+ Fixed bugs in script, where options min_range, max_range, etc. were not used with option --along_profile (and thus lbound, ubound variables did not always exist).

\+ Added more ticks on the colorbar (previously only the min and max values were shown).

\+ Added the possibility to show the colorbar right away. It is the last step of the script so even if user closes the window, everything is done and saved.

image
image

### DIFF
--- a/scilpy/utils/streamlines.py
+++ b/scilpy/utils/streamlines.py
@@ -9,6 +9,7 @@ import numpy as np
 from scipy.spatial import cKDTree
 from sklearn.cluster import KMeans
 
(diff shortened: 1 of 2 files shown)


arnaudbore:

arnaudbore (approved):
Really nice

PR #719, opened 2023-05-15
[ENH] Add soft distance for each filter 

arnaudbore:


arnaudbore:

frheault:
I would prefer if this was a parameter in `filter_grid_roi()` (or same for bdo) and keeping the parameter --soft_distance, this way it would work for drawn roi, slices, and atlas the same way.

It would be possible to use --soft_distance_roi and --soft_distance_bdo, since bdo is in mm (I think)

arnaudbore:
I would like to be able to add distance specifically to each filter we have and not something overall specific to a type.
We have a couple of examples where you only want to have this distance applied to the a specific ROI and not all of them.
I will make it easier to read but do you think we should also have --soft_distance_roi and --soft_distance_bdo to be able to apply an overall distance ?

frheault:
Ok what about:
We keep --soft_distance_roi and --soft_distance_bdo for --drawn_roi and --bdo
And we add a dilation last for --atlas_roi, if it finishes with include or exclude, it means no dilation and if it finishes with a number that number is the number of passes.

the nargs could be '+' with a check on the number accepted 
--atlas_roi ROI_NAME ID MODE CRITERIA (DIST)
4 or 5 (and if 5, ends with integer)

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-719/2/

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-719/3/

arnaudbore:

PR #721, opened 2023-05-17
Correlation ops

frheault:
New operations for scil_image_math.py to do patch-wise correlation (3,3,3) between two images.

Useful between density maps (different algo, test-retest, etc)


arnaudbore:

PR #723, opened 2023-05-31
[ENH] Add new parameter to overwrite filter distance

arnaudbore:


frheault:
I don't see the use for this if we just added the distance (opt) parameter.

This script is already a monster of complexity, so why not just use that new behavior you just added?

arnaudbore:
> I don't see the use for this if we just added the distance (opt) parameter.
> 
> This script is already a monster of complexity, so why not just use that new behavior you just added?

I understand but this parameter will allow you to not take care of the distance individually and just force any either_end both_end to a specific distance. It is specially interesting with a lot of filtering lists you don't want to modify each time you run a test.

frheault:
Compromise: why not having an argument `--overwrite_distance 'mode' 'distance'` (nargs='+') that can overwrite the type of filtering of your choice (and not just _ends_)

With check that you don't accidentally write `--overwrite_distance any 1 --overwrite_distance either_end 2 --overwrite_distance any 3` (writing the same mode twice)

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-723/4/

frheault (approved):
I tested it with 3 BDO and 1 NIFTI to segment an AF and it worked well!

PR #728, opened 2023-06-22
Multi centro labels map

frheault:
Update and simplification of the compute bundle voxel labels map.

- Removal of the 2 parameters that no one uses (to trim streamlines) and that were a headache.
- Switch from memmap to compute distances to a much faster/smarter kdtree 
- Fix a corner case of scil_project_streamlines.py 

Now scil_compute_bundle_voxel_label_map.py support multi-centroids input to better handle complex fanning, but see be somewhat reliable. However, the full complexity of centroids' design is still an open question. As of now, it is still only an option and the default remains the same.

Tested with the last commit of https://github.com/scilus/tractometry_flow/pull/18


arnaudbore:

PR #729, opened 2023-07-02
BUG: Allow tractogram math script to run as such

jhlegarreta:
Allow tractogram math script to run as such: without the
```
if __name__ == "__main__":
    main()
```

block, the script does simply not call `main` and does not perform any computation.

### DIFF
--- a/scripts/scil_tractogram_math.py
+++ b/scripts/scil_tractogram_math.py
@@ -187,3 +187,7 @@ def main():
                                                        args.out_tractogram))
     save_tractogram(new_sft[indices], args.out_tractogram,
                     bbox_valid_check=args.bbox_check)
+
+
+if __name__ == "__main__":
+    main()


arnaudbore:

arnaudbore:
Quick question @jhlegarreta : I understand the issue but I don't get any error. Is this a python error ? Which version of python are you using ?

jhlegarreta:
> I understand the issue but I don't get any error. Is this a python error ? 

I marked it as a BUG because on my end the script does nothing as the `main` method is not being called. No error or exception is thrown, tests are passing but no actual computation is being made. If you remove the block and put random input arguments, no error will be thrown.

> Which version of python are you using ?

Python 3.10.6.

arnaudbore:
@jhlegarreta even tho this an error, it should work since we have:
```
            entry_points={
                'console_scripts': ["{}=scripts.{}:main".format(
                    os.path.basename(s),
                    os.path.basename(s).split(".")[0]) for s in SCRIPTS]
            },
```
How do you install scilpy ? Can you try a fresh install with a new environment ?

jhlegarreta:
> How do you install scilpy ?

Followed the instructions; `pip install -e .`. I'm calling the script from my IDE.

arnaudbore:
I'm going to merge your PR but the issue comes from the fact you are calling the script from your IDE.

jhlegarreta:
>  issue comes from the fact you are calling the script from your IDE.

That should not be an issue Arnaud.

PR #732, opened 2023-08-04
Correct ihMTsat

karanphil:
Here is a small modification to the computation of ihMTsat. Previously, it was in fact the ihMTdR1sat that was calculated. To output this metric, use the `--legacy_sat` option. If not, the real ihMTsat will be computed.

@Manonedde

### DIFF
--- a/scripts/scil_compute_ihMT_maps.py
+++ b/scripts/scil_compute_ihMT_maps.py
@@ -111,7 +111,8 @@ def _build_arg_parser():
     p.add_argument('--single_echo', action='store_true',
                    help='Use this option when there is only one echo.')
     p.add_argument('--legacy_sat', action='store_true',
-                   help='Use this option to choose the ihMTdR1sat contrast.')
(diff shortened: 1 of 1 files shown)


arnaudbore:

scripts/scil_compute_ihMT_maps.py
```diff
@@ -110,6 +110,8 @@ def _build_arg_parser():
                         'Not recommended.')
     p.add_argument('--single_echo', action='store_true',
                    help='Use this option when there is only one echo.')
+    p.add_argument('--legacy_sat', action='store_true',
+                   help='Use this option to choose the ihMTdR1sat contrast.')
```
Manonedde, line 114: I will add a "Not recommended" in the help of the --legacy_sat option.

Manonedde (changes requested):

arnaudbore:

arnaudbore:

PR #744, opened 2023-09-05
Remove pairwise

frheault:
No need for pairwise, if all are identical to the first, then they all are identical.

(I was running a concatenate on 8000 files, so I was suspicious of the time it took.


arnaudbore:

arnaudbore:

PR #747, opened 2023-09-21
Volumes modules

EmmaRenauld:
Preparing a few other refactored modules. 

Goal: To merge this before the scil retreat so that people can do the unit tests.

**Summary:**
 - image.utils --> Stayed the same for now. They are quite 'utils' as we discussed.
 - utils.image --> now image.volume_operations (as discussed). Ex: flip volume, crop volume, apply a transform, etc. However that name already existed:
 - image.volume_operations --> now image.volume_math. They are all the math operations from our math script.
 - image.datasets --> now image.volume_space_management. To be cleaned another day but at least name is more explicit.

**Associated scripts**:
The following scripts are now clean (i.e. the main script is as empty as possible).
- scil_apply_transform_to_image
- scil_count_non_zero_voxels
- scil_resample_volume
- scil_reshape_to_reference
- scil_snr_in_roi
- scil_flip_volume
- scil_image_math
- scil_crop_volume


arnaudbore:

scilpy/image/volume_space_management.py
```diff
@@ -4,6 +4,41 @@
 from dipy.core.interpolation import trilinear_interpolate4d, \
     nearestneighbor_interpolate
 from dipy.io.stateful_tractogram import Origin, Space
+from dipy.segment.mask import bounding_box
+
+
+from scilpy.utils.util import voxel_to_world
+
+
+class WorldBoundingBox(object):
```
AlexVCaron, line 13: We need to discuss this more. This class is used as a pickle in the `scil_crop_volume.py` script. Moving it here won't work, it must be a top of module to be picklable (https://docs.python.org/3/library/pickle.html).

And even, it will bug all pickled WorldBoundingBox kept by scilpy user (a pickle is bound to the module in which it was created). As such, the script parameters `--input_bbox` won't work with old boxes.

If we intend to deprecate those boxes, we should deprecate this pickle usage instead and rely on another saving method for the bounding box. It is legacy and really the worst way of doing this job.

AlexVCaron: @arnaudbore @frheault @EmmaRenauld I would move this code back to the script, if we all agree. With a todo to change the way it works, remove pickling from the solution.

scilpy/tracking/propagator.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, dataset, step_size, rk_order,  space, origin):
         """
         Parameters
         ----------
-        dataset: scilpy.image.datasets.DataVolume
+        dataset: scilpy.image.volume_space_management.DataVolume
```
AlexVCaron, line 36: Replace dataset with volume ? And the objects Dataset to DataVolume ?

AlexVCaron: Here and on lines 258 and 324.

AlexVCaron (changes requested):
Small discussion to have on `WorldBoundingBox` and its usage in `scil_crop_volume.py`

scilpy/image/volume_space_management.py
EmmaRenauld: Ok. @frheault , I will remove scil_crop_volume from files to give to people at the retreat. Seems more complicated than expected.

scilpy/tracking/propagator.py
EmmaRenauld: I'm not sure I understand what you mean. The class DataVolume is in scilpy.image.volume_space_management.DataVolume so this seems correct to me.

scilpy/image/volume_space_management.py
EmmaRenauld: @AlexVCaron Tests were passing, though. Do you have an example of pickle file that we could add to the tests?

AlexVCaron: We do not test the input and output pickle, so we don't check this behavior. I'll prepare a little package for here and the retreat.

arnaudbore:

scilpy/image/volume_space_management.py
AlexVCaron: [Here is some data](https://usherbrooke-my.sharepoint.com/:u:/g/personal/vala2004_usherbrooke_ca/EWlYflA-EUVOmWaE8CcBjSsBo5BI1xZq1X9gWN3d6Sn14Q?e=wEfvA2). You have the pickled bbox, created from the current master.

Loading the box does not work. However, contrary to what I thought, it is still possible to create and load boxes with the bbox class in the module. I still think we should change to another way than pickle, but we'll talk about it at the retreat.

scilpy/tracking/propagator.py
AlexVCaron: I was talking about the parameters names ;) the full class name is good

scilpy/image/volume_space_management.py
EmmaRenauld: Thanks. I will try to play with this eventually.

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-747/3/

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-747/4/

EmmaRenauld:
(Rebased on master. Locally tests are passing).

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-747/5/

arnaudbore:

EmmaRenauld:
@frheault @AlexVCaron @AntoineTheb This is ready!

AlexVCaron (approved):
I have not tested, but everything looks good to me ! If you want to, LGTM.

arnaudbore (approved):

PR #750, opened 2023-09-27
ENH: Add Parallel Transport Tractography to `scil_compute_local_tracking`

AntoineTheb:
Add the option to chose Dipy's implementation of PTT as a direction getter instead of prob or det. Fixed some pep8 warnings also.

PTT has some additional parameters concerning its "probe" mechanism that would be fixed to the default values. Thoughts on this @gabknight ? Should these parameters be available through `scilpy` ?


arnaudbore:

AntoineTheb:
@gabknight I'm having some problems finding good values for the `probe_length` and `probe_radius` parameters of `PTTDirectionGetter`. Trekker's documentation mentions that `probe_length` should not exceed `minRadiusOfCurvature`. However, that parameter does not seem to be available through the direction getter. Is it `probe_radius` ? 

Using default parameters, with `theta`=15 on ISMRM2015:
image

AntoineTheb:
Trekker's documentation also mentions that a step size of 0.05mm or lower should be prioritized. However, this is the result I get (without modifying other parameters than `theta`) when doing so. 

image

AntoineTheb:
Setting `probe_length == step_size` in 56740cc seems to give much better results.

image

arnaudbore:

frheault:
Isn't 0.05mm like super slow to track? And the uncompressed file must be 10Go for --npv 1?

Does it give a nice result at 0.2 - 0.5mm? (closer to our default)?
If it only works well at 0.05 it would have to be very clear in the docstring.

gabknight:
@AntoineTheb `probe_length` and `probe_radius` define the shape of the probe. --> https://www.youtube.com/watch?v=5IUFmNIRCpI&ab_channel=DiffusionImaginginPython-DIPY

Dipy is working the `max_theta` (i.e. max angular deviation) instead of the `maxRadiusOfCurvature`. We plan on changing this in future redesign (https://github.com/dipy/dipy/pull/2575).

There are utility function if you wish to know the corresponding value of `theta` for a given `step_size` and `maxRadiusOfCurvature` here :
- `dipy.tracking.utils.max_angle_from_curvature(min_radius_curvature, step_size)`
- `dipy.tracking.utils.min_radius_curvature_from_angle(max_angle, step_size)`

There is no real point on tracking at 0.05mm I believe, especially that PTT is already 10x slower than dipy prob. We are working on speeding things up, but for now it is just a first implementation. The issue you likely have at 0.05 is that you didn't adjust the `max_theta` parameter, thus allowing streamlines to curve way more than when `step_size = 0.2`

I'm not sure why results are so nice when setting `prob_length` to `step_size`, but this way I guess you force the smoothness in trajectories from the Freenet frame model behind PTT. However, having `prob_length == step_size` make little neighboring voxels exploration, and thus more akin to dipy prob.

gabknight:
@AntoineTheb for the parameters. I wouldn't  change `prob_radius=0` `prob_count=1` and `prob_quality=3`, it is just too slow for now. `prob_length` is more interesting to play with, especially when tracking on phantoms, animal or human data. Having control on this one could be interesting. `data_support_exponent=1` is also fine I believe, and should be changed only when using specific data or reconstructing targeted pathways.

AntoineTheb:
@frheault It is indeed very slow at 0.05mm (couple of hours for interface seeding at npv=5). 

@gabknight thanks for the input ! If I shouldn't make equal to `step`, should I make `probe_length` a parameter of `scil_compute_local_tracking` then ? And suggest setting a `probe_length` similar to the step-size ? I'll keep other parameters to their default values.

gabknight:
@AntoineTheb I think so. in Trekker `prob_length` is 10x the step size, but the step_size is tiny (1/40 the voxel size). With a step_size of 0.2mm, I would try setting the `prob_length` to ~ the voxel size, with a `maxtheta` between 10 and 20 degrees. This should produce smoother trajectories than dipy prob with default params (0.2mm and 20deg).

https://dmritrekker.github.io/manual/trekker.html

arnaudbore:

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-750/6/

arnaudbore:
Build Failed :boom: 
https://jenkins.scil.usherbrooke.ca/job/scilpy/job/PR-750/7/

arnaudbore:

mdesco (commented):
We need to translate your discussion into the help. I'm currenty testing and I'm getting garbage results. I tried the default step size first -> garbage, lowered it to 0.1 -> garbage... 

I'm working on a macaque brain at 0.5 iso mm... 

Any way that, a bit like the "theta" parameter, the step_size parameter is set more intelligently for PTT with respect to voxel_size of the fodf field???

AntoineTheb:
> @AntoineTheb I think so. in Trekker `prob_length` is 10x the step size, but the step_size is tiny (1/40 the voxel size). With a step_size of 0.2mm, I would try setting the `prob_length` to ~ the voxel size, with a `maxtheta` between 10 and 20 degrees. This should produce smoother trajectories than dipy prob with default params (0.2mm and 20deg).
> 
> https://dmritrekker.github.io/manual/trekker.html

@gabknight  I just realized that I kept misreading your comment as setting the probe_length ~ the **step size**. I just pushed a fix to set to to the **voxel size**. 

@mdesco this should fix your problems. I guess we could set the step size according to the voxel size as used by Trekker, but I worry it would make the tracking too slow if we follow their recommendation. With the fix I just pushed, results should be satisfactory with a step size of 0.1/0.2mm.

arnaudbore:

scripts/scil_compute_local_tracking.py
```diff
@@ -12,6 +12,7 @@
 to the previous direction.
 Algo 'prob': a direction drawn from the empirical distribution function defined
 from the SF.
+Algo 'ptt': Parallel-Transport Tractography. See [1] for more details.
```
mdesco, line 15: Should we add a line that says that a smaller step size than usual is recommended with PTT. e.g. 1/5 of prob or det step size???

mdesco (approved):

scripts/scil_compute_local_tracking.py
AntoineTheb: Good point. Maybe not 1/5 of prob, but something along the lines of 0.1-0.2mm or below.

arnaudbore:

arnaudbore (approved):

PR #752, opened 2023-10-02
ENH:  `scil_tractogram_math` support for TCKs

AntoineTheb:
The current behavior of `scil_tractogram_math` (or rather, `is_header_compatible_multiple_files`) does not allow for tractogram operations on TCKs.  This fixes this behavior by comparing the TCK's reference or, in the case of two TCKs, bypassing the comparison as they will  use the same reference.

Fixed some pep8 warnings also. 

edit: Added `--reference` parameter to `scil_verify_space_attributes_compatibility.py` too. It can be considered not useful as the reference used for loading the validated tck could just be passed to the script, but I say why not.

### DIFF
--- a/scilpy/io/utils.py
+++ b/scilpy/io/utils.py
@@ -578,44 +578,47 @@ def is_header_compatible_multiple_files(parser, list_files,
     Verifies the compatibility between the first item in list_files
     and the remaining files in list.
 
+    Arguments
+    ---------
     parser: argument parser
         Will raise an error if a file is not compatible.
-
-    list_files: List
+    list_files: List[str]
         List of files to test
-
     verbose_all_compatible: bool
         If true will print a message when everything is okay
+    reference: str
+        Reference for any .tck passed in `list_files`
     """
     all_valid = True
 
+    # Gather "headers" for all files to compare against
+    # eachother later
+    headers = []
     for filepath in list_files:
         _, in_extension = split_name_with_nii(filepath)
-        if in_extension not in ['.tck', '.trk', '.nii', '.nii.gz']:
-            parser.error('{} does not have a supported extension'.format(
+        if in_extension in ['.trk', '.nii', '.nii.gz']:
+            headers.append(filepath)
+        elif in_extension == '.tck':
+            if reference:
+                headers.append(reference)
+            else:
+                parser.error(
+                    '{} must be provided with a reference.'.format(
+                        filepath))
+        else:
+            parser.error('{} does not have a supported extension.'.format(
                 filepath))
 
-    for curr in list_files[1:]:
-        compatible = True
-        if in_extension == '.tck':
-            # TCKs do not have a header, so compare with the reference's
-            # header.
-            _, curr_extension = split_name_with_nii(curr)
-            # If comparing two TCKs, the same reference will be used by
-            # both so they will be compatible.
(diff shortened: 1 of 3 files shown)


arnaudbore:

frheault:
Seems like it is working alright!

scilpy/io/utils.py
```diff
@@ -589,12 +591,23 @@ def is_header_compatible_multiple_files(parser, list_files,
...
     for curr in list_files[1:]:
-        if not is_header_compatible(list_files[0], curr):
+        compatible = True
+        if in_extension == '.tck':
+            # TCKs do not have a header, so compare with the reference's
+            # header.
+            _, curr_extension = split_name_with_nii(curr)
+            # If comparing two TCKs, the same reference will be used by
+            # both so they will be compatible.
+            if curr_extension != '.tck':
```
frheault, line 606: So since only trk has a header I would reverse that condition, if it is .trk check it?

scilpy/io/utils.py
```diff
@@ -589,12 +591,23 @@ def is_header_compatible_multiple_files(parser, list_files,
...
+                compatible = is_header_compatible(reference, curr)
+        else:
+            compatible = is_header_compatible(list_files[0], curr)
```
frheault, line 609: So here it assumes list_files[0] is a file that has an header to check against.

I think a better way of doing it would be that in the first for loop (`for filepath in list_files:`) you would have to find which of the input has a header (only `['.trk', '.nii', '.nii.gz'])` and remember which position it was and use that position instead of 0 in `list_files[0]`

The idea is that if you have:
- Only TCK: Do not check header since they dont have any
- A mix of TCK and TRK: Check the TRKs and the --reference so they match
- Only TRK: Check the TRKs so they match

So the thing missing in this part is that in_extension should now be used to detect if you have a least one TRK, and if you do have one its position should be remember for the check just after.

AntoineTheb: Sure !

In the first loop, I'll actually gather "headers" (be it nifti/trk headers or a tck's reference's header) and then I'll compare them as usual. Should be clean enough.

arnaudbore:

frheault:
I need to retest tomorrow, but I think your solution is cleaner than what I had in my head!

AntoineTheb:
retest this please

AntoineTheb:
add to whitelist

arnaudbore:
ghprb test

EmmaRenauld (approved):
Not tested but LGTM.

scripts/scil_verify_space_attributes_compatibility.py
```diff
@@ -20,7 +23,7 @@ def _build_arg_parser():
 
     p.add_argument('in_files', nargs='+',
                    help='List of file to compare (trk and nii).')
```
arnaudbore, line 25:
```suggestion
                   help='List of file to compare (trk, tck and nii/nii.gz).')
```

arnaudbore (approved):
LGTM

AntoineTheb:
A bug slipped in where the ref wasn't passed to `is_header_compatible_multiple_files`. It is fixed now. 

@frheault have you had time to test ?

arnaudbore:

arnaudbore:

PR #757, opened 2023-10-04
ENH: Adding option to skip randomization of seed. + other fixes

EmmaRenauld:
- Adding an option to the _dev tracking to have fixed seeds. This allows to create a figure like this:
     
step_size05

- Changing the --discard_last_point option to make it the default choice (as discussed with Gab and other; this makes it more like Dipy)

- Fixing the way the numpy generator and seed are used, as per the[ numpy doc](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html). If you think it's not the best way to do it, I can just delete the last commit.

### DIFF
--- a/scilpy/tracking/tests/notes.txt
+++ b/scilpy/tracking/tests/notes.txt
@@ -1,5 +1,0 @@
-
-
-We will not test the tracker / propagator : too big to be tested, and only used
-in scil_compute_local_tracking_dev, which is intented for developping and
-testing new parameters.
--- a/scilpy/tracking/tests/notes.txt
+++ b/scilpy/tracking/tests/notes.txt
@@ -1,0 +1,5 @@
+
+
+We will not test the tracker / propagator : too big to be tested, and only used
+in scil_compute_local_tracking_dev, which is intented for developping and
+testing new parameters.
--- a/scripts/scil_compute_local_tracking_dev.py
+++ b/scripts/scil_compute_local_tracking_dev.py
@@ -120,6 +120,11 @@ def _build_arg_parser():
              "direction \n(based on the propagator's definition of invalid; "
              "ex when \nangle is too sharp of sh_threshold not reached) are "
              "never added.")
+    track_g.add_argument(
+        "--do_not_randomize_seed_positions", action="store_true",
+        help="By default, seed position is moved randomly inside the voxel. "
+             "Use this option to have all seeds centered at the middle of the "
+             "voxel.")
 
     add_seeding_options(p)
 
@@ -194,8 +199,10 @@ def main():
                       'seeding mask.'.format(args.in_seed))
 
     seed_res = seed_img.header.get_zooms()[:3]
+    randomize_positions = not args.do_not_randomize_seed_positions
     seed_generator = SeedGenerator(seed_data, seed_res,
-                                   space=our_space, origin=our_origin)
+                                   space=our_space, origin=our_origin,
+                                   randomize_positions=randomize_positions)
     if args.npv:
         # toDo. This will not really produce n seeds per voxel, only true
         #  in average.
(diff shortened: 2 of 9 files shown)


arnaudbore:

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -113,13 +113,20 @@ def _build_arg_parser():
...
-             "of \nthe streamline. Default: append them. This is the default "
+        '--keep_last_out_point', action='store_true',
+        help="If set, keep the last point (once out of the tracking mask) "
+             "of \nthe streamline. Default: discard them. This is the default "
              " in \nDipy too. Note that points obtained after an invalid "
              "direction \n(based on the propagator's definition of invalid; "
              "ex when \nangle is too sharp of sh_threshold not reached) are "
-             "never added.")
+             "never added.\n"
+             "REMARK: Our results (our endpoints) seem to differ from dipy's. "
```
CHrlS98, line 123: I think I would prefer not to write `todos` in the script help, but it's just a matter of personal taste.

CHrlS98 (commented):
I tested and it seems to work properly.

However, I think this would be more useful if, instead of disabling jittering the seeding positions, we added a `--nb_repeats` arguments telling how many times we want to track from each seed. The behaviour you want to show could still happen, but it would generalize better and might find other applications than only for teaching purposes?

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -113,13 +113,20 @@ def _build_arg_parser():
...
+             "This should be investigated.")
+    track_g.add_argument(
+        "--do_not_randomize_seed_positions", action="store_true",
```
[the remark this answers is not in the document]
EmmaRenauld: Changed for no_seed_displacement. Ok?

scripts/scil_compute_local_tracking_dev.py
```diff
@@ -113,13 +113,20 @@ def _build_arg_parser():
...
+        help="By default, seed position is moved randomly inside the voxel. "
```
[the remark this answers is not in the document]
EmmaRenauld: Changed.

arnaudbore:

CHrlS98:
> > However, I think this would be more useful if, instead of disabling jittering the seeding positions, we added a `--nb_repeats` arguments telling how many times we want to track from each seed. The behaviour you want to show could still happen, but it would generalize better and might find other applications than only for teaching purposes?
> 
> That is nice, but right now it seems complicated for me. We would have to remember how many times we called get_next_pos.

Wouldn't we just need to return the same seed `n_repeat` times ? This just changes the start position of a streamline, no?

arnaudbore:

EmmaRenauld:
Should be good!

scilpy/tracking/tests/notes.txt
CHrlS98, line 1: I would like to have @arnaudbore 's input on this file.

CHrlS98 (commented):
I would like to validate that the note.txt file in tests is at the right place, but otherwise, everything looks good to me.

scilpy/tracking/tests/notes.txt
arnaudbore: Why not using a regular test file, add this info inside as well as a dummy test that passes.

EmmaRenauld: I agree. It's just that I wasn't sure how to proceed. Do a create a test_tracker and test_propagator and copy the same text in both?

arnaudbore: I agree :+1:

EmmaRenauld: Done

arnaudbore:

CHrlS98 (approved):

PR #769, opened 2023-10-23
RF - changed tracking default sphere from symmetric724 to repulsion724

gabknight:
The `symmetric724` sphere is suboptimal for probabilistic tractography because the vertices density is higher at the poles. 

- Change default sphere to `repulsion724`. This change affects all scripts with input spheres.
- Change the default sphere for GPU tracking.

### DIFF
--- a/scripts/scil_compute_local_tracking.py
+++ b/scripts/scil_compute_local_tracking.py
@@ -371,7 +371,7 @@ def main():
         # data volume
         odf_sh = odf_sh_img.get_fdata(dtype=np.float32)
 
-        #GPU tracking needs the full sphere
+        # GPU tracking needs the full sphere
         sphere = get_sphere(args.sphere).subdivide(args.sub_sphere)
 
         streamlines_generator = GPUTacker(
--- a/scripts/scil_compute_local_tracking.py
+++ b/scripts/scil_compute_local_tracking.py
@@ -22,7 +22,7 @@ implementations:
     * Algo: For the GPU implementation, the only available algorithm is
         Algo 'prob'.
     * Tracking sphere: The only sphere available for GPU tracking is
-        `symmetric724` and `--sub_sphere` is not available for GPU tracking.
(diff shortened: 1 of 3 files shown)


CHrlS98 (changes requested):
For GPU tracking the get_sphere is called in `scilpy.tracking.tracker.GPUTracker._track()` https://github.com/scilus/scilpy/blob/44622e5bdc6ed7ab63b76ce81842da91c8c0f576/scilpy/tracking/tracker.py#L558. Please also change + we could make it an editable parameter while you're at it (if you wanna put the time, was on my todo anyway).

CHrlS98 (commented):
Also we use it as an hemisphere and not a whole sphere, see line 126
https://github.com/scilus/scilpy/blob/44622e5bdc6ed7ab63b76ce81842da91c8c0f576/scripts/scil_compute_local_tracking.py#L126

I don't know in that case if repulsion724 is still better than symmetric724?

gabknight:
ok thx @CHrlS98. I will check this in GPU tracker. Yes, the repulsion spheres in dipy play well with hemisphere.

arnaudbore:

CHrlS98 (approved):

arnaudbore:

arnaudbore:

scripts/scil_compute_local_tracking.py
```diff
@@ -22,7 +22,7 @@
     * Algo: For the GPU implementation, the only available algorithm is
         Algo 'prob'.
     * Tracking sphere: The only sphere available for GPU tracking is
-        `symmetric724` and `--sub_sphere` is not available for GPU tracking.
+        `repulsion724` and `--sub_sphere` is not available for GPU tracking.
```
arnaudbore, line 25: I don't understand, does sub_sphere works with GPU now or not ?

arnaudbore:

arnaudbore (approved):

PR #775, opened 2023-10-23
Move load img

karanphil:
I moved the load_img function from scil_image_math.py to io.utils.py.

@frheault @Manonedde @mdesco


arnaudbore (approved):
LGTM

PR #798, opened 2023-10-28
FIX: bug in scil_score_bundles.py

AntoineTheb:
There was a bug in scil_score_bundles where the invalid bundles were incorrectly loaded if --compute_ic was used in scil_score_tractogram.

The bug caused the same tractogram to be loaded for each invalid bundle, which itself was not necessarily an invalid bundle, messing up the results.


arnaudbore:

PR #808, opened 2023-11-13
[WIP] Add_noddi_ecvf

Manonedde:
Script to calculate the ECVF map from the ICVF map. 

An example of ICVF map is available here : 

/braindata/processedData/manon/data_PR_ECVF/icvf.nii.gz

@Arnaud Is there any data where I can find an ICVF map for test_ script?

@mdesco, @AlexVCaron can you review?


arnaudbore:

AlexVCaron:
@mdesco @arnaudbore I would be of the opinion this could be transformed and go into scil_image_math.py as a more general utility to flip a probability distribution (like a `scil_image_math.py invert_p` or something).

I would keep it as a script if we can add more metrics to be outputed by it. Else I would either port to `scil_image_math.py` or merge with `scil_compute_NODDI.py` directly.

frheault:
scil_image_math.py subtraction 1 MY_NIFTI with the --exclude_background option should do the trick?

I think it would be already working.

AlexVCaron:
We can close this, but for the sake of verbosity we need a set of probability operations in scil_image_math.py soon. The inverse is a simple one, but there are a lot of others which would be interesting. At least we should make existing ones work with more than binary images (union, intersection, ...).

PR #814, opened 2023-11-23
Fixing scil_visualize_bundles_mosaic.py with Pillow 10.0.*

karanphil:
Since our switch to Pillow==10.0.*, the script `scil_visualize_bundles_mosaic.py` was not working, apparently... the function `getsize()` from Pillow does not exist anymore, we should use `.getbbox()` or `.getlength()` instead. I adapted the script.

@AntoineTheb @frheault 

https://pillow.readthedocs.io/en/stable/releasenotes/9.2.0.html


arnaudbore:

AntoineTheb:
Tested locally, can confirm it works and behaviour is the same as before (as in, scilpy 1.5 before the bug was introduced).

AntoineTheb (approved):

Issue #816, opened 2023-11-28
--final_order not clear in new scil_gradients_modify_axes.py 


mdesco:
Currently, the -h does not indicate that final_order option needs the -- before it... 

Can you please harmonize as other scripts. Either it is mandatory positional arguments without --name needed or its in the options with --final_order indicated...

mdesco:
In fact, it seems that it does not handle -xyz order if the minus is at the beginning


- scil_gradients_modify_axes.py dwi.bvec dwi_x.bvec xyz

- scil_gradients_modify_axes.py dwi.bvec dwi_x.bvec -xyz
usage: scil_gradients_modify_axes.py [-h] [-f] in_gradient_sampling_file out_gradient_sampling_file final_order
scil_gradients_modify_axes.py: error: the following arguments are required: final_order

- scil_gradients_modify_axes.py dwi.bvec dwi_x.bvec --final_order -xyz
usage: scil_gradients_modify_axes.py [-h] [-f] in_gradient_sampling_file out_gradient_sampling_file final_order
scil_gradients_modify_axes.py: error: the following arguments are required: final_order

PR #821, opened 2023-11-30
[FIX] Use numbers rather than letters in scil_gradients_modify_axes

EmmaRenauld:
Tentatively fixing issue #816 

We did not find a way to use letters. The arpgarser does not understand text that starts with a minus.

Choices discussed to tell script to flip the x axis:
       1. `x` is normal direction, `X` is flipped.
       2. `x` is normal direction, `x-` is flipped
       3. `x` is normal direction, `ix` is 'inversed' x.
       4. Use numbers:  negative ints are understood. But uses axes = 1, 2, 3 because with 0, 1, 2, we can't inverse 0 with -0.

We chose 4 with @karanphil. @mdesco ok for you?

### DIFF
--- a/scripts/scil_gradients_modify_axes.py
+++ b/scripts/scil_gradients_modify_axes.py
@@ -24,7 +24,9 @@ def _build_arg_parser():
                    help='Path to gradient sampling file. (.bvec or .b)')
 
     p.add_argument('out_gradient_sampling_file',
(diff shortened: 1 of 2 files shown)


karanphil (approved):

arnaudbore:

mdesco (approved):
Thanks for the modifications

PR #824, opened 2023-12-01
Legacy installation

arnaudbore:
For dev purposes, if you don't want legacy scripts to be installed:

```
export SCILPY_LEGACY='False'
pip install -e .
```


EmmaRenauld:
Maybe we could add this information somewhere in a Readme?

arnaudbore:

Issue #831, opened 2023-12-05
compute_divide with --fa option 


mdesco:
Why do we need to provide --fa for the OP map? Why not compute the FA from the linear data provided to the command line?

scil_compute_divide.py --in_dwis ../sub-001_dwi_fwf_LTE.nii.gz ../sub-001_dwi_fwf_PTE.nii.gz --in_bvals ../sub-001_dwi_fwf_LTE.bval ../sub-001_dwi_fwf_PTE.bval --in_bvecs ../sub-001_dwi_fwf_LTE.bvec ../sub-001_dwi_fwf_PTE.bvec --in_bdeltas {1,-0.5} --mask b0_mask_mask.nii.gz --fa fa.nii.gz --processes 6

Thanks

PR #836, opened 2023-12-06
Moving OP out of the default output files from scil_btensor_metrics.py.

karanphil:
Following issue #831 from @mdesco, I removed the order parameter (OP) from the default output metrics. You now need to explicitly tell the OP output file path with `--op`, if not, it will not be computed. You get a parser error if the FA is not also given (with --fa), since it is needed to compute the OP.


arnaudbore:

arnaudbore:

scripts/scil_btensor_metrics.py
```diff
@@ -12,12 +12,16 @@
...
 
-By default, will output all possible files, using default names.
-Specific names can be specified using the file flags specified in the
-"File flags" section.
+By default, will output all possible files, using default names. Thus, this
+script outputs the results from the DIVIDE fit or direct derivatives:
+mean diffusivity (MD), isotropic mean kurtosis (mk_i), anisotropic mean
+kurtosis (mk_a), total mean kurtosis (mk_t) and finally micro-FA (uFA).
+Specific names can be specified using the
+file flags specified in the "File flags" section.
```
mdesco, line 20: Maybe add a sentence about OP + FA here even you repeat yourself slightly...

mdesco (approved):

PR #898, opened 2024-02-13
Fix weighted mean std when infinite values

arnaudbore:
# Quick description

Email from Po-Jui:

**_Mario and I found an unexpected behavior with the function weighted_mean_std in scilpy. np.average is used inside that function to compute the weighted mean with the image saved in a numpy masked array. In this case, the voxels of invalid nan values would be considered in the denominator of the weighted mean computation. To compute the weighted average on a masked numpy array without the invalid voxels, np.ma.average should be used.
Another thing we would like to propose is to check infinity values in the image, in addition to checking nan values. The proposed modification is in blue_**

...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/scilpy/utils/metrics_tools.py
+++ b/scilpy/utils/metrics_tools.py
@@ -157,9 +157,9 @@ def weighted_mean_std(weights, data):
         a tuple containing the mean and standard deviation of the data
     """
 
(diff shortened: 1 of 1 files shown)


karanphil (approved):
Good to go. I tested it with toy arrays. Does not work with None, but with np.nan.

PR #928, opened 2024-02-29
Cleaning scil_gradients_validate_correct handling of peaks

karanphil:
# Quick description

Small PR to remove the need for a `--peak_vals` argument in `scil_gradients_validate_correct.py` when giving multiple peaks per voxel. Since the peaks we use (from `scil_fodf_metrics.py`) are already ordered, we don't need the values to find the maximum amplitude. Also, it would work with evecs.nii.gz from `scil_dti_metrics.py`, but the format is different and we always have access to evecs_v1.nii.gz anyway, so I did not add support for it.

We will also need to add a test for multiple peaks, once we have cleaner test data.

...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/scripts/tests/test_gradients_validate_correct.py
+++ b/scripts/tests/test_gradients_validate_correct.py
@@ -16,7 +16,7 @@ def test_help_option(script_runner):
     assert ret.success
 
 
-def test_execution_processing(script_runner):
+def test_execution_processing_dti_peaks(script_runner):
     os.chdir(os.path.expanduser(tmp_dir.name))
     in_dwi = os.path.join(get_home(), 'processing',
                           'dwi_crop_1000.nii.gz')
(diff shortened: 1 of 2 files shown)


scripts/tests/test_gradients_validate_correct.py
```diff
@@ -33,3 +33,8 @@ def test_execution_processing(script_runner):
     ret = script_runner.run('scil_gradients_validate_correct.py', in_bvec,
                             'evecs_v1.nii.gz', 'fa.nii.gz', 'bvec_corr', '-v')
     assert ret.success
+
+
+def test_execution_processing_fodf_peaks():
```
arnaudbore, line 38: Why not doing it ?

arnaudbore (approved):
Short PR - Quick Review I liket it !

scripts/tests/test_gradients_validate_correct.py
karanphil: Done.

CHrlS98 (approved):

Issue #940, opened 2024-03-07
Image correlation correct?


EmmaRenauld:
Hi!

I tried to add the missing unit test in scil.image.tests.test_volume_math.py for the method correlation. I used:

```
def test_correlation():
    # Note. Not working on 2D data.
    img_data = np.zeros((3, 3, 3)).astype(float)
    affine = np.eye(4)
    img1 = nib.Nifti1Image(img_data, affine)
    img2 = nib.Nifti1Image(img_data + 1, affine)

    # With patch_radius of 1, we get the correlation for each voxel.
    output = correlation([img1, img2], img1, patch_radius=1)
    assert np.all(output == 1)
```

This fails. The output correlation is:
```
[[[-0.20883757 -0.04018329 -0.25135627]
  [-0.10100343 -0.13530837 -0.19371986]
  [ 0.1560971  -0.52060044  0.09855439]]

 [[-0.07638622  0.06555331 -0.15572111]
  [-0.24882393  0.16297379 -0.0308053 ]
  [-0.40140784  0.05187551 -0.09271187]]

 [[ 0.06262147  0.05383836  0.08995959]
  [-0.21807295  0.11019234 -0.26336023]
  [-0.0148444  -0.0608431   0.04018494]]]
```

which makes no sense to me. 

Also, please note that  currently, patch_radius is NEVER used in scilpy. To delete?

@frheault, you made the other tests in that file. Can you help?

frheault:
Can you try putting values everywhere in both image.

Background (0) will have random value to avoid NaN error, so right now you have a image with just background and one with data. That will not give a 1 result

EmmaRenauld:
If a compare a bunch of zeros with a bunch of one and I print the result for one patch, I get:

```
CORRELATION BETWEEN  [0.0005 0.0007 0.0006 0.0005 0.0004 0.0006 0.0004 0.0009 0.001  0.0004
 0.0008 0.0005 0.0006 0.0009 0.0001 0.0001 0.     0.0008 0.0008 0.0009
 0.001  0.0008 0.0005 0.0008 0.0001 0.0006 0.0001] 
AND
 [0.0009 0.0005 0.0004 0.0003 0.0008 0.0005 0.0006 0.     0.0006 0.0006
 0.0006 0.0009 0.0007 1.0004 1.0004 0.0007 1.0001 1.0007 0.0007 0.0002
 0.0001 0.0003 1.0004 1.0006 0.0004 1.001  1.0001]
[...]

PR #942, opened 2024-03-07
Volumes test

EmmaRenauld:
# Quick description

New tests for more code coverage, ye!!

Changes: 
- volume_math: 
      - Tests: added a test for the correlation method, was missing. Fixes issue #940 .
      - Methods: Moved sub-method (`def correlate`) out of the loop.
- volume operations:
     -  Tests: Added 3 tests
     -  Methods:
            -  `register_image`: My initials tests failed because my fake images were too small. Took me a long time to figure out. Added a verification. 
            - `compute_snr`: verified if both noise_mask and noise_map were given. Added more explanation.
- script scil_header_print_info: I wanted a faster way to print the affine. Added an option. You can tell me if there was an other way.

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/scilpy/image/tests/test_volume_math.py
+++ b/scilpy/image/tests/test_volume_math.py
@@ -523,18 +523,12 @@ def test_neighborhood_correlation():
         ("Expected a bad correlation at central point, good around the border,"
          " got: {}").format(output)
 
-    # Test 2: Different backgrounds.
-    # Image 1 = only ones. Image 2 = only background.
-    # Expecting a poor correlation where backgrounds are not the same.
-    # But we get a one in the middle (two uniform patches)
+    # Test 2: Comparing with only background: should be 0 everywhere.
     img_data_2 = np.zeros((3, 3, 3)).astype(float)
     img2 = nib.Nifti1Image(img_data_2, affine)
     output = neighborhood_correlation([img1, img2], img1)
-    expected = np.zeros((3, 3, 3))
-    expected[1, 1, 1] = 1
-    assert np.allclose(output, expected), \
-        ("Expected a good correlation at central point, bad around the border,"
-         " got: {}").format(output)
+    assert np.allclose(output, np.zeros((3, 3, 3))), \
+        "Expected a 0 correlation everywhere, got {}".format(output)
 
 
 def test_dilation():
--- a/scilpy/image/tests/test_volume_math.py
+++ b/scilpy/image/tests/test_volume_math.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-
+import logging
 
 import nibabel as nib
 import numpy as np
@@ -11,12 +11,11 @@ from scipy.ndimage import (binary_closing, binary_dilation,
                            gaussian_filter)
 
 
-from scilpy.image.volume_math import (_validate_imgs,
-                                      _validate_imgs_concat,
+from scilpy.image.volume_math import (_validate_imgs_type,
                                       _validate_length,
                                       _validate_type,
                                       _validate_float,
-                                      cut_up_cube,
+                                      _get_neighbors,
                                       lower_threshold_eq, upper_threshold_eq,
                                       lower_threshold, upper_threshold,
                                       lower_threshold_otsu,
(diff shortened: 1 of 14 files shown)


scilpy/image/tests/test_volume_math.py
```diff
@@ -572,6 +571,29 @@ def test_concatenate():
     assert_array_almost_equal(img_data_1.shape+(2,), output_data.shape)
 
 
+def test_correlation():
+    # Note. Not working on 2D data.
+    # Test 1: Data with background values
+    # Test 2: No background
+    # In both cases, compares a bunch of 0 with a bunch of 1, should be a
+    # perfect correlation.
+    img_data_1 = np.zeros((3, 3, 3)).astype(float)
```
frheault, line 580: Here it wont work with an image full of zeros, because 0 is assume to be background (i.e. no information), to ensure there is no division by zero in np.corrcoef and no resulting NaN we add noise to background.

Therefore the correlation between an image full of zeros and full of ones is very poor (because one is 100% background), an image full of ones and twos would work (no background)

EmmaRenauld: @frheault After doing many tests...

Actually, result of corr is NaN whenever one patch has only the same value everywhere. So instead of adding random values, I managed possible NaN values: 

- 0 if at least one neighborhood was entirely containing background (deals with the case where we are in union of backgrounds, so no need to check for that anymore).
 - 1 if the voxel's neighborhoods are uniform in both images
 - 0 if the voxel's neighborhoods is uniform in one image, but not the other.

Tell me if that's ok with you.

EmmaRenauld:
There were failing tests in test_reproducibility_measures. The error is fixed. But while searching for the error, I modified that other test quite a lot, and discovered that origin of SFT had an impact on results. Placed `sft.to_corner()` higher in `tractogram_pairwise_comparison`.

scilpy/image/volume_math.py
```diff
@@ -447,7 +481,8 @@ def multiplication(input_list, ref_img):
         Multiply multiple images together (danger of underflow and overflow)
     """
     _validate_length(input_list, 2, at_least=True)
-    _validate_imgs(*input_list, ref_img)
+    _validate_imgs_type(*input_list)
+    _validate_same_shape(*input_list, ref_img)
```
arnaudbore, line 485: do we really need to validate this here ? it should have been done in the script, no ?

arnaudbore (commented):
A couple of question but will need another reviewer.

scilpy/image/volume_operations.py
```diff
@@ -131,7 +137,8 @@ def apply_transform(transfo, reference, moving,
 
     Return
```
karanphil, line 138: Returns?

scilpy/image/volume_operations.py
```diff
@@ -189,7 +196,8 @@ def transform_dwi(reg_obj, static, dwi, interpolation='linear'):
...
```
karanphil, line 197: Returns?

scilpy/image/volume_operations.py
```diff
@@ -282,12 +300,10 @@ def register_image(static, static_grid2world, moving, moving_grid2world,
...
 
-def compute_snr(dwi, bval, bvec, b0_thr, mask,
-                noise_mask=None, noise_map=None,
-                split_shells=False,
-                basename=None):
+def compute_snr(dwi, bval, bvec, b0_thr, mask, noise_mask=None, noise_map=None,
+                split_shells=False):
     """
-    Compute snr
+    Computes the SNR. Saves
```
karanphil, line 306: Saves what?

scilpy/image/volume_operations.py
```diff
@@ -301,19 +317,26 @@ def compute_snr(dwi, bval, bvec, b0_thr, mask,
...
+        are given, we will try to discover a noise_mask automatically in the
+        background (from the upper half, to avoid using neck and shoulder).
+    noise_map: nib.Nifti1Image, optional
+        Noise map file in nibabel format. Only one of noise_mask or noise_map
+        may be used.
+    split_shells: bool
+        If true, we will only work with one b-value per shell (the discovered
+        centroids).
 
     Return
```
karanphil, line 331: Returns?

scripts/scil_dwi_compute_snr.py
```diff
@@ -120,12 +119,15 @@ def main():
...
-                         noise_mask=noise_mask,
-                         noise_map=noise_map,
-                         split_shells=args.split_shells,
-                         basename=basename)
+    values, noise_mask = compute_snr(dwi, bvals, bvecs, args.b0_thr,
+                                     mask, noise_mask=noise_mask,
+                                     noise_map=noise_map,
+                                     split_shells=args.split_shells)
+
+    if noise_mask is None and noise_map is None:
```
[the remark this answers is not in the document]
karanphil: not None I guess

karanphil (approved):
Looks good from what I understand...

scilpy/image/volume_math.py
EmmaRenauld: No, because in the script, we don't know which input is an image. It depends on the operation being called. That's why verifications are done in all sub-functions.

Issue #952, opened 2024-03-26
Lost options for PTT tracking


EmmaRenauld:
Method `scilpy.tracking.utils.add_tracking_options_ptt` is never used. 
https://github.com/scilus/scilpy/blob/master/scilpy/tracking/utils.py#L83

Could be added to `scil_tracking_local.py` but currently, its parameters (ex, `args.probe_*`) are never used. Possibly, `get_direction_getter` needs to be changed.

This comes from PR #750 .

Issue #698, opened 2023-03-24
Investigating Dipy tracking??


EmmaRenauld:
Hi.

Here is an image of the Fornix obtained with `scil_compute_local_tractogram` vs `_dev`, with exactly the same parameters, then segmented the Fornix.

Up row: dev. Det = 39 streamlines , Prob = 85 streamlines.
Bottom: dipy. Det = 6 streamllines. Prob = 29 streamlines.

For both Det and Prob, our _dev tracking seems to be filling the space better. Are we sure that interpolation of the mask is managed correctly in Dipy? Should we investigate? My result could be due to random, but if we have time to verify...

![image](https://user-images.githubusercontent.com/4967417/227569709-0216ad18-f8fa-4122-87a2-24529aedd572.png)

arnaudbore:
We should give it a try with the latest version and see if Gab's fix made a difference.

EmmaRenauld:
I didn't finish running the scilpy_tracking_local_det (sooo soooo slow!)

But comparing dipy's proba vs our dev_prob: Issue seems ok.
![dipy_prob](https://github.com/scilus/scilpy/assets/4967417/5fd0bfc8-8b08-42fe-baef-a23cf3dc6bfe)
![dev_prob](https://github.com/scilus/scilpy/assets/4967417/699be239-7845-471e-ad2d-2fdea1379959)

EmmaRenauld:
For comparison: dipy det. If one day I get the dev det, I'll show it here.
![dipy_det](https://github.com/scilus/scilpy/assets/4967417/09ae799b-1a6e-4cb4-ba7e-d66a00895bf9)

PR #953, opened 2024-03-26
Close issue 698

EmmaRenauld:
# Quick description

Remove the toDo, related to Issue #698 

Also noticed that an option was added in utils (`--sh_to_pmf`) but it is only used in the main tracking script, not in the _dev. Moved option.

Also added a comment about the missing ptt_options usage. See Issue #952.

### DIFF
--- a/scilpy/tracking/utils.py
+++ b/scilpy/tracking/utils.py
@@ -35,6 +35,10 @@ class TrackingDirection(list):
 
 
 def add_mandatory_options_tracking(p):
+    """
+    Args that are required in both scil_tracking_local and
+    scil_tracking_local_dev scripts.
+    """
     p.add_argument('in_odf',
                    help='File containing the orientation diffusion function \n'
                         'as spherical harmonics file (.nii.gz). Ex: ODF or '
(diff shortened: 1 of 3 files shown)


scilpy/tracking/utils.py
```diff
@@ -71,16 +79,13 @@ def add_tracking_options(p):
...
-                         help='If set, map sherical harmonics to spherical '
-                              'function (pmf) before tracking (faster, '
-                              'requires more memory)')
     add_sh_basis_args(track_g)
 
     return track_g
 
 
 def add_tracking_ptt_options(p):
+    """ NOT USED IN SCILPY!!! TO DELETE??? """
```
CHrlS98, line 88: Maybe we should take advantage of this PR to either remove or use this function?

CHrlS98 (commented):
A single comment related to issue #952 but appart from this, all good!

scilpy/tracking/utils.py
EmmaRenauld: I asked Gab what to do. He said he would check it. That's why I added it as issue instead.

CHrlS98: Ok, all good then

arnaudbore (approved):
GTG

PR #955, opened 2024-03-27
Fix remove invalid overlap pts

frheault:
# Quick description

This is an improvement of the remove_invalid_streamlines stuff in scilpy.
Laurent used RTT on mouse brain and all streamlines had overlapping points (0mm step size for one step, no idea why, my guess is the seed forward and backward are added twice). My previous code was removing the entire streamline (so empty results). A better solution is to discard such a point and leave the streamline visually intact.

I add two function next to the cut_invalid_streamlines() function in utils.streamlines. No idea why it is there and there is no test yet for all of these. I do not have the time to change the organization or to add test. But it is working on Laurent's data and gives the same results on typical data.

...

## Type of change

Check the relevant options.

...

# Checklist


arnaudbore (approved):
GTG - already gave my comments to Francois

PR #961, opened 2024-04-03
Tractogram scripts parts4

EmmaRenauld:
# Quick description

Modified the tractogram filtering scripts! Looks like a lot of changes but the core methods should not have changed. Only the managing of all inputs and options.

-  `scil_tractogram_filter_by_length`: ok. Cite other filtering scripts.
- `scil_tractogram_filter_by_orientation`: ok. Cite other filtering scripts.
- `scil_tractogram_filter_by_anatomy`:  *Will be difficult to review, sorry!*
     - Clarified doc.
     - There was a lot of repeated code. Merged into _finalize_step or into `_finish_all`.
     - Moved method `binarize_labels` to `image.labels.get_binary_mask_from_labels`
     - Removed method `dilate_mask`. Now uses scipy, like we do in `scil_volume_math.py`! Dilation ratio now in voxels. Confirmed by @frheault 
     - Method `compute_outliers`is not required anymore; I made sure that all filtering methods used also return the outlier streamlines or their ids.
     - Other methods (`save_intermediate_sft`, `save_rejected`, `display_count`, `save_count`) are now all included in `_finalize_step`.
- `scil_tractogram_filter_by_roi`: *Will be difficult to review, sorry!*
     - Clarified doc.
     - Removed the explanation that conditions will be applied sequentially. Not entirely true. We get the argparser's `--drawn_roi` and `--atlas_roi` separately, for instance. To do it sequentially, we would need to get the input *as is* and parse it ourselves. However, I did change the management of the ` filtering_list` option. Before, it was indeed sequential *only for the filtering_list options!* (and other options such as `--drawn_roi` were done before which was not clear from explanation!!). I can revert my changes if you want, to keep sequential management. But then, I would explain better in the doc that it's only for filtering_list, and I would prevent using filtering_list with any other ROI option. Any comment? @frheault 
     - Moved stuff around to answer the "toDo" that was written: `# Todo. Prepare now the names of other files (ex, ROI) and verify if exist and compatible.`
     - Question: Why is there an option --overwrite_distance?? Just weird for me. User could just call it correctly the first time! Still used? @frheault 

## Type of change

Check the relevant options.

# Checklist


scilpy/image/volume_operations.py
```diff
@@ -641,3 +641,36 @@ def merge_metrics(*arrays, beta=1.0):
...
 
     return ma.filled(boosted_mean, fill_value=np.nan)
+
+
+def dilate_mask(mask, mask_shape, vox_size, radius):
+    """
+    Dilate the foreground in a binary mask according to a radius (in mm)
+
+    toDo: in scil_volume_math, the DILATION option calls binary_dilation from
+     scipy.ndimage. Couldn't we use the same??
```
CHrlS98, line 651: Reformulate the todo so it sounds a bit less dramatic

scripts/scil_tractogram_filter_by_anatomy.py
```diff
@@ -134,61 +140,58 @@ def _build_arg_parser():
...
-    # Change values of those background
-    mask = mask.flatten()
-    mask[id_background.T] = mask[id_label.T]
-    mask = mask.reshape(mask_shape)
+    save_tractogram(final_sft, out_sft_name, args.no_empty)
+    if args.save_rejected:
+        save_tractogram(total_outliers, rejected_sft_name, args.no_empty)
 
-    return mask
+    exit(0)
```
CHrlS98, line 191: I think it would be better/clearer if the `exit(0)` was in the main.

scripts/scil_tractogram_filter_by_roi.py
```diff
@@ -75,28 +101,28 @@ def _build_arg_parser():
     p.add_argument('out_tractogram',
                    help='Path of the output tractogram file.')
 
-    p.add_argument('--drawn_roi', nargs='+', action='append',
+    p.add_argument('--drawn_roi', nargs='+', action='append', default=[],
```
CHrlS98, line 104: isn't `'append'` the default `action` for `nargs='+'`?

CHrlS98 (commented):
Minor comments

arnaudbore (approved):
Seems GTG, two huge scripts I would need to compare before merging.

scilpy/image/volume_operations.py
EmmaRenauld: Oups. That's actually already done. Will remove the todo.

scripts/scil_tractogram_filter_by_roi.py
EmmaRenauld: No, appends make it so that you can call the same option many times:
`--draw_roi option1 option2  --draw_roi option3 option4`

And the result will be `args.drawn_roi=[[option1, option2], [option3, option4]]`

EmmaRenauld:
Writing here, so that we don't forget:
As discovered by @arnaudbore, the new dilation usage creates a difference in the results are compared to the master branch. 
gm_mask on master: 839 238 voxels.
gm_mask here: 854 041 voxels.

scripts/scil_tractogram_filter_by_roi.py
CHrlS98: ah ok I get it

EmmaRenauld:
> Writing here, so that we don't forget: As discovered by @arnaudbore, the new dilation usage creates a difference in the results are compared to the master branch. gm_mask on master: 839 238 voxels. gm_mask here: 854 041 voxels.

The end of the discussion was: 

Should we modify the option in the script to have a --dilate_radius in number of voxel (= number of pass) like in scil_volume_math, or do we need to keep it in mm?

frheault (approved):
About your questions for the 'sequential' nature of scil_tractogram_filter_with_roi.py. I think it is not actually so important to mention if it is sequential or not since we only support logical AND so the results are the same.

PR #984, opened 2024-04-23
Delete script fix_trk

EmmaRenauld:
[WIP]

Todo. @arnaudbore , you told me you were managing this. Adding my attempt at changes here so that they are not lost. See my comment "hein???" in the main.


scripts/scil_tractogram_fix_trk.py
```diff
@@ -128,69 +134,72 @@ def main():
...
         # Startrack flips the TRK
         flip_axis = ['x']
-        new_sft.to_vox()
-        new_sft.streamlines._data -= get_axis_flip_vector(flip_axis)
-        new_sft = flip_sft(new_sft, flip_axis)
-        new_sft.to_rasmm()
-
-    else:
+        sft.to_vox()
+        sft.streamlines._data -= get_axis_flip_vector(flip_axis)  # --------------------_> HEin?
```
arnaudbore, line 160: @frheault can you help with this one please ?

arnaudbore: + L185

arnaudbore (approved):
@frheault can you double check this one please :)

scripts/scil_tractogram_fix_trk.py
frheault: I dont have data from StarTrack, but since there is a flip in X for everything we cannot use 'to_corner()' or 'to_center()' since it would shift it in the opposite direction. I believe this was the equivalent of centering the streamlines to the voxel before flipping.

DSI is flip in X/Y (LPS instead of RAS), so the shift is made to center streamlines to the voxel before the flip.

arnaudbore (approved):

PR #986, opened 2024-04-26
Fix verifications in volume_resample

EmmaRenauld:
Reverted recent change in scil_volume_resample, that input image and ref had to be compatible: not here! Ref is the reference to which to reshape.

- Added a test in the script test to use the --ref option.
- Added tests in the the unit test for all options in volume_resample
- Fixed option names in function resample_volume: arg `res` meant: the final shape of the image. Usually, resolution is the voxel shape! Renamed.


arnaudbore (approved):
GTG

scilpy/image/volume_operations.py
```diff
@@ -488,40 +491,38 @@ def resample_volume(img, ref=None, res=None, iso_min=False, zoom=None,
...
-            res = res * 3
-        new_zooms = tuple((o / r) * z for o, r,
-                          z in zip(original_res, res, original_zooms))
+    elif volume_shape is not None:
+        if iso_min or voxel_res:
+            raise ValueError(error_msg)
+        if len(volume_shape) == 1:
+            volume_shape = volume_shape * 3
+
+        new_zooms = tuple((o / r) * z for o, r, z in
```
karanphil, line 510: Perhaps `v` instead of `r` here?

scripts/scil_volume_resample.py
```diff
@@ -84,9 +84,20 @@ def main():
 
     img = nib.load(args.in_image)
 
+    ref_img = None
+    if args.ref:
+        ref_img = nib.load(args.ref)
+        # Must not verify that headers are compatible. But can verify that, at
+        # least, that the last column of their affine are compatible.
```
karanphil, line 91: Remove this `that`. It is already in the previous line.

karanphil: Also, `is compatible`

karanphil (approved):
LGTM! (just 2 small comments)

scripts/scil_volume_resample.py
EmmaRenauld: I think it's rather `affines are compatible`

PR #991, opened 2024-05-13
Tractogram scripts part6

EmmaRenauld:
# Quick description

Nearly the end for the tractogram scripts revision! Here, reviewing scripts:

- scil_tractogram_remove_invalid
- scil_tractogram_resample
    - MAJOR FIX: Since a change when `--tube_radius` was added, to decide whether we were upsampling or downsampling, there was: `if args.nb_streamlines > original_number or args.tube_radius:`. But there is a default value to `tube_radius`, so this was always True, and script was now always upsampling!!!
    - Since there is a default, and we check the value range, this check: `if not args.point_wise_std and not args.tube_radius` is not useful.
- scil_tractogram_seed_density_map
- scil_tractogram_split.

Missing scripts are: scil_tractogram_segment* (4 scripts). Do be done in another (big) PR.

...

## Type of change

Check the relevant options.

- [X] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist


scripts/scil_tractogram_resample.py
```diff
@@ -56,29 +63,33 @@ def _build_arg_parser():
...
-                                  action='store_true',
-                                  help='Keep invalid newly generated streamlines'
-                                       ' that may go out of the \nbounding box.')
+    up_group = p.add_argument_group('Upsampling params')
+    up_group.add_argument('--point_wise_std',
+                          type=ranged_type(float, 0, None, min_excluded=True),
+                          default=1,
+                          help='Noise to add to existing streamlines '
+                               'points to generate new ones [%(default)s].')
+    up_group.add_argument('--tube_radius',
```
frheault, line 76: For that particular case, isn't `, type=float, nargs='?', const=1,` enough. Like in `scil_tractogram_compute_density_map.py`
If you don't use it, its none
If you use it and type nothing its 1.0
If you use it and type something its your value

EmmaRenauld: I could, but the default was already 1 before my changes. What do you want the default to be if you are upsampling? To not do it, or keep as is; to do it with 1?

(Same question for point_wise_std)

arnaudbore: If you are upsampling it still has to be set if the user wants otherwise we don't use tube_radius.

For point_wise_std I guess it has to be default 1, otherwise without this or default 0 what the streamlines will look like ?

arnaudbore (approved):
Except the quick question about tube LGTM

scripts/scil_tractogram_resample.py
frheault: We can remove the default too, but I think the user needs a range of values that make sense (for tube radius)
A value is too small for tube_radius and point_wise_std and the streamline is identical and the tractogram will look weird (like all streamlines are duplicated exactly next to each other) and too big and you make streamline go through regions they should not.

I think the default was 1-1 just to show in the help that the order of magnitude (and in millimeter) user should use.

EmmaRenauld: Ok, so, if I understand, I could:
1) remove both defaults and set them to const
2) if we are upsampling, verify that at least one value has been set between point wise and tube 
3) verify that the values are in a normal range, else print a warning

What could be "normal" ranges for both values?

scripts/scil_tractogram_resample.py
```diff
@@ -3,28 +3,34 @@
...
-To prevent upsample if not desired use --never_upsample.
+- If the number of requested streamlines is lower than the streamline count:
+DOWNSAMPLE
+- If the number of requested streamlines is higher than the streamline count:
+UPSAMPLE  (to prevent upsampling when not desired, use --never_upsample).
 
-Can be useful to build training sets for machine learning algorithms, to
-upsample under-represented bundles or downsample over-represented bundles.
+Upsampling:
+    Works by generating new streamlines, either by adding gaussian noise to
```
CHrlS98, line 14: misses a "or" to that "either"

CHrlS98 (commented):
just the small doc issue

scripts/scil_tractogram_resample.py
AntoineTheb: Specifically, the "or" is generating streamlines via parallel transport.

scripts/scil_tractogram_resample.py
```diff
@@ -56,29 +63,33 @@ def _build_arg_parser():
 
     p.add_argument('--never_upsample', action='store_true',
                    help='Make sure to never upsample a tractogram.\n'
-                        'Useful when downsample batch of files using bash.')
+                        'Useful when downsampling a batch of files using '
+                        'bash.')
```
AntoineTheb, line 67: I would remove "using bash". Kind of implicit by mentionning a "batch of files" and not everyone uses bash (Mac uses zsh by default, for example)

scripts/scil_tractogram_resample.py
```diff
@@ -56,29 +63,33 @@ def _build_arg_parser():
...
+    up_group.add_argument('--point_wise_std',
+                          type=ranged_type(float, 0, None, min_excluded=True),
+                          default=1,
+                          help='Noise to add to existing streamlines '
+                               'points to generate new ones [%(default)s].')
+    up_group.add_argument('--tube_radius',
+                          type=ranged_type(float, 0, None, min_excluded=True),
+                          default=1,
+                          help='Maximum distance to generate streamlines '
+                               'around the original ones [%(default)s].')
```
AntoineTheb, line 80: I would add a unit there. What is "1" here ? 1 voxel ? 1mm ?

scripts/scil_tractogram_resample.py
```diff
@@ -109,32 +120,28 @@ def main():
...
-                         "--tube_radius is required")
-        sft = upsample_tractogram(sft, args.nb_streamlines,
-                                  args.point_wise_std, args.tube_radius,
-                                  args.gaussian, args.error_rate, args.seed)
+    # Processing
+    if args.nb_streamlines > original_number:
+        if args.never_upsample:
+            logging.info(
+                "Number of streamlines is higher than in the original "
+                "tractogram. Cannot downsample. Out_tractogram will be a copy "
```
AntoineTheb, line 137: Cannot ~down~upsample ?

scripts/scil_tractogram_split.py
```diff
@@ -122,12 +122,16 @@ def main():
         # streamlines. Should be of size close to 0. Not using it.
         sfts = split_sft_randomly_per_cluster(
             sft, chunk_sizes, args.seed, args.qbx_thresholds)
+        logging.info("Splitting per cluster may lead to a small variability "
+                     "in the final tractogram sizes. We got sizes: {}. "
```
AntoineTheb, line 126: ~We got sizes~ Size are: {}

AntoineTheb (changes requested):

scripts/scil_tractogram_resample.py
EmmaRenauld: (I am pushing changes now with a warning above 10. I can change again if needed)

scripts/scil_tractogram_resample.py
EmmaRenauld: I wrote mm. That's right, @frheault ?

EmmaRenauld:
Ok. After trying to fix things, I discovered more things to fix. So:
In `scil_tractogram_resample.py`:
- now point_wise_std and tube_radius are facultative (but at least one must be set if upsampling)
- before, the new streamlines were created starting from `rng.choice(len(sft), nb, replace=True)`. Issues:
    - With the replace, you have no ensurance that all of the initial streamlines are selected.
    - From these chosen streamlines, noise was added to ALL of them.
    - So you end up with a tractogram that contains none of your initial streamlines. Fixed. The initial streamlines are now untouched. Only the streamlines added from upsampling are created with noise.
- There was a `if len(s) < 3` that was faulty. Unit test added. If condition removed: everything works without that verification.

scripts/scil_tractogram_resample.py
arnaudbore: not sure I don't see any formula with the voxel resolution. Once we know we should update it to each function it uses it.

frheault: The operation is performed in word space (rasmm) so yes, it would be In mm

scripts/scil_tractogram_resample.py
frheault: That's the behavior I would expect, we can revise when Maxime try it if its not intuitive, but for now I think that perfect.

arnaudbore (approved):

PR #992, opened 2024-05-16
Alter tractogram (trim, cut, replace, subsample, transform)

frheault:
# Quick description

New script to simulate alteration of tractogram (should be used for bundle, so maybe I will rename)
The operation are describe in the docstring of the script:

```
This script is used to alter a bundle to reach specific minimum dice
coefficient from the original bundle. The script will subsample, trim, cut,
replace or transform the streamlines until the minimum dice is reached.
(cannot be combined in one run, use the script multiple times if needed)

All operation use a dichotomic search to find the parameter that gets as close
as possible to the specified minimum dice coefficient (with an epsilon for
convergence).

The subsample operation will remove streamlines until the minimum dice is
reached. This affect the whole bundle.

The trim operation will used the lowest density voxels (starting at 1) to
remove points from the streamlines until the minimum dice is reached.
This typically affect the edge of the bundle.

The cut operation will remove points from the start (or end) streamlines until
the minimum dice is reached. This affect one end of the bundle.

The replace operation will upsample the tractogram (generate new streamlines
with noise) and then subsample the tractogram. This effectively replace
streamlines with similar ones until the minimum dice is reached.
This affect the whole bundle.

The transform operation will apply random transformations to the streamlines
until the minimum dice is reached. This affect the whole bundle.
```

...

## Type of change

Check the relevant options.

...

# Checklist


scilpy/tractograms/tractogram_operations.py
```diff
@@ -898,6 +960,355 @@ def split_sft_randomly_per_cluster(orig_sft, chunk_sizes, seed, thresholds):
...
+        the original tractogram will be used.
+
+    Returns
+    -------
+    new_sft: StatefulTractogram
+        The tractogram with a subset of streamlines in the same space as the
+        input tractogram.
+    """
+    # Import in function to avoid circular import error
+    from scilpy.tractanalysis.reproducibility_measures import compute_dice_voxel
```
EmmaRenauld, line 990: This means that our management is not perfect yet.

`tractanalysis.reproducibility_measures.compute_dice_voxel` seems to be usable with any map. So it is not necessarily a tractanalysis method. Maybe it could go in `image.volume_operations` (or in volume_math if we also want to offer this option in scil_volume_math). Or in stats.something but I don't like it as much.

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+
```
EmmaRenauld, line 4: I would indeed rename. But not just bundle: the goal of this script is not simply to alter. I would imagine that a scil_tractogram_alter.py would ask user *how* to alter. Here it decides alone (if no option is provided). Can we put a longer, but more significative name? scil_bundle_alter_to_dice? 

Also, I see that in your new methods (ex, subsample_streamlines_alter), you have an option `baseline_sft=None` that is never used in your script. Can you add an option `--reference_bundle` in your script?

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+"""
+This script is used to alter a bundle to reach specific minimum dice
+coefficient from the original bundle. The script will subsample, trim, cut,
```
EmmaRenauld, line 7: It took me a little while to understand the goal of the script. Maybe something like this for the first sentence?

`This script is used to alter a bundle while keeping it similar enough to the original version (keeping the dice
coefficient above a given threshold).`

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+replace or transform the streamlines until the minimum dice is reached.
+(cannot be combined in one run, use the script multiple times if needed)
```
EmmaRenauld, line 9: This is not clear for me. What cannot be combined? Do you mean Cannot be used on multiple bundles in one run? Or the various operations choices cannot be combined?

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+All operation use a dichotomic search to find the parameter that gets as close
```
EmmaRenauld, line 11: operations.

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+as possible to the specified minimum dice coefficient (with an epsilon for
+convergence).
+
+The subsample operation will remove streamlines until the minimum dice is
+reached. This affect the whole bundle.
```
EmmaRenauld, line 16: affects

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+The trim operation will used the lowest density voxels (starting at 1) to
```
EmmaRenauld, line 18: will use

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+remove points from the streamlines until the minimum dice is reached.
+This typically affect the edge of the bundle.
```
EmmaRenauld, line 20: affects the edges

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+The cut operation will remove points from the start (or end) streamlines until
+the minimum dice is reached. This affect one end of the bundle.
```
EmmaRenauld, line 23: affects.

With the current description, I think that the difference between cut and trim is not clear. The vocabulary: edge of the bundle, end of the bundle, probably sound like the same thing to many users.

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+The replace operation will upsample the tractogram (generate new streamlines
+with noise) and then subsample the tractogram. This effectively replace
+streamlines with similar ones until the minimum dice is reached.
+This affect the whole bundle.
+
+The transform operation will apply random transformations to the streamlines
+until the minimum dice is reached. This affect the whole bundle.
+"""
```
EmmaRenauld, line 32: Please cite the other scripts. 

Ex, "For more description of the upsampling parameters, see scil_tractogram_resample"

------------

Also, can you add: the operations are performed in the following order: 1) subsampling, 2) trimming, 3) cutting, 4) upsampling

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+The trim operation will used the lowest density voxels (starting at 1) to
+remove points from the streamlines until the minimum dice is reached.
+This typically affect the edge of the bundle.
+
+The cut operation will remove points from the start (or end) streamlines until
+the minimum dice is reached. This affect one end of the bundle.
+
+The replace operation will upsample the tractogram (generate new streamlines
+with noise) and then subsample the tractogram. This effectively replace
```
EmmaRenauld, line 26: replaces

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+streamlines with similar ones until the minimum dice is reached.
+This affect the whole bundle.
```
EmmaRenauld, line 28: affects

scilpy/tractograms/tractogram_operations.py
```diff
@@ -24,15 +25,35 @@
...
 
 
 def shuffle_streamlines(sft, rng_seed=None):
+    """
+    Shuffle the streamlines of a tractogram.
+
+    Parameters
+    ----------
+    sft: StatefulTractogram
+        The tractogram to shuffle (will slice the streamline, DPS and DPP).
```
EmmaRenauld, line 49: "to slice" is a python explanation. Can we avoid nerdy terms? Simply, will shuffle?

scilpy/tractograms/tractogram_operations.py
```diff
@@ -47,6 +68,47 @@ def shuffle_streamlines(sft, rng_seed=None):
...
+    rng_seed: int
+        Random seed.
+
+    Returns
+    -------
+    shuffled_sft: StatefulTractogram
+        The shuffled tractogram.
+    """
+    if sft.data_per_point is not None and len(sft.data_per_point) > 0:
+        logging.warning('Shuffling streamlines orientation. DDP will be '
```
EmmaRenauld, line 90: DPP

scilpy/tractograms/tractogram_operations.py
```diff
@@ -47,6 +68,47 @@ def shuffle_streamlines(sft, rng_seed=None):
...
+                        'lost.')
+
+    rng = np.random.RandomState(rng_seed)
+    streamlines = sft.streamlines
```
EmmaRenauld, line 94: No need for that variable. sft.streamlines is only four letters more.

scilpy/tractograms/tractogram_operations.py
```diff
@@ -47,6 +68,47 @@ def shuffle_streamlines(sft, rng_seed=None):
     return shuffled_sft
 
 
+def shuffle_streamlines_orientation(sft, rng_seed=None):
+    """
+    Shuffle the orientation of the streamlines. The streamlines are not
```
EmmaRenauld, line 73: It does not shuffle all streamlines. "Shuffle some streamlines randomly..."

scilpy/tractograms/tractogram_operations.py
```diff
@@ -47,6 +68,47 @@ def shuffle_streamlines(sft, rng_seed=None):
...
+    reversed, only the orientation of the points within the streamlines is
+    shuffled.
```
EmmaRenauld, line 75: I am not sure I understand the difference.

scilpy/tractograms/tractogram_operations.py
```diff
@@ -898,6 +960,355 @@ def split_sft_randomly_per_cluster(orig_sft, chunk_sizes, seed, thresholds):
...
+    new_sft = StatefulTractogram.from_sft(streamlines, sft)
+    new_sft.to_space(space)
+    new_sft.to_origin(origin)
+    return compress_sft(new_sft)
+
+
+def replace_streamlines_alter(sft, min_dice=0.90, epsilon=0.01):
+    """
+    Replace streamlines based on a dice similarity metric.
+    The function will upsamples the streamlines (with parallel transport),
```
EmmaRenauld, line 1109: upsample

scilpy/tractograms/tractogram_operations.py
```diff
@@ -898,6 +960,355 @@ def split_sft_randomly_per_cluster(orig_sft, chunk_sizes, seed, thresholds):
...
+                  'downsample.')
+    upsampled_sft = upsample_tractogram(sft, len(sft) * 2, point_wise_std=0.5,
+                                        tube_radius=1.0, gaussian=None,
+                                        error_rate=0.1, seed=1234)
+    return subsample_streamlines_alter(upsampled_sft, min_dice, epsilon,
+                                       baseline_sft=sft)
+
+
+def trim_streamlines_alter(sft, min_dice=0.90, epsilon=0.01):
+    # Import in function to avoid circular import error
```
EmmaRenauld, line 1141: Missing docstring

scripts/scil_tractogram_alter.py
```diff
@@ -0,0 +1,156 @@
...
+
+The transform operation will apply random transformations to the streamlines
+until the minimum dice is reached. This affect the whole bundle.
```
EmmaRenauld, line 31: In the description of the method, you only mention rotations. Here, you say any transformation. Can you be more precise?

scripts/tests/test_tractogram_alter.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+def test_help_option(script_runner):
```
EmmaRenauld, line 4: More tests required.

EmmaRenauld (changes requested):
Round 1 of comments. I didn't really read the code yet, only the descriptions.

Pep8 also needs to be addressed.

scripts/scil_tractogram_alter.py
frheault: I am not sure what you mean by "Here it decides alone (if no option is provided)."

I agree that the name  scil_bundle_alter.py does not mean much
Here are a few more suggestions:
- scil_bundle_alter_preserve_overlap.py
- scil_bundle_alter_optimize_overlap.py
- scil_bundle_alter_to_target_dice.py

(alter could be replaced by adjust, vary, adapt, modify, but I like alter because of the definition: change or cause to change in character or composition, typically in a **comparatively small but significant way**)

baseline_sft: It is used in another function (replace)

scripts/scil_tractogram_alter.py
frheault: various operations choices cannot be combined?

scilpy/tractograms/tractogram_operations.py
frheault: Yes its unclear, it is not a blanket reverse all streamlines, it is a shuffle head/tail of random streamlines

scilpy/tractograms/tractogram_operations.py
frheault: I believe that it makes more sense to have the `compuet_dice_voxel` and `compute_dice_streamline`, and the bundle adjacency voxel and bundle adjacency voxel next to each other than splitting them because of the module name 'tractanalysis'.

Like `min_dist_to_centroid` works with point cloud, not just streamline. or `compute_f1_score` or `compute_f1_overlap_overreach` works with mask too. I believe it is alright since almost everything we do in scilpy is tract-oriented, even if some functions could be used differently sometimes.

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+multiple times if needed.
+
+All operations use a dichotomic search to find the parameter that gets as close
+as possible to the specified minimum dice coefficient (with an epsilon for
+convergence).
+
+- The subsample operation will remove streamlines until the minimum dice is
+  reached. This affect the whole bundle.
+
+- The trim operation will used the lowest density voxels (starting at 1) to
```
EmmaRenauld, line 21: will use

There are also still five "This affect" instead of "This affects".

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+  remove points from the streamlines until the minimum dice is reached.
+  This typically affect the edge of the bundle.
+
+- The cut operation will remove points from the start (or end) streamlines until
```
EmmaRenauld, line 25: Careful to pep8.

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+  the minimum dice is reached. This affect one end of the bundle.
+
+- The replace operation will upsample the tractogram (generate new streamlines
+  with noise) and then subsample the tractogram. This effectively replace
```
EmmaRenauld, line 29: replaces

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+    g1.add_argument('--cut', action='store_true',
+                    help='Cut streamlines from endpoints.')
+    g1.add_argument('--replace', action='store_true',
+                    help='Replace streamlines with similar ones.')
+    g1.add_argument('--transform', action='store_true',
+                    help='Transform streamlines using a random linear '
+                         'transformation.')
+    g.add_argument('--from_end', action='store_true',
+                   help='Cut from the other end of streamlines.\n'
+                        'Only available with --cut.')
```
EmmaRenauld, line 85: Not sure I understand the explanation...

EmmaRenauld (commented):
Clearer, thanks!

Still a few grammar errors to be corrected, and I think the circular error should be managed, but other than that, LGTM!

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+                             assert_outputs_exist)
+from scilpy.tractograms.streamline_operations import \
+    remove_overlapping_points_streamlines, \
+    remove_single_point_streamlines, cut_invalid_streamlines
+from scilpy.tractograms.tractogram_operations import transform_streamlines_alter, \
+    trim_streamlines_alter, cut_streamlines_alter, subsample_streamlines_alter, \
+    replace_streamlines_alter, shuffle_streamlines, shuffle_streamlines_orientation
+
+
+def buildArgsParser():
```
arnaudbore, line 57:
```suggestion
def _build_arg_parser():
```

scripts/scil_bundle_alter_to_target_dice.py
```diff
@@ -0,0 +1,159 @@
...
+
+    add_overwrite_arg(p)
+    add_verbose_arg(p)
+    add_reference_arg(p)
+
+    return p
+
+
+def main():
+    parser = buildArgsParser()
```
arnaudbore, line 104:
```suggestion
    parser = _build_arg_parser()
```

arnaudbore (approved):
2 comments from Em and 2 commit suggestions from me and we merge !!

scilpy/tractograms/tractogram_operations.py
arnaudbore: @frheault ?

PR #994, opened 2024-05-22
New script to compute similarity between mask and atlas (nii)

frheault:
# Quick description

This is similar to scil_bundle_pairwise_comparison.py, but for mask and atlas (as nifiti). This is an easy way to quantify the similarity between wmparc of bundle label maps or BET mask in scan-rescan or across subjects for example.

This is generating a JSON file that contains four measures: dice, adjacency_voxels, overlap and overreach.

Everything has to be in uint8 or uint16, coregistered nifti.

...

## Type of change

Check the relevant options.

...

# Checklist


scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,202 @@
...
+    add_processes_arg(p)
+    add_reference_arg(p)
+    add_json_args(p)
+    add_verbose_arg(p)
+    add_overwrite_arg(p)
+
+    return p
+
+
+def compute_all_measures(args):
```
AntoineTheb, line 68: This function should handle loading and calling the appropriate sub module. Also needs docstring

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,202 @@
...
+    _, _, voxel_size, _ = get_reference_info(img_1)
+    voxel_size = np.product(voxel_size)
+
+    # Exclude 0 (background)
+    unique_values_1 = np.unique(data_1)[1:]
+    unique_values_2 = np.unique(data_2)[1:]
+    union_values = np.union1d(unique_values_1, unique_values_2)
+
+    dict_measures = {}
+    for val in union_values:
```
AntoineTheb, line 94: This should go into a module or reuse code that is probably already in scilpy.

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,202 @@
...
+def compute_all_measures(args):
+    filename_1 = args[0][0]
+    filename_2 = args[0][1]
+    adjency_no_overlap = args[1]
+    ratio = args[2]
+
+    img_1, dtype_1 = load_img(filename_1)
+
+    if np.issubdtype(dtype_1, np.floating):
+        raise ValueError('Input file {} is of type float.'.format(filename_1))
```
AntoineTheb, line 77: Why only float ? The comments in the header of the script mention that the inputs must be uint8 or uint16, but here int32 would pass the test.

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,202 @@
...
+        # These measures are in mm
+        bundle_adjacency_voxel = compute_bundle_adjacency_voxel(
+            binary_1, binary_2,
+            non_overlap=adjency_no_overlap)
+
+        # These measures are between 0 and 1
+        dice_vox, _ = compute_dice_voxel(binary_1,
+                                         binary_2)
+
+        measures_name = ['adjacency_voxels',
```
AntoineTheb, line 119: Could be put elsewhere outside this function

AntoineTheb (changes requested):
First pass, broad comments

scripts/scil_volume_pairwise_comparison.py
frheault: Similar code is in the bundle pairwise comparison, should I refactor both in a separate PR?

This does not fit nicely anywhere, except tractanalysis/reproducibility_measures except it the current script can be used with anything. 

It could be in utils/workflow (new file) if you believe some high-level complex function that uses 3-4 modules at once should be somewhere to be imported.

scripts/scil_volume_pairwise_comparison.py
frheault: This was simply to make the error message a bit clearer in the context of this script (if used from Freesurfer atlas for example)

The function get_data_as_labels() will convert int32 to uint16 automatically

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+"""
+Evaluate pair-wise similarity measures of masks and atlas.
+All volumes must be in the same space.
+
+For the voxel representation, the computed similarity measures are:
+    adjacency_voxels, dice_voxels, volume_overlap, volume_overreach.
+
+This can work for BET mask, WMPARC, bundle label maps. The datatype of the
+input volumes must be uint8 (mask) or uint16 (label map and atlas).
+"""
```
EmmaRenauld, line 13: Add a reference to scil_bundle_pairwise_comparison

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Evaluate pair-wise similarity measures of masks and atlas.
+All volumes must be in the same space.
+
+For the voxel representation, the computed similarity measures are:
```
EmmaRenauld, line 8: You copied the text from the other script. But here, you don't have any other representation, so it sounds a little weird to have "For the voxel representation"

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+from scilpy.tractanalysis.reproducibility_measures import compute_dice_voxel, \
+    compute_bundle_adjacency_voxel
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_volumes', nargs='+',
+                   help='Path of the input volumes.')
```
EmmaRenauld, line 45: volume(s)?

Or if there should always be at least two, please add a verification and add it in the explanation, please.

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+    p.add_argument('out_json',
+                   help='Path of the output json file.')
+
+    p.add_argument('--adjency_no_overlap', action='store_true',
+                   help='If set, do not count zeros in the average BA.')
```
EmmaRenauld, line 50: What is BA? Not defined anywhere.

I think the variable's name is not super clear. `--ignore_zeros_in_BA, help="If set, do not count zeros in the average bundle adgency (BA)`?

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+
+    p.add_argument('--single_compare',
```
EmmaRenauld, line 52: I would put:
`    p.add_argument('--single_compare', metavar='volume_file',`
`                   help='Compare inputs to this single file. If not used, all files are compared two-by-two')`

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+                   help='Compare inputs to this single file.')
+    p.add_argument('--ratio', action='store_true',
+                   help='Compute overlap and overreach as a ratio over the '
+                        'reference volume.\nCan only be used if also using '
+                        '--single_compare`.')
```
EmmaRenauld, line 57: Else, they are computed in terms of number of voxels.

scripts/scil_volume_pairwise_comparison.py
EmmaRenauld: If you can try to refactor so that the loading is kept in this file (lines ~69-90), but the computation (92-142) is done somewhere else, it would be best, I think.

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+
+    return dict_measures
+
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+    logging.getLogger().setLevel(logging.getLevelName(args.verbose))
+
+    assert_inputs_exist(parser, args.in_volumes, args.reference)
```
EmmaRenauld, line 150: If all inputs are volumes (no tck inputs), there is no need for the --reference arg. Should be removed in line 60, 150 and 153. Or did I miss something in the explanation of the inputs?

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def compute_all_measures(args):
+    filename_1 = args[0][0]
+    filename_2 = args[0][1]
+    adjency_no_overlap = args[1]
+    ratio = args[2]
+
+    img_1, dtype_1 = load_img(filename_1)
+
```
EmmaRenauld, line 75: Could you please add a bit more verbose stuff. Ex, : `logging.info("Comparing file X and file Y")`

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+    if args.ratio and not args.single_compare:
+        parser.error('Can only compute ratio if also using `single_compare`')
+
+    nbr_cpu = validate_nbr_processes(parser, args)
+    if nbr_cpu > 1:
+        pool = multiprocessing.Pool(nbr_cpu)
+
+    if args.single_compare:
+        # Move the single_compare only once, at the end.
```
EmmaRenauld, line 163: This comment is not so clear for someone who doesn't know how the itertool.product works (me!)

`# Remove the single_compare from inputs and combine it to all other files.`

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+
+"""
+Evaluate pair-wise similarity measures of masks and atlas.
+All volumes must be in the same space.
+
+For the voxel representation, the computed similarity measures are:
+    adjacency_voxels, dice_voxels, volume_overlap, volume_overreach.
+
+This can work for BET mask, WMPARC, bundle label maps. The datatype of the
+input volumes must be uint8 (mask) or uint16 (label map and atlas).
```
EmmaRenauld, line 12: Could there be a --force_input_type option, if you don't want to save your file on disk with other input type?

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,206 @@
...
+
+    dict_measures = {}
+    for val in union_values:
+        binary_1 = np.zeros(data_1.shape, dtype=np.uint8)
+        binary_1[data_1 == val] = 1
+        binary_2 = np.zeros(data_2.shape, dtype=np.uint8)
+        binary_2[data_2 == val] = 1
+
+        # These measures are in mm^3
+        volume_overlap = np.count_nonzero(binary_1 * binary_2)
```
EmmaRenauld, line 105: See if you could use:
scilpy.tractanalysis.scoring.compute_f1_overlap_overreach, which can be tested with unit tests.

scripts/scil_volume_pairwise_comparison.py
EmmaRenauld: Actually, the word "adjacency" is only used in these two scripts (based on scil_search_keywords). I'm not even sure what it means. Could you define it?

scripts/scil_volume_pairwise_comparison.py
frheault: Without --single_compare you need at least two, if using --single_compare you can have one (Added to the docstring)

scripts/scil_volume_pairwise_comparison.py
frheault: Almost all of the scilpy scripts enforce labels as uint16 and mask as uint8, I think we want to force user to change their data type.

scripts/scil_volume_pairwise_comparison.py
frheault: Since this does not always use <ratio>, it is not always normalized between 0 - 1 (this script is not for GT). So I will leave it like that.

frheault:
@ThoumyreStanislas Could you try this branch before merging to compare atlas (labels of bundle in scan-rescan, I think you have penthera) and maybe lesions masks?

That would be helpful to improve the docstring

scripts/scil_volume_pairwise_comparison.py
```diff
@@ -0,0 +1,170 @@
...
+                             validate_nbr_processes)
+from scilpy.image.labels import get_data_as_labels
+from scilpy.tractanalysis.reproducibility_measures import compare_volume_wrapper
+
+
+def _build_arg_parser():
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawTextHelpFormatter)
+    p.add_argument('in_volumes', nargs='+',
```
arnaudbore, line 57: I think you should put `nargs='*' ` which is 2 or more volumes. You don't want only one volume.

scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -667,3 +667,156 @@ def tractogram_pairwise_comparison(sft_one, sft_two, mask, nbr_cpu=1,
     heatmap = merge_metrics(acc_data, corr_data, diff_data_norm)
 
     return acc_data, corr_data, diff_data_norm, heatmap, mask
+
+
+def compare_volume_wrapper(data_1, data_2, voxel_size=1, ratio=False,
+                           adjency_no_overlap=False):
+    # Exclude 0 (background)
```
arnaudbore, line 674: some docstring would be nice

scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -667,3 +667,156 @@ def tractogram_pairwise_comparison(sft_one, sft_two, mask, nbr_cpu=1,
...
+    return dict_measures
+
+
+def compare_bundle_wrapper(density_1, density_2, endpoints_density_1,
+                           endpoints_density_2, bundle_1, bundle_2,
+                           centroids_1=None, centroids_2=None, voxel_size=1,
+                           ratio=False, streamline_dice=False,
+                           disable_streamline_distance=False,
+                           bundle_adjency_no_overlap=False):
+    # These measures are in mm^3
```
arnaudbore, line 733: here as well

arnaudbore (commented):
A couple of things to fix, but we are getting closer !

PR #1005, opened 2024-06-22
Add create surface

arnaudbore:
# Quick description

First attempt to bring some of the nice scripts from Etienne to scilpy.
I choose to reduce the scope of the original script to fit scilpy requirements
I kept the expected behavior of the script used in nf-set.

...

## Type of change

Check the relevant options.

Use scilpy atlas test data
...

# Checklist


scripts/scil_surface_from_labels.py
```diff
@@ -0,0 +1,141 @@
...
+
+    assert_inputs_exist(parser, args.in_label)
+    assert_outputs_exist(parser, args, args.out_surface)
+
+    # Load volume
+    labels_img = nib.load(args.in_label)
+    labels_volume = get_data_as_labels(labels_img)
+
+    # Removed indices
+    mask = get_binary_mask_from_labels(labels_volume, args.indices)
```
frheault, line 104: As of now, this is crashing unless given (so I would use required at least). 
But since this is a script <from_labels> I think the default should be:
generate all the labels (from np.unique() then the script does the for loop itself) into an output folder, if the user wants just a few then --indices uses just one.
similar to the split labels script we could use an --out_dir OR --out_suffix .

This is my recommendation because if you run this on Freesurfer atlas, most people don't know the labels one by one and they probably want to generate a bunch of independent surface. And if they want to merge a few labels into a single mask, I propose that @arnaudbore you reuse the same logic as in scil_tractogram_filter_with_roi.py (like --indices 1 2 3 4 (do only 4 surfaces instead of all) VS --merge_indices 1:6 9 10:15 (do a single surface from a bunch of labels) and these two would be mutually exclusive options (default being all labels as multiple independent surfaces)

scripts/scil_surface_from_labels.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Script to convert segmented volume to a surface with marching cube'
+
+Example : use wmparc.a2009s.nii.gz with some aseg.stats indices
+
+scil_surface_from_volume.py s1a1/mask/S1-A1_wmparc.a2009s.nii.gz\\
+    -v -index 16  --vox2vtk -opening 2 -smooth 2
```
frheault, line 10: index?

--opening --smooth

frheault (commented):
I believe this script should work with a single mask (uint8) as well as labels (uint16), Like if I want a glass brain right now I have to artificially convert my BET mask to unint16

scripts/scil_surface_from_labels.py
arnaudbore: Here is my trade off: if you want 4 surfaces you use a loop otherwise it's now working as in scil_tractogram_filter_with_roi.py and combine every single indices provided with the possibility of using 1:6 10 22

scripts/scil_surface_from_labels.py
```diff
@@ -0,0 +1,165 @@
...
+
+    p.add_argument('--value', default=0.5,
+                   type=ranged_type(float, 0, None, min_excluded=False),
+                   help='Isosurface threshold value used. '
+                        'This value is called isovalue '
+                        'in mbcube. [%(default)s]')
+
+    # Need a group here
+    morpho_g = p.add_argument_group('Morphology options')
+    morpho_g.add_argument('--smooth', default=0.0,
```
EmmaRenauld, line 67: I think that for all options below, it would be more like our things elsewhere to put no default, and below, to check if args.smooth is not None rather than if >0. It avoids an error where user could put a negative value and it will ignore it without saying anything.

EmmaRenauld (commented):
Looks good, but I didn't test, as I am not so sure how to manage surfaces.

Could we / should we add an explanation on how to see results? Ex, say that you can load in this or that software?

scripts/scil_surface_from_labels.py
```diff
@@ -0,0 +1,165 @@
...
+                   help='Output surface (.vtk)')
+
+    p.add_argument('--indices', nargs='+',
+                   help='List of labels indices to use for the surface.')
+
+    p.add_argument('--value', default=0.5,
+                   type=ranged_type(float, 0, None, min_excluded=False),
+                   help='Isosurface threshold value used. '
+                        'This value is called isovalue '
+                        'in mbcube. [%(default)s]')
```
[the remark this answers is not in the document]
arnaudbore: No :D @StongeEtienne ?

scripts/scil_surface_from_labels.py
frheault: How does it work if you want all surfaces from Desikan? You have to know and write 
`for i in range {7..72} {1001..1035} {2001..2035}; do scil_surface_from_labels.py ...; done` and then get around 50 errors (some values do not exist) ?

I think it needs a `--do_all_labels_indepependantly` kind of option (which implies providing no option for indices, or at least ignores it) and a for loop in Python.

Not all atlas are contiguous (so very hard to do with {1..10}), some atlas have hundreds of labels (long to list) and it's kind of annoying to have a labels_list.txt (like in connectoflow) on hand to do `for i in $(cat labels_list.txt); do scil_surface_from_labels.py ...; done`, especially if you have a custom created atlas or kind of random values.

scripts/scil_surface_from_labels.py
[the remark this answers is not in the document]
StongeEtienne: It is the "threshold" value: the isosurface = level set.
For a binary mask (with 0 and 1) , 0.5 will generate a surface in the middle of the transition.
https://github.com/pmneila/PyMCubes/tree/master

However, it "could" be used to create a surface at any "level-set" for a volume with float values.

Currently in the script, since only "binary" format is given, 0.5 could be fixed.
(this was not the behavior in my previous script)

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+                          help='Dilation: number of iterations. [%(default)s]')
+    morpho_g.add_argument('--opening',
+                          type=ranged_type(int, 0, None, min_excluded=False),
+                          help='Opening (dilation of the erosion): number '
+                               'of iterations. [%(default)s]')
+    morpho_g.add_argument('--closing',
+                          type=ranged_type(int, 0, None, min_excluded=False),
+                          help='Closing (erosion of the dilation): number '
+                               'of iterations. [%(default)s]')
+
```
StongeEtienne, line 88: --max_label   option was removed,
Maybe this was wanted, it was useful to remove spurious voxel from a mask 
(was not use for SET)

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+    p.add_argument('--fill', action='store_true',
+                   help='Fill holes in the image.')
+
+    p.add_argument('--vox2vtk', action='store_true',
```
StongeEtienne, line 92: now the option is "reversed", by default surface are not in voxel space they are in "vtk" /  LPS world space.
The behavior should be change, or the option renamed
(this is critical for SET)

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+    p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
+                                formatter_class=argparse.RawTextHelpFormatter)
+
+    g = p.add_argument_group("Input (Labels or Mask)")
+    mxg = g.add_mutually_exclusive_group(required=True)
+    mxg.add_argument('--in_labels',
+                     help='Path of the atlas (nii or nii.gz). '
+                     'It will use the indices provided with --indices. '
+                     'If no indices are provided, it will use all indices.')
+    mxg.add_argument('--in_mask',
```
StongeEtienne, line 54: why both option ? a mask is simply a binary label volume ? 
"get_data_as_labels()" should work on both, so a simple "volume_input" could be use

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+        if args.indices:
+            mask = merge_labels_into_mask(labels_volume,
+                                          ' '.join(args.indices))
+        else:
+            logging.warning('No indices provided, '
+                            'it will use all indices.')
+            mask = merge_labels_into_mask(labels_volume,
+                                          "1:" + str(labels_volume.max()))
+    else:
+        # Load mask
```
StongeEtienne, line 124: see my previous comment on "mask"

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+        mask = binary_fill_holes(mask)
+
+    # Extract marching cube surface from mask
+    vertices, triangles = mcubes.marching_cubes(mask, args.value)
+
+    # Generate mesh
+    mesh = trimeshpy.trimesh_vtk.TriMesh_Vtk(triangles.astype(int), vertices)
+
+    # Transformation based on the Nifti affine
+    if not args.vox2vtk:
```
StongeEtienne, line 148: see my previous comment on "vox2vtk", which is now reversed since a "not" was added in front of the argument

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+        if args.in_labels:
+            mesh.set_vertices(vtk_u.vox_to_vtk(mesh.get_vertices(),
+                                               labels_img))
+        else:
+            mesh.set_vertices(vtk_u.vox_to_vtk(mesh.get_vertices(),
+                                               mask_img))
```
StongeEtienne, line 154: see my previous comment on "mask"

StongeEtienne (commented):
1) I think --in_labels and --in_mask, can be a single volume input (instead of an option).
2) There is an issue with the "vox2vtk", since now using the "--vox2vtk" tag keep it in voxel space, and by default it transform the surface to VTK world LPS space (aligned in mi-brain).

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+def _build_arg_parser():
+    p = argparse.ArgumentParser(description=__doc__, epilog=EPILOG,
+                                formatter_class=argparse.RawTextHelpFormatter)
+
+    g = p.add_argument_group("Input (Labels or Mask)")
+    mxg = g.add_mutually_exclusive_group(required=True)
+    mxg.add_argument('--in_labels',
+                     help='Path of the atlas (nii or nii.gz). '
+                     'It will use the indices provided with --indices. '
+                     'If no indices are provided, it will use all indices.')
```
StongeEtienne, line 53: "it will use all indices." is not completely true, since it takes all positives  >= 1 values.
It could be rewritten to "it use all positive indices".
(Yes some label volumes have negative values, some freesurfer labels too; but no scilpy does not seems to support it since label_volume are loaded as uint16 in get_data_as_labels() )

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Removed indices
+        if args.indices:
+            mask = merge_labels_into_mask(labels_volume,
+                                          ' '.join(args.indices))
+        else:
+            logging.warning('No indices provided, '
+                            'it will use all indices.')
+            mask = merge_labels_into_mask(labels_volume,
+                                          "1:" + str(labels_volume.max()))
```
StongeEtienne, line 122: Personally, I think this is not the best, fastest nor prettiest to get a mask from all indices.
"mask = labels_volume > 0" or "mask = labels_volume >= 1" should do the job.

scripts/scil_surface_create.py
arnaudbore: I think this could/should be done outside of the script.

scripts/scil_surface_create.py
arnaudbore: This has been fixed

arnaudbore:
> 1. I think --in_labels and --in_mask, can be a single volume input (instead of an option).

This is a decision we made a long time ago and we actually we emphasis on the difference between these two objects. We want our users to understand the difference.

> 2. There is an issue with the "vox2vtk", since now using the "--vox2vtk" tag keep it in voxel space, and by default it 
transform the surface to VTK world LPS space (aligned in mi-brain).

This has been fixed

scripts/scil_surface_from_labels.py
arnaudbore: Done it should the way you wanted now.

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,204 @@
...
+
+def main():
+    parser = _build_arg_parser()
+    args = parser.parse_args()
+    logging.getLogger().setLevel(logging.getLevelName(args.verbose))
+
+    assert_inputs_exist(parser, [], [args.in_labels, args.in_mask])
+    assert_outputs_exist(parser, args, args.out_surface)
+
+    assert args.out_surface.endswith('.vtk')
```
frheault, line 123: MI-Brain supports obj, ply, vtk (and others), so we shouldn't limit ourselves to vtk (and saving's code supports it all).

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,204 @@
...
+                               'of iterations. [%(default)s]')
+    morpho_g.add_argument('--closing',
+                          type=ranged_type(int, 0, None, min_excluded=False),
+                          help='Closing (erosion of the dilation): number '
+                               'of iterations. [%(default)s]')
+
+    p.add_argument('--fill', action='store_true',
+                   help='Fill holes in the image. [%(default)s]')
+
+    p.add_argument('--not_vox2vtk', action='store_true',
```
frheault, line 106: Why not vtk2vox?

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,204 @@
...
+                                                 area_weighted=False,
+                                                 backward_step=True)
+            mesh.set_vertices(new_vertices)
+
+        if len(masks) == 1:
+            vtk_u.save_polydata(mesh.get_polydata(),
+                                args.out_surface,
+                                legacy_vtk_format=True)
+        else:
+            out_name = args.out_surface.replace('.vtk',
```
frheault, line 196: Related to the other comment about format/extension, pick the extension of the out_surface and replace that instead of limiting it to vtk

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,204 @@
...
+
+    g1 = p.add_argument_group("Input (Labels or Mask)")
+    mxg = g1.add_mutually_exclusive_group(required=True)
+    mxg.add_argument('--in_labels',
+                     help='Path of the atlas (nii or nii.gz).\n'
+                          'You can provide a list of indices with '
+                          '--list_indices or create a surface per '
+                          'index with --each_index.\n'
+                          'If no indices are provided, '
+                          'it will use all indices.')
```
frheault, line 57: maybe replace use by merge? Or say the labels will be converted to a binary mask

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,204 @@
...
+
+"""
+Script to create a surface with marching cube from a mask or a label image.
+The surface will be readable with software like MI-Brain.
+
+Example : use wmparc.a2009s.nii.gz with some aseg.stats indices
+
+scil_surface_create.py out_surface.vtk \\
+    --in_labels s1a1/mask/S1-A1_wmparc.a2009s.nii.gz\\
+    --list_indices 16:32 --vox2vtk --opening 2 --smooth 2 -v
```
frheault, line 12: vox2vtk was changed (or remove since it is custom usage)

frheault (approved):

scripts/scil_surface_create.py
arnaudbore: Fixed

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,205 @@
...
+    g2 = p.add_argument_group("Options for labels input")
+    mxg2 = g2.add_mutually_exclusive_group()
+    mxg2.add_argument('--list_indices', nargs='+',
+                      help='List of labels indices to use for the surface.')
+    mxg2.add_argument('--each_index', action='store_true',
+                      help='Create a surface per index. It will use the '
+                           'out_surface basename to create the output files.')
+
+    g3 = p.add_argument_group('Options for volume input')
+    g3.add_argument('--value', default=0.5,
```
StongeEtienne, line 77: Since this value is force to 0.5 later on, I suggest to remove the args.value = 0.5 at line  # 129, and # 148
otherwise, remove this option and use a "global constant"

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,205 @@
...
+    logging.getLogger().setLevel(logging.getLevelName(args.verbose))
+
+    assert_inputs_exist(parser, [], [args.in_labels, args.in_mask])
+    assert_outputs_exist(parser, args, args.out_surface)
+
+    masks = []
+
+    if args.in_labels:
+        # Default value for isosurface
+        args.value = 0.5
```
StongeEtienne, line 129: see previous comment, otherwise  # 129, and # 148 could be grouped before the if

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,205 @@
...
+            indices = np.unique(labels_volume)[1:]
+            for index in indices:
+                masks.append(merge_labels_into_mask(labels_volume, str(index)))
+        else:
+            logging.warning('No indices provided, '
+                            'it will use all indices.')
+            masks.append(labels_volume > 0)
+    elif args.in_mask:
+        # Default value for isosurface
+        args.value = 0.5
```
StongeEtienne, line 148: see previous comment, otherwise  # 129, and # 148 could be grouped before the if

scripts/scil_surface_create.py
```diff
@@ -0,0 +1,205 @@
...
+    morpho_g.add_argument('--closing',
+                          type=ranged_type(int, 0, None, min_excluded=False),
+                          help='Closing (erosion of the dilation): number '
+                               'of iterations. [%(default)s]')
+
+    p.add_argument('--fill', action='store_true',
+                   help='Fill holes in the image. [%(default)s]')
+
+    p.add_argument('--vtk2vox', action='store_true',
+                   help='Do not transform to vox2vtk. [%(default)s]')
```
StongeEtienne, line 109: maybe use simply "keep output surface in voxel space"

StongeEtienne (approved):

Issue #975, opened 2024-04-17
Keyword-only arguments


CHrlS98:
Keyword-only arguments https://peps.python.org/pep-3102/ could help make the code more robust. Basically, it forces the user to write the keyword for optional keyword arguments when calling a function.

__Example__
Function declaration:
```
def function(a, b, *, c=something, d=anything)
```

Function call:
```
my_a = ...
my_b = ...
my_c = ...
my_d = ...

# valid
out = function(my_a, my_b, c=my_c, d=my_d)

# invalid
out = function(my_a, my_b, my_c, my_d)
```

arnaudbore:
Improve scil documentation about coding standard: 
https://github.com/scilus/scil-documentation

PR #1012, opened 2024-07-24
[FIX] Missing parameter for background viz

levje:
# Quick description
When running `scil_viz_fodf.py path/to/fodf.nii.gz --background path/to/bg/image.nii.gz`, I get an issue saying that 'nearest' isn't a matplotlib colormap. This is caused because the interpolation argument is given as a positional argument whilst there's a missing argument to specify the colormap.

I explicitly put the default value there (None), but I'm also opened to change it to some other value. I could also change the function call to use keyword arguments instead of positional arguments (e.g. `interpolation=args.bg_interpolation`). 

...

## Type of change

Check the relevant options.

- [ x ] Bug fix (non-breaking change which fixes an issue)

# Checklist

### DIFF
--- a/scripts/scil_viz_fodf.py
+++ b/scripts/scil_viz_fodf.py
@@ -307,12 +307,11 @@ def main():
         bg_actor = create_texture_slicer(data['bg'],
                                          args.axis_name,
                                          args.slice_index,
-                                         mask,
-                                         args.bg_range,
-                                         args.bg_opacity,
-                                         args.bg_offset,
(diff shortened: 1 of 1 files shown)


arnaudbore:
Exactly what @CHrlS98 suggested here: https://github.com/scilus/scilpy/issues/975
Good catch, please follow his instructions and do it for the other occurrence of this call in this file. 
Don't need to do it elsewhere for now :)

PR #1015, opened 2024-07-25
[ENH] Uniformize surface apply transformation

arnaudbore:
# Quick description

Quick cleaning to make it look like more scil_tractogram_apply_transform.py
...

## Type of change

Check the relevant options.

...

# Checklist


scripts/scil_surface_apply_transform.py
```diff
@@ -68,23 +71,23 @@ def main():
...
-    if args.ants_warp:
-        # Load warp
-        warp_img = nib.load(args.ants_warp)
+    deformation_data = None
+    if args.in_deformation is not None:
+        deformation_data = np.squeeze(nib.load(
+            args.in_deformation).get_fdata(dtype=np.float32))
 
-    mesh = apply_transform(mesh, affine, warp_img)
+    mesh = apply_transform(mesh, transfo, deformation_data,
```
frheault, line 89: Line 40 in that function: warp_img = np.squeeze(ants_warp.get_fdata(dtype=np.float32))

Your function expect a nib image, you give a numpy array.
AttributeError: 'numpy.ndarray' object has no attribute 'get_fdata'

frheault, line 89: In that function, if I dont use --inverse I get

UnboundLocalError: local variable 'inv_affine' referenced before assignment

scripts/scil_surface_apply_transform.py
```diff
@@ -68,23 +71,23 @@ def main():
...
-    affine = np.loadtxt(args.ants_affine)
+    # Load transformation
+    transfo = load_matrix_in_any_format(args.in_transfo)
 
-    # Non linear transformation
-    if args.ants_warp:
-        # Load warp
-        warp_img = nib.load(args.ants_warp)
+    deformation_data = None
+    if args.in_deformation is not None:
```
frheault, line 85: The affine from the original image is required by the function, so there is a problem with passing only numpy stuff

scripts/scil_surface_apply_transform.py
```diff
@@ -68,23 +71,23 @@ def main():
...
 
-    assert_inputs_exist(parser, [args.in_surface, args.ants_affine],
-                        args.ants_warp)
+    assert_inputs_exist(parser, [args.in_moving_surface, args.in_transfo],
+                        args.in_deformation)
     assert_outputs_exist(parser, args, args.out_surface)
 
     # Load mesh
-    mesh = load_mesh_from_file(args.in_surface)
+    mesh = load_mesh_from_file(args.in_moving_surface)
```
frheault, line 79: A flip xy must be done after loading to go to RAS (from LPS) and the same flip must be applied before saving to go from RAS to LPS

frheault (commented):
This can wait after your vacation, this should not be long but I can help you understand the problems (I have testing data)

scripts/scil_surface_apply_transform.py
```diff
@@ -8,7 +8,7 @@
 Example usage from T1 to b0 using ANTs transforms:
 > ConvertTransformFile 3 output0GenericAffine.mat vtk_transfo.txt --hm
 > scil_surface_apply_transform.py lh_white_lps.vtk affine.txt lh_white_b0.vtk\\
-    --ants_warp warp.nii.gz
+    --in_deformation warp.nii.gz --inverse
```
frheault, line 11: Could you use vtk_transfo.txt in the example (since it is the output filename from the line above) and use the output1InverseWarp.nii.gz file

scripts/scil_surface_apply_transform.py
```diff
@@ -68,26 +71,25 @@ def main():
...
+    if args.in_deformation is not None:
+        deformation_data = nib.load(args.in_deformation)
 
-    mesh = apply_transform(mesh, affine, warp_img)
+    mesh = apply_transform(mesh, transfo, deformation_data,
+                           inverse=args.inverse)
 
     # Save mesh
-    mesh.save(args.out_surface)
+    save_polydata(mesh.__polydata__, args.out_surface, legacy_vtk_format=True)
```
frheault, line 92: I think we decided that the only way to save in legacy was to use the convert. The default for all script was to save the new format (even if it does not work in MI-Brain).

I personally recommend using PLY instead (as long as MI-Brain does not update)

frheault (approved):
I think it is done, except two small comments

scripts/scil_surface_apply_transform.py
arnaudbore: I don't recall we did say that. Look at scil_surface_convert.py we save it so MI-Brain can read them.

PR #1020, opened 2024-08-01
Modify lesions_info to support labels instead of mask

frheault:
# Quick description

lesions_info now use labels as input because @AntoineTheb created the script scil_labels_from_mask.py. This avoids duplicated processing in the script, this also will be required in the future because I will add a few lesions-related scripts.
...

## Type of change

Check the relevant options.

- [x] This change requires a documentation update

...

# Checklist

### DIFF
--- a/scilpy/image/labels.py
+++ b/scilpy/image/labels.py
@@ -99,6 +99,7 @@ def get_labels_from_mask(mask_data, labels=None, background_label=0,
         for label in range(1, nb_structures + 1):
             if np.count_nonzero(label_map == label) < min_voxel_count:
                 label_map[label_map == label] = 0
+        mask_data = label_map > 0
         label_map, nb_structures = ndi.label(mask_data)
 
     # Assign labels to each blob if provided
--- a/scilpy/image/labels.py
+++ b/scilpy/image/labels.py
@@ -96,11 +96,17 @@ def get_labels_from_mask(mask_data, labels=None, background_label=0,
     # Get the number of structures and assign labels to each blob
     label_map, nb_structures = ndi.label(mask_data)
     if min_voxel_count:
+        new_count = 0
         for label in range(1, nb_structures + 1):
             if np.count_nonzero(label_map == label) < min_voxel_count:
                 label_map[label_map == label] = 0
-        mask_data = label_map > 0
-        label_map, nb_structures = ndi.label(mask_data)
+            else:
+                new_count += 1
+                label_map[label_map == label] = new_count
+        logging.debug(
+            f"Ignored blob {nb_structures-new_count} with fewer "
+            "than {min_voxel_count} voxels")
+        nb_structures = new_count
 
     # Assign labels to each blob if provided
     if labels:
--- a/scilpy/image/labels.py
+++ b/scilpy/image/labels.py
@@ -68,7 +68,8 @@ def get_binary_mask_from_labels(atlas, label_list):
     return mask
 
 
-def get_labels_from_mask(mask_data, labels=None, background_label=0):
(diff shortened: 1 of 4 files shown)


scilpy/image/labels.py
```diff
@@ -91,6 +95,12 @@ def get_labels_from_mask(mask_data, labels=None, background_label=0):
     """
     # Get the number of structures and assign labels to each blob
     label_map, nb_structures = ndi.label(mask_data)
+    if min_voxel_count:
+        for label in range(1, nb_structures + 1):
+            if np.count_nonzero(label_map == label) < min_voxel_count:
+                label_map[label_map == label] = 0
+        label_map, nb_structures = ndi.label(mask_data)
```
arnaudbore, line 102: I think this should not be there.

arnaudbore (approved):
Apart from the extra call, I think we're good.

scripts/scil_lesions_info.py
```diff
@@ -3,11 +3,15 @@
 
 """
 This script will output informations about lesion load in bundle(s).
-The input can either be streamlines, binary bundle map, or a bundle voxel
+The input can either be streamlines, binary bundle mask, or a bundle voxel
 label map.
 
+The input lesion file is a labeled volume (.nii.gz) where each lesion is
```
AntoineTheb, line 9: Confusing documentation with respect to the previous line.

AntoineTheb (changes requested):
Small comment besides what Arnaud pointed out

scilpy/image/labels.py
frheault: It relabels lesions after discarding the small one, this is to avoid lesions in the thousands despite having maybe a few hundred 'valid lesions'.

This is inside the `if` so it is done only if there is a threshold, I could also add a True/False in the for loop to having recomputing if there was no 'invalid lesions'

EmmaRenauld: Which is faster / clearer? Your version, or:
```
if min_voxel_count:
        new_count = 0
        for label in range(1, nb_structures + 1):
            if np.count_nonzero(label_map == label) < min_voxel_count:
                label_map[label_map == label] = 0
            else:
                  new_count +=1
                  label_map[label_map == label] = new_count
```

EmmaRenauld: I would also add a :
`logging.debug("Ignored blob with x voxels")` in the loop.

scripts/scil_lesions_info.py
```diff
@@ -3,11 +3,15 @@
...
+represented by a unique label. To label a lesion file use
+scil_labels_from_mask.py
+
+Then, the second input can either be streamlines, binary bundle mask, or a
+bundle voxel label map.
 
 To be considered a valid lesion, the lesion volume must be at least
-min_lesion_vol mm3. This avoid the detection of thousand of single voxel
+--min_lesion_vol mm3. This avoid the detection of thousand of single voxel
```
EmmaRenauld, line 14: This avoids.
Thousands takes an s. But I would not use a number. This avoids the detection of a too high number of lesions.

scripts/scil_labels_from_mask.py
```diff
@@ -34,6 +34,10 @@ def _build_arg_parser():
     p.add_argument('--background_label', default=0, type=int,
                    help='Label to assign to the background. [%(default)s]')
 
+    p.add_argument('--min_volume', type=float, default=7,
+                   help='Minimum volume in mm3 [%(default)s],'
+                        'Useful for lesions.')
+
```
EmmaRenauld, line 40: Please add this option in the script test.

EmmaRenauld (approved):
Small modifs, then good to go, I think.

PR #1047, opened 2024-11-05
ENH: better bundle endpoints maps

AntoineTheb:
# Quick description
 
This PR improves bundle endpoint maps computation. Before, the map was only computed on streamline points exactly, which may be problematic if the step size is too large (see fig. 1 for ref). Now, the map includes every voxel traversed by the endpoints (however many points are selected). Moreover, the user can now select `n` millimeters instead of points, which makes much more sense when dealing with compressed streamlines.

TODO: Fix the breaking test.

## Type of change

Check the relevant options.

- [X] New feature (non-breaking change which adds functionality)

image
Fig 1.: Before, 3 points

image
After: After, 10mm.

# Checklist

- [X] I added relevant citations to scripts, modules and functions docstrings and descriptions
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I moved all functions from the script file (except the argparser and main) to scilpy modules

### DIFF
--- a/scilpy/tractograms/streamline_and_mask_operations.py
+++ b/scilpy/tractograms/streamline_and_mask_operations.py
@@ -77,31 +77,23 @@ def get_head_tail_density_maps(sft, point_to_select=1):
     dimensions = sft.dimensions
     streamlines = sft.streamlines
 
+    list_indices, points_to_indices = uncompress(sft.streamlines, return_mapping=True)
+
     endpoints_map_head = np.zeros(dimensions)
     endpoints_map_tail = np.zeros(dimensions)
 
     # A possible optimization would be to compute all coordinates first
     # and then do the np.add.at only once.
-    for streamline in streamlines:
-
-        # Resample the streamline to make sure we have enough points
-        nb_point = max(len(streamline), point_to_select*2)
-        streamline = set_number_of_points(streamline, nb_point)
+    for indices, points in zip(list_indices, points_to_indices):
 
         # Get the head and tail coordinates
-        points_list_head = streamline[0:point_to_select, :]
-        points_list_tail = streamline[-point_to_select:, :]
-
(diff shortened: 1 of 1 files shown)


scilpy/tractograms/streamline_and_mask_operations.py
```diff
@@ -59,8 +64,11 @@ def get_head_tail_density_maps(sft, point_to_select=1):
         The streamlines to compute endpoints density from.
     point_to_select: int
         Instead of computing the density based on the first and last points,
-        select more than one at each end. To support compressed streamlines,
-        a resampling to 0.5mm per segment is performed.
```
AntoineTheb, line 63: This actually wasn't performed.

scripts/scil_bundle_compute_endpoints_map.py
```diff
@@ -53,6 +53,9 @@ def _build_arg_parser():
     p.add_argument('--nb_points', type=int, default=1,
                    help="Number of points to consider at the extremities"
                         " of the streamlines. [%(default)s]")
+    p.add_argument('--mm', action='store_true',
+                   help='Compute the endpoints in mm instead of nb. of '
+                        'points. Useful for compressed streamlines.')
```
EmmaRenauld, line 58: I think this could be a little clearer. Like, if user adds `--mm`, but not --nb_points`, is it clear to them that it will take 1mm?

Maybe.... `If set, the value --nb_points will instead be understood as the length, in mm`. But I'm not sure that's perfect either.

EmmaRenauld (approved):

scripts/scil_bundle_compute_endpoints_map.py
AntoineTheb: How about now ?

PR #1063, opened 2024-11-29
Next connectivity scripts

EmmaRenauld:
Cleaning:

- scil_connectivity_compare_populations: already clean! simply trying to clarify the help!
- scil_connectivity_normalize: already clean! simply trying to clarify the help!
- scil_connectivity_reorder_rois: already nearly clean! simply trying to clarify the help! Some code was written twice; simplified.

- scil_connectivity_compute_pca: This script was not yet written with scilpy's standard. Bigger cleaning.
     - simplified loops in cases where numpy can do it in one call
     - simplifed mask usage
     - separated the functions for plotting just for an easier read.


frheault (commented):
Everything looks alright, but I asked @gagnonanthony  to review <in-depth> the PCA script since he wrote it (and maybe even re-test it himself). I would wait for his OK

scripts/scil_connectivity_compute_pca.py
```diff
@@ -150,179 +192,117 @@ def autolabel(rects, axs):
...
-
-        # Creating input structure.
         logging.info('Creating PCA input structure with common edges...')
-        df = generate_pca_input(dictionary)
 
-    # Setting 0 values to nan.
-    df[df == 0] = 'nan'
-
-    # Standardized the data.
+        common_edges_mask = np.sum(matrices_per_metric[0]) != 0
```
gagnonanthony, line 239: I'm unsure about this; the common edges are supposed to be across all input subjects. Using the `--all_edges` flag returns the same results as if we don't use it.

gagnonanthony (commented):
Thanks for reformatting the `scil_connectivity_compute_pca` script! Clearly there was some work to be done. I've tested it quickly and compared it to the previous version, and it seems that the typical edges selection is not working correctly. Using the `all_edges` or not returns the same results.

EmmaRenauld:
Sorry! Should be fixed

scripts/scil_connectivity_compute_pca.py
```diff
@@ -150,179 +191,128 @@ def autolabel(rects, axs):
...
+            tmp_common = np.sum(tmp_subj_masks, axis=0) == nb_subjects
+            if not np.array_equal(common_edges_mask, tmp_common):
+                parser.error("Different binary masks (common edge) have been "
+                             "generated from metric {} than other metrics. "
+                             "Please validate your input data."
+                             .format(args.metrics[i]))
+
+        plt.figure()
+        plt.imshow(common_edges_mask)
+        plt.title(common_edges_mask)
```
gagnonanthony, line 255: This is returning an error (I guess it's a typo!). I like the addition of the matrix visualization of the common connection by the way, great addition!

gagnonanthony (commented):
I confirm the results are now identical to before! Thanks @EmmaRenauld

arnaudbore (approved):
GTG

PR #1066, opened 2024-12-02
Script dps math

VincentBeaud:
# Quick description

Replace `scil_tractogram_add_dps.py` for a more general version `scil_tractogram_dps_math.py`. This new script allows "add", "delete" and "export" operations on data-per-streamline.

## Type of change

Check the relevant options.

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Example: I want to extract the seeds from `tractogram.trk`. I use the following command line.
`scil_tractogram_dps_math tractogram.trk export --export_dps_key seeds --export_dps_file seeds.txt`

# Checklist


CHrlS98 (approved):
Looks great, thanks!

CHrlS98:
@arnaudbore PR looks good to me, can you have a look?

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Add, extract or delete information to each streamline from a tractogram
```
mdesco, line 5: Add, extract,  or delete dps (data_per_streamline) information to ....

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,124 @@
...
+file. Can be for example SIFT2 weights, processing information, bundle IDs,
+tracking seeds, etc.
+
+Input and output tractograms must always be .trk to simplify filetype checks
+for each operation.
```
mdesco, line 10: The real reason is that tck format do not support dps, no? Maybe delete the end of the sentence "to simplify ... "

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,124 @@
...
+"""
```
mdesco, line 11: Before the end of the help, can you provide an example command line for this script. I'm new to this and I don't know how I would call it. Not sure what a dps_key is ... One or two examples would help. Thanks

mdesco (commented):
A few minor comments

scripts/scil_tractogram_dps_math.py
VincentBeaud: It's because the original script could take .tck as input and add dps to form a .trk. I don't do that because it's too complicated with all the different operations.

I'll make it clearer. :)

CHrlS98: If it's the same script I don't think we should remove the option for .tck, it could currently be used by someone in some pipeline somewhere and that change could break those pipelines.

CHrlS98: If the operation is not `add`, a .tck input should always raise an error. However, for saving I think it's ok to always save .trk.

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+    p.add_argument('operation', metavar='OPERATION',
+                   choices=['add', 'delete', 'export'],
+                   help='The type of operation to be performed on the\n'
+                        'tractogram\'s data_per_streamline at the given\n'
+                        'key. Must be one of the following: [%(choices)s].\n'
+                        'The additional arguments required for each\n'
+                        'operation are specified under each group below.')
+    p.add_argument('dps_key', type=str,
+                   help='Where to find the data to be exported,\n'
+                        'in the tractogram.')
```
arnaudbore, line 51: change for something like this: "Key name used for the operation."

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+def _build_arg_parser():
+
+    p = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawTextHelpFormatter)
+
+    p.add_argument('in_tractogram',
+                   help='Input tractogram (.trk).')
+    p.add_argument('operation', metavar='OPERATION',
+                   choices=['add', 'delete', 'export'],
```
arnaudbore, line 43: i would use import, export, delete

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+                        'The additional arguments required for each\n'
+                        'operation are specified under each group below.')
+    p.add_argument('dps_key', type=str,
+                   help='Where to find the data to be exported,\n'
+                        'in the tractogram.')
+
+    p.add_argument('--out_tractogram',
+                   help='Output tractogram (.trk). Required for any mutation.')
+
+    add_args = p.add_argument_group('Add operation',
```
arnaudbore, line 56: Operation \"import\" - mandatory options

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+                                    'Requires the out_tractogram argument.')
```
arnaudbore, line 57: you can delete this

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+    add_args.add_argument('--add_dps_file',
+                          help='File containing the data to add to\n'
+                               'streamlines (.txt, .npy or .mat).')
+
+    _ = p.add_argument_group('Delete operation',
+                             'Requires the out_tractogram argument.')
+
+    export_args = p.add_argument_group('Export operation')
```
arnaudbore, line 65: Operation \"export\" - mandatory options

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+                        'tractogram\'s data_per_streamline at the given\n'
+                        'key. Must be one of the following: [%(choices)s].\n'
+                        'The additional arguments required for each\n'
+                        'operation are specified under each group below.')
+    p.add_argument('dps_key', type=str,
+                   help='Where to find the data to be exported,\n'
+                        'in the tractogram.')
+
+    p.add_argument('--out_tractogram',
+                   help='Output tractogram (.trk). Required for any mutation.')
```
arnaudbore, line 54: Mandatory for \"import\" and \"delete\" operations.

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+
+    add_args = p.add_argument_group('Add operation',
+                                    'Requires the out_tractogram argument.')
+    add_args.add_argument('--add_dps_file',
```
arnaudbore, line 58: use in_dps_file and out_dps_file

scripts/scil_tractogram_dps_math.py
```diff
@@ -0,0 +1,129 @@
...
+        sft.data_per_streamline[args.dps_key] = data
+
+        save_tractogram(sft, args.out_tractogram)
+
+    if args.operation == 'delete':
+        del sft.data_per_streamline[args.dps_key]
+
+        save_tractogram(sft, args.out_tractogram)
+
+    if args.operation == 'export':
```
arnaudbore, line 117: you need to add some conditions to make sure export_dps_file has been provided or put a default output dps filename.

arnaudbore (changes requested):
a couple of things to fix before moving forward.

scripts/scil_tractogram_dps_math.py
VincentBeaud: @CHrlS98 It would break any pipeline using it either way, it's not the same script and it has different options. But alright I don't think it's a bad idea to leave .tck support for add.

CHrlS98: Ah oui c'est vrai c'est même plus le même script. Do as you wish, if it's too complicated to support .tck you can let it go.

VincentBeaud:
@arnaudbore Ok c'est terminé. Au fait, j'ai corrigé un mini truc de doc dans scilpy.io.streamlines. Tu iras voir.

frheault:
I just saw that scil_tractogram_dpp_math.py has a --mode dps or dpp option. I think it would be important to harmonize both script (like remove the dps mode from scil_tractogram_dpp_math.py) and make sure options are mirrored in both. Can you look into that? (I can help)

PR #1084, opened 2024-12-12
Condense distance matrix that was not. Pep8 + test pass

### LINKED ISSUE #1060: Todo: Fix scipy warning
image

VincentBeaud:
# Quick description

Closes #1060 

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/scilpy/connectivity/matrix_tools.py
+++ b/scilpy/connectivity/matrix_tools.py
@@ -7,7 +7,6 @@ import bct
 from dipy.utils.optpkg import optional_package
 import numpy as np
 from scipy.cluster import hierarchy
-from scipy.spatial.distance import squareform
 
 from scilpy.image.labels import get_data_as_labels
 from scilpy.stats.matrix_stats import omega_sigma
@@ -37,11 +36,9 @@ def compute_olo(array):
     if array.ndim != 2:
         raise ValueError('RCM can only be applied to 2D array.')
 
-    condensed_array = squareform(array)
-
-    Z = hierarchy.ward(condensed_array)
+    Z = hierarchy.ward(array)
     perm = hierarchy.leaves_list(
-        hierarchy.optimal_leaf_ordering(Z, condensed_array))
+        hierarchy.optimal_leaf_ordering(Z, array))
 
     return perm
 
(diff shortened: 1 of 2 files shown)


VincentBeaud:
Passes test, but no validation that it doesn't change the result. It shouldn't.. but maybe have @frheault validate the fix since I think he's more familiar with spatial manipulations
@arnaudbore

scilpy/connectivity/matrix_tools.py
```diff
@@ -36,9 +37,11 @@ def compute_olo(array):
     if array.ndim != 2:
         raise ValueError('RCM can only be applied to 2D array.')
 
-    Z = hierarchy.ward(array)
+    condensed_array = squareform(array)
```
frheault, line 40: I don't understand why squareform would help here? A connectivity array (like streamline count) is already a form of distance matrix. I think it is fine as it was, no?

What was your goal when you found and added squareform?

VincentBeaud: When trying to solve this issue, I learned that a distance matrix is symmetric and can be condensed into a a list or vector. Similarly to a graph adjacency matrix that you transform into an adjacency list.

The error was that the matrix looked "suspiciously like an uncondensed distance matrix" when used in the optimal_leaf_ordering function. Using a symmetric (redundant) distance matrix in scipy functions gives a warning because it's suboptimal. (I suppose)

Squareform converts a uncondensed distance matrix into a condensed one and vice versa. The error disapeared after that.

frheault: Ok but our matrices is real data, we can't change it to fix an error. Did you verify if the input/output still made sense?

Otherwise I would leave it like that.

VincentBeaud: I'm not sure what you mean by real data. Maybe we should walk about it in person. 😄

VincentBeaud:
Okay after discussion, squareform seems to result in the wrong permutations because there is data on the diagonal of the matrix that needs to be taken into account. Works with test data, but not with real data.

The warning was already set to be muted in matrix_tools.py, but somehow it didn't work with the tests. So I added another line in the test to mute the same warning. Now we don't see it.

VincentBeaud:
Adding the simplefilter at the top of the test script does not seem to work. This is why I put it inside the test case.

PR #1091, opened 2024-12-12
Update tests - parallele

arnaudbore:
# Quick description

Parallel tests 
...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Run tests
         run: |
           export C_INCLUDE_PATH=$pythonLocation/include/python${{ steps.python-selector.outputs.python-version }}:$C_INCLUDE_PATH
-          pytest --cov-report term-missing:skip-covered
+          pytest -n logical --cov-report term-missing:skip-covered
 
       - name: Save test results and coverage
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Run tests
         run: |
           export C_INCLUDE_PATH=$pythonLocation/include/python${{ steps.python-selector.outputs.python-version }}:$C_INCLUDE_PATH
-          pytest --cov-report term-missing:skip-covered
+          pytest -n logical --cov-report term-missing:skip-covered
```
AlexVCaron, line 59: Do we want to always activate pytest-xdist for all users when they run tests ? If yes, I would put that in `addopts` in the  `pytest.ini` config file. Also, you would need to add `pytest-xdist` and `pytest-xdist[psutil]` to the `required_plugins` list.

requirements.txt
```diff
@@ -35,6 +35,7 @@ pybids==0.16.*
 pyparsing==3.0.*
 PySocks==1.7.*
 pytest==7.2.*
+pytest-xdist==3.6.*
```
AlexVCaron, line 38:
```suggestion
pytest-xdist==3.6.*
pytest-xdist[psutil]==3.6.*
```

.github/workflows/test.yml
AlexVCaron: Add this : `--dist worksteal`. it makes the test worker steal job from each other when they finnish early.

AlexVCaron (changes requested):
A few comments to make this even better, then LGTM !

PR #1102, opened 2024-12-12
Unit tests connectivity + FIX MAP_COORDINATES USAGE

EmmaRenauld:
Adding unit tests for connectivity methods

- Deleted method `apply_olo`. Not used anywhere inside scilpy.
- Fixed `compute_triu_connectivity_from_labels`:
    - Was putting the streamline count at the row / column of the matrix for the label's id, not the index of that label in the matrix.
    - Was not using correctly map_coordinates.
 
Led to correcting map_coordinates a bit everywhere. **Explanation:**

Without `mode='nearest'`  (meaning, pad to nearest value when out of range), the default is to pad to 0 when outside the volume. The issue is that the first half of the voxels at the limit is considered out of range, as per following picture (thanks @AntoineTheb ): 

image

### DIFF
--- a/scilpy/connectivity/connectivity.py
+++ b/scilpy/connectivity/connectivity.py
@@ -117,7 +117,7 @@ def compute_triu_connectivity_from_labels(tractogram, data_labels,
     return matrix, ordered_labels, start_labels, end_labels
 
 
-def load_node_nifti(directory, in_label, out_label, ref_img):
+def _load_node_nifti(directory, in_label, out_label, ref_img):
     in_filename = os.path.join(directory,
                                '{}_{}.nii.gz'.format(in_label, out_label))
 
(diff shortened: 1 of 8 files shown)


EmmaRenauld:
@frheault should I also change it here? I was less sure: (in `transform_warp_sft`):
image

and in `surface_operations.apply_transform`:
image

AntoineTheb (approved):
LGTM

*however*

The one issue I have with using `map_coordinates` with `mode=nearest` is if we are mapping a point that is actually out of the volume. Before this PR, the returned value would be `cval`, the default out-of-bounds value (e.g .0), which can be "caught" and dealt with. Now, the value would be the nearest point (`map_coordinates` deals with points, not voxels) and we would not know if we are actually out of bounds. 

In an ideal world, we would use dipy's interpolation function, but it's kind of unwieldy. An other solution would be to have our own interpolation function (maybe a wrapper around `map_coordinates` that checks volume bounds). 

There are a few cases where this would be problematic, like in the brainstem or the cerebellum if the FOV is really tight, but as it is it should be fine for most cases.

EmmaRenauld:
I agree. We usually test our streamlines for invalid streamlines when it is important be we could do some checking. I will add an issue and we can see about it later.

arnaudbore (approved):

Issue #891, opened 2024-02-06
SFT's hidden `_data` should never be used.


EmmaRenauld:
As discovered with François, the sft.data_per_point[key]._data is not always updated when shuffling or modifying the SFT.

This is probably true for all other usages of hidden _data. 

I cleaned one line of code (see #890) , but code should be verified everyhwere.

@frheault

arnaudbore:
is it only dpp and dps ? or every single one like sft.streamlines._data ?

frheault:
It is true of that one too, luckily Scilpy scripts are usually doing the right thing, but this is very dangerous when modifying a script, developing a new feature or hacking away in IPython.

The array sequence use something similar to https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html

This means that if you slice an array sequence you are playing with a view and the underlying data is untouched

PR #890, opened 2024-02-02
Robustify the coloring

EmmaRenauld:
# Quick description

In some cases, the hidden _data is not ordered correctly. Ex:

```
new_order = np.flip(np.arange(len(sft)))
sft = sft[new_order]
```

The sft.data_per_point[key]._data is not always updated!

Accessing the data directly instead.

! **Note: There are many other _data usage in that file. I didn't test them all.**

It also made me notice that the test name was not correct.
...

## Type of change

Check the relevant options.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Provide data, screenshots, command line to test (if relevant)

...

# Checklist

- [x] My code follows the style guidelines of this project (run [autopep8](https://pypi.org/project/autopep8/))
- [x] I added relevant citations to scripts, modules and functions docstrings and descriptions
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I moved all functions from the script file (except the argparser and main) to scilpy modules
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1105, opened 2024-12-12
ENH: remove `sft._data` usage part 1

AntoineTheb:
# Quick description

As per #891, although does not fix it completely.

## Type of change

Check the relevant options.

- [X] Bug fix (non-breaking change which fixes an issue)

No

# Checklist

### DIFF
--- a/scilpy/tractograms/dps_and_dpp_management.py
+++ b/scilpy/tractograms/dps_and_dpp_management.py
@@ -1,12 +1,50 @@
 # -*- coding: utf-8 -*-
 import numpy as np
 
-from scilpy.viz.color import clip_and_normalize_data_for_cmap
+from nibabel.streamlines import ArraySequence
 
 
-def add_data_as_color_dpp(sft, cmap, data, clip_outliers=False, min_range=None,
-                          max_range=None, min_cmap=None, max_cmap=None,
-                          log=False, LUT=None):
+def get_data_as_arraysequence(data, ref_sft):
+    """ Get data in the same shape as a reference StatefulTractogram's
+    streamlines, so it can be used to set data_per_point or
+    data_per_streamline. The data may represent one value per streamline or one
+    value per point. The function will return an ArraySequence with the same
+    shape as the streamlines.
+
+    Parameters
+    ----------
+    data: np.ndarray
+        The data to convert to ArraySequence.
+    ref_sft: StatefulTractogram
+        The reference StatefulTractogram containing the streamlines.
+
+    Returns
+    -------
+    data_as_arraysequence: ArraySequence
+        The data as an ArraySequence.
+    """
+
+    if data.shape[0] == len(ref_sft):
+        data_as_arraysequence = ArraySequence(data)
+    elif data.shape[0] == ref_sft._get_point_count():
+        data_as_arraysequence = ArraySequence()
+        # This function was created to avoid messing with _data, _offsets and
+        # _lengths, so this feel kind of bad. However, the other way would be
+        # to create a new ArraySequence and iterate over the streamlines, but
+        # that would be way slower.
+        data_as_arraysequence._data = data
+        data_as_arraysequence._offsets = ref_sft.streamlines._offsets
+        data_as_arraysequence._lengths = ref_sft.streamlines._lengths
+    else:
+        raise ValueError("Data has the wrong shape. Expecting either one value"
+                         " per streamline ({}) or one per point ({}) but got "
+                         "{}."
+                         .format(len(ref_sft), len(ref_sft.streamlines._data),
+                                 data.shape[0]))
+    return data_as_arraysequence
+
+
+def add_data_as_color_dpp(sft, color):
     """
     Normalizes data between 0 and 1 for an easier management with colormaps.
     The real lower bound and upperbound are returned.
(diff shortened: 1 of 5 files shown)


scilpy/tractograms/dps_and_dpp_management.py
```diff
@@ -1,12 +1,50 @@
...
+
+    if data.shape[0] == len(ref_sft):
+        data_as_arraysequence = ArraySequence(data)
+    elif data.shape[0] == ref_sft._get_point_count():
+        data_as_arraysequence = ArraySequence()
+        # This function was created to avoid messing with _data, _offsets and
+        # _lengths, so this feel kind of bad. However, the other way would be
+        # to create a new ArraySequence and iterate over the streamlines, but
+        # that would be way slower.
+        data_as_arraysequence._data = data
```
EmmaRenauld, line 35: Isn't there a way to hstack the data, or vstack, and then convert to array sequence?

In my PR #890 , I had done this. Is it really that much slower?

image

frheault: Something I often do is that I create an array sequence with my streamlines like this:
dpp = sft.streamlines.copy()
dpp._data = my_array.copy()

That way everything is initialized with the right length, memory safe.

AntoineTheb: @EmmaRenauld I feel like iterating over all streamlines would be problematic for large (>1M) tractograms. 
@frheault the problem with your approach is the dpp is still "prealocated" using the streamlines, which is weird. I think this is a deeper problem stemming from the way ArraySequences work but I'd rather not have to preallocate data per points as the streamlines' points.  Especially if more complicated processing is needed and some DPP may be left to the streamline's value by accident.

AntoineTheb: @EmmaRenauld @frheault should be a bit cleaner now.

scilpy/viz/color.py
```diff
@@ -283,10 +283,10 @@ def prepare_colorbar_figure(cmap, lbound, ubound, nb_values=255, nb_ticks=10,
     return fig
 
 
-def ambiant_occlusion(sft, colors, factor=4):
+def ambient_occlusion(sft, colors, factor=4):
     """
     Apply ambiant occlusion to a set of colors based on point density
-    around each points. 
+    around each points.
```
EmmaRenauld, line 289: each point (no s).

scilpy/tractograms/tests/test_dps_and_dpp_management.py
```diff
@@ -27,45 +30,112 @@ def _get_small_sft():
...
     assert len(colored_sft.data_per_streamline.keys()) == 0
     assert list(colored_sft.data_per_point.keys()) == ['color']
-    assert lbound == 4
-    assert ubound == 5
-    # Lowest cmap color should be first point of second streamline.
-    # Same value for all points.
-    colors_first_line = colored_sft.data_per_point['color'][0]
-    assert np.array_equal(colors_first_line[0, :], lowest_color)
-    assert np.all(colors_first_line[1:, :] == colors_first_line[0, :])
+
```
EmmaRenauld, line 109: Same here.

scilpy/tractograms/tests/test_dps_and_dpp_management.py
```diff
@@ -27,45 +30,112 @@ def _get_small_sft():
...
-    highest_color = np.asarray(lut(1.0)[0:3]) * 255
+def test_add_data_as_dpp_1_per_point():
 
     fake_sft = _get_small_sft()
+    cmap = get_lookup_table('jet')
 
     # Not testing the clipping options. Will be tested through viz.utils tests
 
     # Test 1: One value per point.
     # Lowest cmap color should be first point of second streamline.
```
EmmaRenauld, line 82: Can you remove this sentence (`Lowest cmap color should be first point of second streamline.`). It sound like we will test the values, but we don't anymore, because the method "add_data_as_color_dpp" does not do any clipping / changing the range anymore.

It is possible to add these tests that I had prepared into the tests of clip_and_normalize_data_for_cmap?

scilpy/tractograms/tests/test_dps_and_dpp_management.py
```diff
@@ -27,45 +30,112 @@ def _get_small_sft():
...
-                          highest_color)
+
+
+def test_add_data_as_dpp_1_per_streamline():
+
+    fake_sft = _get_small_sft()
+    cmap = get_lookup_table('jet')
 
     # Test 2: One value per streamline
     # Lowest cmap color should be every point in first streamline
```
EmmaRenauld, line 99: This sentence too can be removed. Not tested anymore.

EmmaRenauld (approved):
Looks good thanks.

AntoineTheb:
@arnaudbore

EmmaRenauld:
> I also found a bug in streamline_operations.py at line 286

We already have tests for it: https://github.com/scilus/scilpy/blob/master/scilpy/tractograms/tests/test_streamline_operations.py,
created in PR #1027. Can you explain the bug, so that I can add more tests if they are not ok?

AntoineTheb:
@frheault can you provide the test data you used ?

frheault:
> > I also found a bug in streamline_operations.py at line 286
> 
> We already have tests for it: https://github.com/scilus/scilpy/blob/master/scilpy/tractograms/tests/test_streamline_operations.py, created in PR #1027. Can you explain the bug, so that I can add more tests if they are not ok?

On the data I sent Antoine, the --cut was not working because the DPP was one item short of the streamlines points data and the SFT was not liking it.

(scil_tractogram_remove_invalid.py on IFOF big with the --cut option)
[test_pr_1105.zip](https://github.com/user-attachments/files/18608970/test_pr_1105.zip)

AntoineTheb:
@frheault the file IFOF_big in your data was actually making `scil_tractogram_remove_invalid` crash as there was a bug in the script. I have fixed it. Were you experiencing this crash ? I could not find any problem with the dpp of IFOF_big though ...

scilpy/tractograms/streamline_operations.py
```diff
@@ -258,34 +258,34 @@ def cut_invalid_streamlines(sft, epsilon=0.001):
...
-                    sft.streamlines[ind][best_pos[0]:best_pos[1]])
-                cutting_counter += 1
+
+            in_vol = np.logical_and(
+                sft.streamlines[ind] >= epsilon,
+                sft.streamlines[ind] < sft.dimensions - epsilon).all(axis=1)
+
+            # Get segments in the streamline that are within the volume using
+            # ndi.label
+            blobs, _ = ndi.label(in_vol)
```
AntoineTheb, line 268: @frheault @arnaudbore I have greatly simplified and sped up the invalid-cutting process. Let me know how you feel about it.

PR #1113, opened 2024-12-18
Fixing the skip_b0_check function

### LINKED ISSUE #1108: Argument skip_b0_check always overwrites with minimum
I think there might be a major bug with how we deal with the b0 checks. A bit of context:

- The argument `b0_threshold` sets a maximum value for the b0s.
- The fonction `check_b0_threshold` in `scilpy.gradients.bvec_bval_tools` then compares the `b0_threshold` with the minimal b-value `min_bval`.
- By default, if the `min_bval` is higher than `b0_threshold`, it will raise an Error.
- The `skip_b0_check` allows us to raise a warning instead and overwrite `b0_threshold` to be the same as `min_bval`.
- However, there is no way to raise the warning but keep the `b0_threshold` as is.
- The `add_skip_b0_check_arg` function has a `will_overwrite_with_min` argument that seems completely useless and even misleading, because it only affects the help message but not the behavior of `skip_b0_check`. For instance, I tought that `will_overwrite_with_min=False` would keep the `b0_threshold` as is. This is important in cases where you have no b0s and don't want to take the lowest shell as a b0.

My question is: What is the use of this `will_overwrite_with_min` argument, which is even mandatory. Am I missing something?

I think we should find a way to pass the bool value of `will_overwrite_with_min` to the `check_b0_threshold` function, or split the function in two (skip_b0_check and overwrite_b0_thr_with_min).

karanphil:
# Quick description

See issue #1108 for more details. In summary, I realized that the `skip_b0_check` option forces the `b0_threshold` to become the minimal b-value, which doesn't work in some cases. Many scripts used the `check_b0_threshold` function without saving the outputted threshold, because we did not want to overwrite the `b0_threshold`. However, the warning message from `check_b0_threshold` with  `skip_b0_check` still said that the `b0_threshold` was changed to the minimal b-value, which was false in this case. 

A simple fix was to add a `overwrite_with_min` option to `check_b0_threshold`, so that if it is False, the warning says that the script will continue without b0s, as it does. I added this `overwrite_with_min=False` option to all scripts that had `will_overwrite_with_min=False` when calling `add_skip_b0_check_arg`. I also changed the help when this `will_overwrite_with_min` option is False (it was not doing what was said).

Closes #1108 

...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/scilpy/gradients/bvec_bval_tools.py
+++ b/scilpy/gradients/bvec_bval_tools.py
@@ -55,7 +55,8 @@ def normalize_bvecs(bvecs):
     return bvecs
 
 
-def check_b0_threshold(min_bval, b0_thr, skip_b0_check):
+def check_b0_threshold(min_bval, b0_thr, skip_b0_check,
+                       overwrite_with_min=True):
     """
     Check if the minimal bvalue is under the threshold. If not, raise an
     error to ask user to update the b0_thr.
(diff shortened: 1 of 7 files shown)


arnaudbore (approved):
LGTM

EmmaRenauld (approved):

PR #1133, opened 2025-01-28
Add test for volume_metrics

EmmaRenauld:
Adding some unit tests for file volume_metrics: piesno

- Clarifying that piesno in 4D gives one value per slice (3rd dimension), not per volume.
- In dipy, the loop that we were doing manually in our piesno function is already implemented. Deleted.


EmmaRenauld:
Tests pass locally.

Note that test_denoising_nlmeans.py is really slow (3 mins 53 locally).

arnaudbore (approved):

PR #1147, opened 2025-03-17
Fibertube orientation distribution

VincentBeaud:
# Quick description

Main contributions:
- Add Fibertube ODFs as a new propagation method of Fibertube Tracking.
- Add support of deterministic ftODF tracking through peak extraction.
- Support seeding in the **center** of the first segment of each fibertube, instead of only being random placement. Default is still random.
- Add new "endpoint_distance" metric to scoring.

Bugfix:
- Support fibertube scoring with ftODF by adding seed_ids to data_per_streamline. For more info about that, reach out to me It's a long story.
- Fix tracking not working at extreme res. (Due to bad sphere-collision sampling)
- Fix fibertube density script storing the same collision more than once.
- Fix scoring giving bad VC, IC, NC values.

Other:
- Update of relevant documentation.
- Tests + Pep8 for all features
- Because Fibertube Tracking is very slow, it benefits from a smaller "loading tick" when tracking. I made `min_iter` a parameter of Tracker with a default of 100. (like previously)

## Type of change

Check the relevant options.

- [x] This change requires a documentation update

Add diameters of 0.001 to any tractogram (to form "fibertubes") using:
```
scil_tractogram_dps_math.py [my_trk] import diameters --in_dps_single_value 0.001 --out_tractogram fibertubes.trk
```

You can then run a quick standard tracking using:
```
scil_fibertube_tracking.py fibertubes.trk tracking.trk --blur_radius 1 --step_size 1 --nb_fibertubes 1 --nb_seeds_per_fibertube 4 --processes 4
```
It uses the volume-weighted gound-truth orientations as a distribution to propagate.

You can run an ftODF tracking using:
```
scil_fibertube_tracking.py fibertubes.trk tracking_ftODF.trk --blur_radius 1 --step_size 1 --nb_fibertubes 1 --nb_seeds_per_fibertube 4 --processes 4 --use_ftODF --theta 20
```
It produces an ODF and uses the normal scilpy propagator.

Naturally, picking a direction stochastically on an ODF is worse than picking a random ground-truth orientation. So it should generally diverge more.

# Checklist


VincentBeaud:
@CHrlS98 @mdesco @gabknight 
En espérant que le projet est pas trop loins pour vous 😅 
C'est une feature que j'ai faite durant mon stage passé.

docs/source/documentation/fibertube_tracking.rst
```diff
@@ -199,21 +199,34 @@ Tracking
...
+an array of directions. Unfortunately, this process is a little too good
+when compared to normal tracking. This is because each potential direction
+is an actual fibertube orientation. It is not possible to step in between
+them and get lost.
+
+To align as best as possible the performances of fibertube tracking with
+traditional tractography given the same resolution, the fibertube
+directions near each tracking position should be mapped on a sphere
+and then approximated with spherical harmonics. This gives us a
+fibertube ODF or ftODF. An ftODF is nothing short of a local, volume-weighted TODI!
```
mdesco, line 223: An -> A ftODF

scripts/scil_fibertube_tracking.py
```diff
@@ -122,6 +127,35 @@ def _build_arg_parser():
              '(based on the propagator\'s definition of invalid) \n'
              'are never added.')
 
+    ftod_g = p.add_argument_group(
+        'ftODF Options',
+        'Options required if you want to perform fibertube tracking using\n'
+        'fibertube orientation distribution (ftODF).\n'
+        'If you\'re not familiar with these options, please ignore them.')
```
mdesco, line 134: Is this necessary? Maybe mention to go read the demo for more information about the ftODF

scripts/scil_fibertube_tracking.py
```diff
@@ -122,6 +127,35 @@ def _build_arg_parser():
...
+
+    ftod_g.add_argument('--use_ftODF', action='store_true',
+                        help='If set, will build a fibertube orientation\n'
+                        'distribution function at each tracking step. It \n'
+                        'also allows the use of the scilpy ODFPropagator \n'
+                        'instead of FibertubePropagator. Because this \n'
+                        'option saves each streamline in a random order, \n'
+                        'the resulting tractogram cannot be scored using \n'
+                        'scil_fibertube_score_tractogram.py.')
```
mdesco, line 143: I'm not sure I understand the last part? Why can you not score this output? How will it be evaluated then? I remember you showing me scoring for this method in your paper...

mdesco: Does this ftODF have a sh_order? Or an underlying sphere? How do you decide what sphere to use?

mdesco (commented):
Please address these comments and then I will test the new scripts.

scripts/scil_fibertube_tracking.py
VincentBeaud: First comment was a bigger problem than expected, but now it's fixed with modifications in the tracking and scoring scripts.

As for the sphere, you have the same sphere options as scil_tracking_local_dev.py, and I just added sh_order defaulted to 8. (it was previously hardcoded as 8)

docs/source/documentation/fibertube_tracking.rst
```diff
@@ -199,21 +199,35 @@ Tracking
...
 
-
-.. image:: https://github.com/user-attachments/assets/0308c206-c396-41c5-a0e1-bb69b692c101
+.. image:: https://github.com/user-attachments/assets/782bb6d2-0e5c-48a5-8606-1d95d6675e0d
    :alt: Visualization of the blurring sphere intersecting with segments
 
+This is similar to computing the Track Orientation Distribution (TOD)
+but it is not yet represented as a spherical function. It is merely
+an array of directions. Unfortunately, this process is a little too good
+when compared to normal tracking. This is because each potential direction
```
mdesco, line 215: "Unfortunately, ..." -> This does not sound like science. It is not unfortunate. It is explanable and a scientific fact. Please rephrase and keep a boring scientific writing style :)

docs/source/documentation/fibertube_tracking.rst
```diff
@@ -290,24 +317,34 @@ giving us the following output in ``reconstruction_metrics.json``:
...
+- 40% of streamlines had the end of their own fibertube as
+  their arrival fibertube segment. (``"vc_ratio": 0.3``)
+- 40% of streamlines did connect their own fibertube, but instead another fibertube.
+  (``"ic_ratio": 0.4``)
+- 26% of streamlines had an arrival fibertube segment that
+  was not a start nor end segment. (``"nc_ratio": 0.2``)
+- Lastly, we notice that the streamline with the "worst" trajectory was on average
+  ~9.5mm away from its fibertube. (``"mae_max": 9.507027053725844``)
+- Streamlines terminated on average 25.68mm away from the ending of their own
+  fibertube. (``endpoint_dist_mean": 25.675430285869695``)
```
mdesco, line 344: Does that make sense in terms of number??? If the max error is 9.5mm, how can the average error be 25mm??? Seems wrong and odd.

VincentBeaud: It makes more sense with a picture haha.
Here is the tractogram from the demo:
Blue: fibertubes that were seeded, red: streamlines, yellow: absolute error. I added the max_endpoint_dist in pink.
image

As you can see, the worse absolute error (AE) != the worse endpoint distance. This is because AE is computed point per point, and if the streamline has less point than the fibertube, well.. I can't compute all the AE for the fibertube.

So the worse MAE (instead of AE) is even less, considering it's an average across a whole streamline.

I could add this image to the demo so that readers can understand what they're looking at.

scilpy/tractanalysis/fibertube_scoring.py
```diff
@@ -327,12 +330,44 @@ def get_external_vector_from_centerline_vector(vector, r1, r2):
...
+
 @njit
-def resolve_origin_seeding(seeds, centerlines, diameters):
+def associate_seeds_to_fibertubes(seeds, centerlines, diameters):
     """
     Given seeds generated in the first segment of fibertubes (origin seeding)
-    and a set of fibertubes, associates each seed with their corresponding
-    fibertube.
+    and a set of fibertubes, associates each seed/streamline to their
+    corresponding fibertube.
```
CHrlS98, line 370: Missing description for seeds and diameters in section `Parameters` below

scilpy/tractanalysis/fibertube_scoring.py
```diff
@@ -381,9 +418,9 @@ def mean_reconstruction_error(centerlines, centerlines_length, diameters,
         process.
     streamlines_length: ndarray,
         Fixed array containing the number of coordinates of each streamline
-    seeds_fiber: list
+    fibertube_of_seeds: list
```
CHrlS98, line 421: Odd variable name, can't you think of something clearer? `seeded_fibertube_index` maybe?

scripts/scil_fibertube_score_tractogram.py
```diff
@@ -6,12 +6,6 @@
 tracking, computes metrics about the quality of individual fiber
 reconstruction.
 
-IMPORTANT: Streamlines given as input to be scored should be forward-only,
-which means they are saved so that [0] is the seeding position and [-1] is
-the end.
-TODO: Add the seed's segment index as dps, to allow different seeding methods
-and forward_only=False.
-
 Each streamline is associated with an "Arrival fibertube segment", which is
```
CHrlS98, line 9: For info I'm not a fan of the name "Arrival fibertube segment". "termination segment"? "last segment"?

scripts/scil_fibertube_tracking.py
```diff
@@ -122,12 +129,45 @@ def _build_arg_parser():
...
+                        'also allows the use of the scilpy ODFPropagator \n'
+                        'instead of FibertubePropagator.')
+    ftod_g.add_argument('--algo', default='prob', choices=['det', 'prob'],
+                        help='Algorithm to use with ftODF. If ftODF is \n'
+                             'NOT used, this argument is not considered. \n'
+                             '[%(default)s]')
+    add_sphere_arg(ftod_g, symmetric_only=False)
+    add_sh_basis_args(ftod_g)
+    ftod_g.add_argument('--sh_order',
+                        type=int, default=8, choices=[4, 8, 16],
```
CHrlS98, line 150: why are we limited to these choices? I understand it should be even, but it could be 6 no?

CHrlS98 (changes requested):
Here you go @VincentBeaud . That is a big PR.

scilpy/tracking/fibertube_utils.py
```diff
@@ -14,8 +15,14 @@ def streamlines_to_segments(streamlines, verbose=False):
     ----------
     streamlines : list
         Streamlines to segment. This function is compatible with streamlines
-        as a fixed array, as long as the padding value is a number. Padding
-        will also be present in the result value.
+        as a fixed array, as long as the padding value is a number and the
```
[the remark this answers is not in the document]
VincentBeaud: I think I thought about "None" when I wrote this. Or some kind of undefined value.

scripts/scil_fibertube_tracking.py
VincentBeaud: I'm not familiar with the math behind it and assumed it was  powers of 2 only. I'll change it.

scilpy/tracking/fibertube_utils.py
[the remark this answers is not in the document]
VincentBeaud: I'll remove it. 👍🏻

scilpy/tractanalysis/fibertube_scoring.py
```diff
@@ -47,23 +46,26 @@ def fibertube_density(sft, samples_per_voxel_axis, verbose=False):
         fibertubes). This is useful for calculating measurements on the
         various density values, like mean, median, etc.
     collision_grid: ndarray
-        Per-voxel density of fibertube collisions.
+        Per-voxel number of fibertube collisions.
     collision_flat: list
-        Per-voxel density of fibertubes collisions as a list, containing only
+        Per-voxel number of fibertubes collisions as a list, containing only
```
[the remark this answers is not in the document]
VincentBeaud: The "flat" one contains only voxel values that are within the TDI binary mask. So if you do np.ravel(collision_grid), you'll get 12288 values, while collision_flat only has 118. This is useful if you want to do the mean of all "valid" voxels.

The function could return the following: density_grid, collision_grid, mask

But then you need play around with np.where, np.ravel and np.erase to finally get what you want. I figured this was a simpler, more plug and play alternative.

Do you think I should change it? Or better explain it?

scripts/scil_fibertube_score_tractogram.py
VincentBeaud: I agree that it's an ugly name. But "termination segment" and "last segment" don't necessarily imply that I'm talking about a Fibertube segment specifically. It could be the last segment of the streamline.

Even "termination fibertube segment" sounds like I'm just talking about the final segment of any given fibertube. But maybe it would be better?

CHrlS98: "fibertube segment of termination" maybe? but since you describe what it is right after, it should clarify what you mean anyway. It's your call, you know better what you want to say.

scilpy/tractanalysis/fibertube_scoring.py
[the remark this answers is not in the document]
CHrlS98: Maybe it should be include the word `valid` then? I understand it contains the number of collisions per valid voxel inside the flat TDI mask.

scripts/scil_fibertube_tracking.py
```diff
@@ -215,30 +265,54 @@ def main():
...
     fake_mask = DataVolume(fake_mask_data, in_sft.voxel_sizes, 'nearest')
-    datavolume = FibertubeDataVolume(centerlines, diameters, in_sft,
+
+    if args.use_ftODF:
+        logging.debug("Instantiating FTODF datavolume")
+        datavolume = FTODFDataVolume(centerlines, diameters, in_sft,
                                      args.blur_radius,
-                                     np.random.default_rng(args.rng_seed))
+                                     np.random.default_rng(args.rng_seed),
+                                     sh_basis, args.sh_order)
```
CHrlS98, line 274: Do you test that your sh_order is a multiple of 2 somewhere?

CHrlS98 (commented):
Seems pretty good to me. A few comments remaining. I also answered some of our unresolved discussions from my previous PR.

scripts/scil_fibertube_score_tractogram.py
VincentBeaud: Yeah I'm not against that. Better than "arrival".

scripts/scil_fibertube_tracking.py
VincentBeaud: Done through Dipy. 👍🏻

CHrlS98 (approved):

PR #1196, opened 2025-07-10
Feature/python 3.12

frheault:
# Quick description

Getting ready for the next DIPY release and a future switch to Python3.12 !
90% of the file change are the tests that were generating a VERY annoying warning (very long) about script.runner now needing a list of parameters instead of just using kwargs internally
...

## Type of change

Check the relevant options.

...

# Checklist


.python-version
```diff
@@ -1,2 +1,2 @@
-3.10
->=3.9,<3.12
+3.12
+>=3.11, < 3.13
```
AntoineTheb, line 2: Do we really want to drop support for Python 3.10 ?

scilpy/io/utils.py
```diff
@@ -1249,3 +1273,17 @@ def get_default_screenshotting_data(args, peaks=True):
             ovl_imgs,
             ovl_colors,
             peaks_imgs)
+
+
+def convert_stateful_str_to_enum(args):
```
AntoineTheb, line 1278: Is this related to Python 3.12 ? Otherwise move to another PR

scilpy/tractograms/streamline_operations.py
```diff
@@ -656,6 +656,8 @@ def resample_streamlines_step_size(sft, step_size):
 
     # Return to original space
     resampled_sft.to_space(orig_space)
+    resampled_sft.streamlines._data = resampled_sft.streamlines._data.astype(
```
AntoineTheb, line 659: I thought accessing `._data` was forbidden or highly discouraged.

scilpy/tractograms/tests/test_streamline_and_mask_operations.py
```diff
@@ -88,7 +89,7 @@ def test_get_endpoints_density_map_five_points():
     """
 
     sft, reference, *_ = _setup_files()
-
+    print(sft.streamlines._data.dtype)
```
AntoineTheb, line 92: Remove print

scripts/scil_surface_assign_custom_color.py
```diff
@@ -0,0 +1,104 @@
+"""
```
AntoineTheb, line 1: Pretty sure this has nothing to do with Python3.12

scripts/scil_surface_assign_uniform_color.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
```
AntoineTheb, line 1: Same, not related to Python3.12 port

scilpy/tractograms/streamline_operations.py
frheault: Yes, but nibabel default _data dtype is now using `float` so it default to float64 which crash our cython. I don't want to play with cython.

(It is discouraged to slice, overwrite, reshape or shuffle. But I feel that a cast isn't as bad)

scilpy/tractograms/dps_and_dpp_management.py
```diff
@@ -252,26 +254,21 @@ def perform_operation_on_dpp(op_name, sft, dpp_name, endpoints_only=False):
...
     """
     call_op = OPERATIONS[op_name]
     if endpoints_only:
         new_data_per_point = []
         for s in sft.data_per_point[dpp_name]:
-            this_data_per_point = np.nan * np.ones((len(s), 1))
-            this_data_per_point[0] = call_op(s[0])
-            this_data_per_point[-1] = call_op(s[-1])
-            new_data_per_point.append(np.asarray(this_data_per_point)[:, None])
+            this_data_per_point = [call_op(s[0]), call_op(s[-1])]
```
EmmaRenauld, line 264: This is not good in the sense that what is returned is no longer usable as dpp. The docstring says it should return a new_dpp. 

In test_dps_and_dpp_management, at line 186, you could add a 
`    fake_sft.data_per_point['my_dpp2'] = dpp`
and see that it fails. The previous code was always returning arraysequences of the same length as the streamlines.

scilpy/tractograms/streamline_operations.py
EmmaRenauld: Ok. I think eventually you should add this to dipy's stateful tractogram. Like a sft.to_dtype method.

EmmaRenauld (commented):
I thinks its good to go after my comments are fixed.

scilpy/tractograms/tests/test_streamline_and_mask_operations.py
```diff
@@ -62,6 +62,7 @@ def _setup_files():
 
     # Load sft
     sft = load_tractogram(in_sft, reference)
+    sft.streamlines._data = sft.streamlines._data.astype(np.float32)
```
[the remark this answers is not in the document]
frheault: Otherwise cython code is crashing. Streamlines can be 16/32/64 but our cython code only allows one datatype

README.md
```diff
@@ -14,22 +14,10 @@
 **Scilpy** mainly comprises tools and utilities to quickly work with diffusion MRI. Most of the tools are based
 on or are wrappers of the [DIPY] library, and most of them will eventually be migrated to [DIPY]. Those tools implement the recommended workflows and parameters used in the lab.
 
-## Install scilpy as a user
-
-We highly encourage to install scilpy in a virtual environnement. Once done and you're in your virtual environnement you can run this command:
-
-```
-export SETUPTOOLS_USE_DISTUTILS=stdlib
-pip install scilpy
```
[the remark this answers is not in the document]
frheault: If someone is reading the readme of how to install scilpy, I don't think they want the non-programmer installation instructions.

scilpy/tractograms/dps_and_dpp_management.py
frheault: It is valid as dpp because the return converts everything as a ArraySequence, which is the expected type for dpp (list of array was just used to later initialized an ArraySequence)

frheault: A more efficient way to store endpoints only operation would be to use dps with a 2D array store per streamline, rather than storing a lot of nan to then only use 2 values. But I will keep that for later.

frheault:
@EmmaRenauld The dtype operation in SFT already exist, BUT it creates a copy and internally just does the same operation. It would be a huge waste of RAM to duplicate SFT here for an operation that isn't unsafe (we are not using a view or slicing the SFT).

PR #1213, opened 2025-08-08
Add final viz tests (after PR1137)

EmmaRenauld:
After PR #1137, I thought I still had work to do to be able to add these two new tests, but it looks like afterall everything was already fixed!

So, just new tests.


EmmaRenauld:
BTW, I have no idea how viz_bundle_screenshot_mosaic had some coverage, because the wrong --help was called in its test file. Complete mystery to me..

PR #1242, opened 2025-09-25
ENH: tractometer-style volume compare

AntoineTheb:
# Quick description

Change the behavior of `scil_volume_pairwise_comparison` in the case of `--single_compare`. 

If the option is not included, overreach is computed as the number/ratio of voxels in either volumes but not in both, a sort of logical xor. However, if comparing volumes against a reference one (ie using `--single_compare`), it would make more sense to compute the overreach as the number/ratio of voxels in the candidate volume that are not in the reference. This second behavior would be closer to a "tractometer-style" overreach.

This is a breaking change in the case of `--single_compare`.

## Type of change

Check the relevant options.

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

TODO

# Checklist

- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas


arnaudbore (approved):
LGTM

PR #1245, opened 2025-09-30
Fix of codecov components paths

AlexVCaron:
# Quick description

The regex for the library component is not working, the resulting coverage component is just empty.
...

## Type of change

Check the relevant options.

We'll see on the coverage report that will be uploaded for this PR if it works or not

# Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -32,4 +32,4 @@ component_management:
         - "!src/scilpy/cli/"
 
 comment:
-  layout: "condensed_header, diff, components"
+  layout: "condensed_header, diff, flags, components"


EmmaRenauld:
It's nearly there!

When I select unittest, then I get a 0% coverage on src/cli, which overall decreases the percentage to 13%. It think that witout this, we would see ~30%. Can we tell it to ignore the cli path for unittest ? Or is it too complicated?

image

EmmaRenauld:
Also I just noticed that the numbers for the unittest in other files are good, but the numbers for the smoke test are the same as the total coverage from both.

If it can help you: Here are the number that I had noted a few months ago for:
- dwi.operations: 79 with smoke, 72 with unit, 89 total
- dwi.utils: 55 with smoke, 91 with unit, 91 total
- reconst.bingham: 43 with smoke, 88 with unit, 94 total

.github/workflows/test.yml
```diff
@@ -67,12 +69,19 @@ jobs:
...
-        run: |
-          export C_INCLUDE_PATH=$pythonLocation/include/python${{ steps.python-selector.outputs.python-version }}:$C_INCLUDE_PATH
           scil_data_download
           scil_search_keywords test --regenerate_help_files --processes 8
-          uv run --active pytest --cov-report term-missing:skip-covered --dist=loadgroup
+
+      - name: Run unit tests
+        run: |
+          uv run --active pytest src/scilpy --cov-report term-missing:skip-covered --dist=loadgroup --ignore=src/scilpy/cli
+          mv .test_reports unit_test_reports
```
EmmaRenauld, line 78: Maybe the reason that smoke tests have the total coverage is that here you don't do `mv .coverage .coverage.unit`?

AlexVCaron: Actually, the coverage report is also written in the `.test_reports` folder, so we only need to upload its content to coverage and it's sufficient.

I'm looking into the reports right now to find why we get those new numbers

EmmaRenauld:
> I tried to have it only with the flags, not sure if it's possible though. I need to look further in the documentation of `coverage`, it's a bit of a pain.
> 
> But right now, you can get the percentages you want by using both **flags** and **components** to filter the coverage report. Here, I selected the `unittest` flag and the `Library` component : image

Oh! That's what you told us yesterday, sorry!. That's ok for me, then!

EmmaRenauld (approved):
I'm really happy that this is done!

Issue #1206, opened 2025-07-29
Verify out of bound before using map_coordinates


EmmaRenauld:
See explanation in PR #1102

PR #1270, opened 2025-10-31
Add verification of close-to-bound voxel with map_coordinates

EmmaRenauld:
In PR #1102, we changed the padding to nearest. Antoine suggested to add a verification for out-of-bound values. It was not done in PR1102 to keep it simple, but was added as issue #1206. This finishes PR1102.

### DIFF
--- a/src/scilpy/image/volume_space_management.py
+++ b/src/scilpy/image/volume_space_management.py
@@ -1,8 +1,11 @@
 # -*- coding: utf-8 -*-
-import numpy as np
+import logging
 
+import numpy as np
 from numba_kdtree import KDTree
 from numba import njit
+from scipy.ndimage import map_coordinates
+
 from scilpy.tracking.fibertube_utils import (streamlines_to_segments,
                                              point_in_cylinder,
                                              sphere_cylinder_intersection)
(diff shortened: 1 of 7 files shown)


arnaudbore (dismissed):
quick fix and LGTM

src/scilpy/image/volume_space_management.py
```diff
@@ -16,6 +19,39 @@
...
+
+    Returns
+    -------
+    data: np.ndarray
+        The interpolated data.
+    """
+    if (np.any(np.logical_or(points[0] < 0, points[0] > data.shape[0])) or
+        np.any(np.logical_or(points[1] < 0, points[1] > data.shape[1])) or
+        np.any(np.logical_or(points[2] < 0,points[2] > data.shape[2]))) :
+        logging.warning("Careful! You are interpolating outside of boundaries "
```
AntoineTheb: I don't like this as it deviates from the expected behavior of the function. I would instead try to match the expected behavior of `map_coordinates` and return a `cval` if the indices are actually out of bounds instead of logging a message (which may be swallowed).

AntoineTheb (changes requested):
See comment. I would strongly recommend we actually deal with out of bounds coordinates in the function. Otherwise, we'll redo the same logic every time we call our own `map_coordinates`.

EmmaRenauld:
Summarizing because it has been a long time since we managed this:

We have an issue with 3D images (with voxels) when using interpolation with scipy. map_coordinates manages coordinates with a origin=center manner: the first voxel (coordinate [0, 0, 0]) spans from [-0.5, -0.5, -0.5] to [0.5, 0.5, 0.5]. But scipy also considers that everything between -0.5 and 0 is outside the image, and then scipy pads the outside of the image to 0 by default. Si if we want to find the image value there, we will always get 0 if default values as used.

With mode='nearest': (i.e. padding the outside values to the nearest voxel that is in the volume): we're ok.

So here: splitting the voxels on the limit (the half voxels all around) and using them with nearest. Other voxels are used as usual with user choices.

(I also reversed changes we made in #1102, where we added mode 'nearest' everywhere). Now it's only used for voxels on the limit.

src/scilpy/image/volume_space_management.py
```diff
@@ -16,6 +19,79 @@
...
+    shape = np.asarray(data.shape)[:, None]
+
+    # Points that lie in the "half voxel" border region.
+    limit_points = np.any(
+        ((points >= -0.5) & (points < 0)) |
+        ((points > shape - 1) & (points <= shape - 0.5)),
+        axis=0)
+
+    # Points correctly managed by scipy
+    good_points = ~limit_points
```
AntoineTheb, line 75: Unless I missed something, this will also include points that are actually outside the volume, not just at the border. These "bad" points should return a cval imo

AntoineTheb (changes requested):
Sorry to hammer on this, but I think there's still a problem with this function. IMO it should behave exactly as map_coordinate should, just with voxels instead of points, but if that's outside of the scope, it should at least let the user know if coordinates are out of bounds.

PR #1289, opened 2025-12-02
Adding visualization commands and cleaning b-tensor files (to the docs)

karanphil:
# Quick description

I added the command to visualize the fODFs (in the ssst, msmt and memsmt cases).

I also split the b-tensor CSD (memsmt-CSD) and the use of DIVIDE on b-tensor data.

The wm_mask and mask (they are the same) used in the ssst-CSD demo are kind of broken... they have a lot of voxels outside of the brain, and the mask being limited to single-fiber voxels does not produce nice images for the fODF visualization. However, I think it is needed to shorten the computing time.

image


arnaudbore (approved):
GTG

PR #1295, opened 2026-01-06
Transfert functions for reorient

frheault:
# Quick description

New functions to handle reorientations of nibabel images to get Numpy array where L is L, R is R, etc.
...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/src/scilpy/io/stateful_image.py
+++ b/src/scilpy/io/stateful_image.py
@@ -117,6 +117,26 @@ class StatefulImage(nib.Nifti1Image):
                              original_voxel_sizes=reference._original_voxel_sizes,
                              original_axcodes=reference._original_axcodes)
 
+    @staticmethod
+    def convert_to_simg(img):
+        """
+        Initialize a StatefulImage from an existing Nifti1Image.
+
+        This constructor allows creating a StatefulImage directly from a
+        Nifti1Image, preserving its original orientation information.
+
+        Parameters
+        ----------
+        img : nib.Nifti1Image
+            The Nifti1Image to initialize from.
+        """
+        return StatefulImage(img.dataobj, img.affine, header=img.header,
+                             original_affine=img.affine.copy(),
+                             original_dimensions=img.header.get_data_shape(),
+                             original_voxel_sizes=img.header.get_zooms(),
+                             original_axcodes=nib.orientations.aff2axcodes(
+                                 img.affine))
+
     def reorient_to_original(self):
         """
         Reorient the in-memory image to its original orientation.
--- a/src/scilpy/io/tests/test_stateful_image.py
+++ b/src/scilpy/io/tests/test_stateful_image.py
@@ -105,7 +105,7 @@ def test_to_reference_stateful_image():
         ref_img = StatefulImage.load(file_path)
 
         with pytest.raises(TypeError,
-                           match="Reference object cannot be a StatefulImage."):
+                           match="Reference object must not be a StatefulImage."):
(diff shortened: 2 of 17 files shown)


frheault:
This is now aim to merge into master.

Introducing the SFI : StatefulImage
You can look at an example [here](https://drive.google.com/file/d/14cyMGPpZHN-vngfO9wInCn-_nqyMwrjA/view?usp=sharing)!

(The SFI default to RAS)
reorientation_comparison_sfi

I highly suggest someone tries it with my example. I will modify a single script (scil_volume_crop maybe?) later.

src/scilpy/io/stateful_image.py
```diff
@@ -0,0 +1,182 @@
...
+            An instance of StatefulImage with data in the target orientation.
+        """
+        img = nib.load(filename)
+
+        original_affine = img.affine.copy()
+        original_axcodes = nib.orientations.aff2axcodes(img.affine)
+        original_dims = img.header.get_data_shape()
+        original_voxel_sizes = img.header.get_zooms()
+
+        if to_orientation:
```
EmmaRenauld: to_orientation is a string or a tuple, not a bool. Shouldn't it be `if to_orientation == 'RAS'` ?

src/scilpy/io/tests/test_stateful_image.py
```diff
@@ -0,0 +1,196 @@
...
+def test_load_and_reorient(dummy_nifti_file):
+    """
+    Test loading a NIfTI file and reorienting it to RAS.
+    """
+    file_path, _, _ = dummy_nifti_file
+    img = StatefulImage.load(file_path, to_orientation='RAS')
+
+    assert isinstance(img, StatefulImage)
+    assert img.axcodes == ('R', 'A', 'S')
+    assert img.original_axcodes == ('R', 'A', 'S')
```
EmmaRenauld: Couldn't we do a test where the original code is different than RAS?

src/scilpy/io/stateful_image.py
frheault: Ah good catch, it was a None in previous version so this was a test to see if it was set.

I will do a function like valid_orientation to verify if it is ok (String, tuple or list, an iterable with R A and S in it should be accepted)

CHrlS98 (commented):
I like it. Here is my first round of comments. An example will help understanding the code. Also, are the peaks extracted using scilpy displayed correctly inside mi-brain even when the stride is different from [1,2,3,4] ? If so will the new method break the interoperability with mi-brain?

src/scilpy/io/stateful_image.py
```diff
@@ -131,10 +144,24 @@ def to_lps(self):
...
+
+        Raises
+        ------
+        TypeError
+            If ``obj`` is an instance of ``StatefulImage``.
+        """
+
         if isinstance(obj, StatefulImage):
-            raise TypeError("Reference object must not be a StatefulImage.")
+            raise TypeError("Reference object cannot be a StatefulImage.")
```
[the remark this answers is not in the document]
frheault: The subsequent call to get_reference_info() is designed to accept various objects (header, path, nifti, trk, etc.)

Enumerating and asserting every valid type that obj can be would be duplicating what is in that function. Here excluding one specific object that could be problematic type (StatefulImage), is a more practical (and since SFI is new, the handling of that specific type is not well done in Dipy).

So this is just to catch that new SFI type and help the user understand, and then get_reference_info() should 'catch' the rest.

PR #1336, opened 2026-05-20
Addition of warnings about shells choices in FRF and FODF fit

frheault:
# Quick description

This is a manual extraction from my work on StatefulImage once I realized that it is very easy to provide inaccurate shells in our scripts. While mrtrix take hard decisions (always pick the highest shells) in SSST, we do not. We trust the user, but we should at least warn users. There is a heuristic to determine if the provided shells are too far apart. If the delta between shells is higher than the smallest non-zero: warning
(A file with b800 and b1200 is fine, but b700 and b2000 are too big)

In addition to the problem, but to the 'weird' FODF being produced by our SSST when shells were too far apart, there was a warning in our tracking script that detected these 'weird' FODF as peaks, so I also improved the heuristic to distinguish peaks (often of length 15) from FODF (at SH order 4, also length 15). However, when FRF/FODF process with appropriate shells, this was rarely an issue.

...

## Type of change

Check the relevant options.

...

# Checklist

### DIFF
--- a/src/scilpy/cli/scil_frf_ssst.py
+++ b/src/scilpy/cli/scil_frf_ssst.py
@@ -21,7 +21,7 @@ import nibabel as nib
 import numpy as np
 
 from scilpy.gradients.bvec_bval_tools import (check_b0_threshold,
-                                              identify_shells)
+                                              check_shells_frf)
 from scilpy.io.image import get_data_as_mask
 from scilpy.io.utils import (add_b0_thresh_arg, add_overwrite_arg,
                              add_precision_arg,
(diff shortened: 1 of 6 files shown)


src/scilpy/cli/scil_frf_ssst.py
```diff
@@ -111,6 +112,21 @@ def main():
...
+        if len(shells_centroids) > 0 else 0
+    max_non_b0_delta = np.ediff1d(shells_centroids)[0] \
+        if len(shells_centroids) > 1 else 0
+    if max_non_b0_delta >= min_non_b0_shell:
+        logging.warning(
+            'Your shells seem to be very far apart (max delta: {}, '
+            'min non-b0 shell: {}). This might cause problems for the '
+            'estimation of the FRF. Consider using scil_frf_msmt.py.'
+            .format(max_non_b0_delta, min_non_b0_shell))
+
```
arnaudbore, line 116: Good, I think this should into bvec_bval_tools with a test and maybe use parser also to print the warning.

src/scilpy/reconst/utils.py
```diff
@@ -57,3 +59,68 @@ def get_sphere_neighbours(sphere, max_angle):
                     np.outer(zs, zs))
     neighbours = scalar_prods >= np.cos(max_angle)
     return neighbours
+
+
+def is_data_peaks(img_data):
```
arnaudbore, line 64: Add a quick test. I know you won't be able to cover all the code here but we could come up with 2 or 3 tests that will increase coverage.

arnaudbore (commented):
Few comments, It's going to be a fast merged PR.

arnaudbore (approved):
GTG

PR #1338, opened 2026-06-05
(WIP) Reorientation to Canonical RAS with gradient support

frheault:
# Quick description

This PR is the second part of the StatefulImage transition to fully support default orientation to RAS to increase interoperalibity with mrtrix.

Here is testing data:
https://usherbrooke-my.sharepoint.com/:f:/g/personal/rhef1902_usherbrooke_ca/IgCVPHBi54bhR6Cts-QTexW2AeC_dHOexQHtmdXBHX2fEuQ?e=N1L2Md

Overall the idea is that either your bvecs and your DWI are already compatible now you can load them together and it internally flip/swap axes to canonical RAS orientation for fitting, modelling and tracking. So this is a minimal set of modifications to have a good idea if the sequence of steps is working.

If your bvecs are an orientation problem, I have now changed the scil_gradients_validate_correct script to reorient your bvecs to fit the DWI in any strides (or voxel order) and be ready for DTI or FODF fit. Extensions to future scripts, such as msmt, COMMIT, fixels or others, will be done later as it requires more thinking and this PR is big enough by itself.

This breaks previously processed DTI, evecs, FOD and peaks. But it does works much better with mrtrix and avoid manual change to 1,2,3 stride.

...

## Type of change

Check the relevant options.

...

# Checklist


src/scilpy/cli/scil_fodf_msmt.py
```diff
@@ -132,18 +131,31 @@ def main():
...
-    bvals, bvecs = read_bvals_bvecs(args.in_bval, args.in_bvec)
+
+    simg = StatefulImage.load(args.in_dwi)
+    simg.load_gradients(args.in_bval, args.in_bvec)
+
+    # Orientation standardization?
+    # Reconstruction logic (dipy/scilpy) often prefers specific orientation.
+    # We reorient secondary inputs to match the primary one.
+    # If we want to be fully robust, we could force RAS here, but let's see.
+    # scil_frf_msmt used to_ras(), so let's be consistent.
```
CHrlS98, line 142: A bit cryptic for a comment

src/scilpy/tractanalysis/reproducibility_measures.py
```diff
@@ -19,14 +25,18 @@
...
+# Multiprocessing globals
+sft_1 = None
+sft_2 = None
+matched_points_1 = None
+matched_points_2 = None
+tree_1 = None
+tree_2 = None
+sh_data_1 = None
+sh_data_2 = None
+B = None
```
CHrlS98, line 39: not how we usually handle multiprocessing... why?

src/scilpy/io/stateful_image.py
```diff
@@ -66,59 +109,292 @@ def load(cls, filename, to_orientation="RAS"):
...
+            rotated_mu2 = np.dot(mu2, R.T)
+            bingham_data[..., 1:4] = rotated_mu1.reshape(
+                original_shape[:4] + (3,))
+            bingham_data[..., 4:7] = rotated_mu2.reshape(
+                original_shape[:4] + (3,))
+            return bingham_data
+
+        # Handle 5D data
+        if len(original_shape) == 5:
+            # We treat each "lobe" independently for rotation if it's not SH
```
CHrlS98, line 241: misleading comment, this case only flatten 4th and 5th dimension into a single dimension for is_data_peaks no?

src/scilpy/cli/scil_sh_to_sf.py
```diff
@@ -206,10 +219,25 @@ def main():
...
+        if not args.sphere:
+            # Reconstruct world bvecs with b0s if necessary
+            full_world_bvecs = np.zeros((len(new_bvecs), 3))
+            start_idx = data_b0.shape[-1] if args.in_b0 else 0
+            full_world_bvecs[start_idx:] = sphere.vertices
+
+            # Use a dummy StatefulImage to save them in voxel space
+            simg_out = StatefulImage.create_from(sf, simg_sh)
+            simg_out.attach_world_gradients(
+                [0] * len(new_bvecs), full_world_bvecs)
```
arnaudbore:
```suggestion
                new_bvals, full_world_bvecs)
```

CHrlS98:
I added my pipeline for testing asymmetric ODFs and Bingham fit under `test_asym_and_bingham_out` in your sharepoint folder.

frheault:
@arnaudbore All the legacy support for the orientations (--is_voxel_space) is removed, new script to rotate SH and Peaks from old convention (voxel) to new convention (world) and vice-versa. I caught an bug in NODDI/FW because we were saving bvecs and sending them toward AMICO (I saw that by doing EUDX tracking on fit_dir.nii.gz, otherwise the rest is identical). I rerun all tests of scilpy (pytest) and the one on OneDrive and the FODF, Peaks, tracking, etc. all looks great no matter the strides or my sideway brain.

(I even caught a problem with the data in the random stride, I just ran the scil_gradients_validate_correct (could be name scil_gradients_validate_correct_orientation?), it detected the mismatch and fixed it and all was good.

MEMSMT and MSMT seems to work great, Charles said that asym and bingham ran and there was no apparent mistake.

arnaudbore (commented):
forgot this last month.

arnaudbore (commented):
Don't forget this one. 👍

src/scilpy/cli/scil_fodf_ssst.py
```diff
@@ -77,25 +84,26 @@ def main():
...
     sh_order = args.sh_order
     sh_basis, is_legacy = parse_sh_basis_arg(args)
 
-    # Checking data and sh_order
-    if data.shape[-1] < (sh_order + 1) * (sh_order + 2) / 2:
-        logging.warning(
-            'We recommend having at least {} unique DWI volumes, but you '
-            'currently have {} volumes. Try lowering the parameter sh_order '
-            'in case of non convergence.'.format(
-                (sh_order + 1) * (sh_order + 2) / 2, data.shape[-1]))
```
CHrlS98, line 97: Not a check we do anymore?

src/scilpy/cli/scil_gradients_validate_correct.py
```diff
@@ -78,50 +78,110 @@ def main():
...
+        # Transform bvecs
+        # Note: Dipy expects bvecs as (N, 3). We apply the transform to axes.
+        # G' = G @ T
+        bvecs_candidate = bvecs @ t
+
+        gtab_candidate = gradient_table(bvals, bvecs=bvecs_candidate,
+                                        b0_threshold=args.b0_threshold)
+        tenmodel_candidate = TensorModel(gtab_candidate, fit_method='WLS',
+                                         min_signal=np.min(data[data > 0]))
+
```
CHrlS98, line 146: This is a big refactor of this script... Isn't there a lot of stuff in `main()` that could be moved to module?

src/scilpy/cli/scil_bundle_generate_priors.py
```diff
@@ -104,40 +106,48 @@ def main():
...
+    sh_shape = input_sh_3d.shape
+    sh_order = find_order_from_nb_coeff(sh_shape)
+
+    simg_mask = StatefulImage.load(args.in_mask)
+    simg_mask.to_ras()
+    mask_data = get_data_as_mask(simg_mask)
 
     sft = load_tractogram_with_reference(parser, args, args.in_bundle)
+    sft.to_rasmm()
+    sft = StatefulTractogram(sft.streamlines, simg_sh, Space.RASMM)
```
CHrlS98, line 125: Not sure I understand this line, isn't sft already aware of its vox to world transform?

src/scilpy/cli/scil_NODDI_maps.py
```diff
@@ -123,7 +123,11 @@ def main():
     assert_headers_compatible(parser, args.in_dwi, optional=args.mask)
 
     # Generate a scheme file from the bvals and bvecs files
-    bvals, _ = read_bvals_bvecs(args.in_bval, args.in_bvec)
+    simg = StatefulImage.load(args.in_dwi)
```
CHrlS98, line 126: sometimes it's `StatefulImage.load` and other times its `load_img`, when should we use which?

src/scilpy/cli/scil_qball_metrics.py
```diff
@@ -172,31 +166,30 @@ def main():
                                 num_processes=nbr_processes)
 
     if args.gfa:
-        nib.save(nib.Nifti1Image(odfpeaks.gfa.astype(np.float32), img.affine),
-                 args.gfa)
+        res = odfpeaks.gfa.astype(np.float32)
+        StatefulImage.create_from(res, simg).save(args.gfa)
```
CHrlS98, line 170: couldn't we have a `save_img` just like `load_img` that would do just that? Not that it matters much since it would just be a oneliner, but could be clearer when reading.

src/scilpy/cli/scil_search_keywords.py
```diff
@@ -98,9 +98,10 @@ def main():
     nbr_cpu = validate_nbr_processes(parser, args)
 
     if args.verbose == "WARNING":
-        logging.getLogger().setLevel(logging.INFO)
+        logging.basicConfig(level=logging.INFO, format='%(message)s', force=True)
     else:
-        logging.getLogger().setLevel(logging.getLevelName(args.verbose))
+        logging.basicConfig(level=logging.getLevelName(args.verbose),
+                            format='%(message)s', force=True)
```
CHrlS98, line 104: sure about this?

src/scilpy/cli/scil_search_keywords.py
CHrlS98, line 1: unrelated to PR

src/scilpy/cli/scil_sh_reorient.py
```diff
@@ -0,0 +1,87 @@
...
+This script DOES NOT change the voxel grid order, only the spatial
+orientation of what is written inside the voxels.
+
+Legacy Dipy format used to store orientations relative to the voxel grid.
+MRtrix and modern Scilpy store orientations relative to the scanner world
+coordinates, making them invariant to the voxel grid's ordering.
+
+Use this script to update legacy data to world space, or convert back to
+voxel space if needed. Despite the script name (scil_sh_reorient), it
+works on any directional data stored in the volume, including peak directions.
```
CHrlS98, line 15: does it work on bingham distributions?

CHrlS98: if not then I'd update doc to say it only works on sh or peaks

src/scilpy/cli/scil_sh_reorient.py
```diff
@@ -0,0 +1,87 @@
...
+    assert_outputs_exist(parser, args, args.out_image)
+
+    sh_basis, is_legacy = parse_sh_basis_arg(args)
+
+    simg = StatefulImage.load(args.in_image, to_orientation=None,
+                              is_orientation=True,
+                              sh_basis=sh_basis,
+                              is_legacy=is_legacy)
+
+    is_peaks = True if args.peaks else None
```
CHrlS98, line 72: just pass `args.peaks` directly no?

src/scilpy/cli/scil_tracking_local.py
```diff
@@ -189,23 +191,28 @@ def main():
...
+
     # Make sure the data is isotropic. Else, the strategy used
     # when providing information to dipy (i.e. working as if in voxel space)
     # will not yield correct results. Tracking is performed in voxel space
     # in both the GPU and CPU cases.
-    odf_sh_img = nib.load(args.in_odf)
-    if not np.allclose(np.mean(odf_sh_img.header.get_zooms()[:3]),
-                       odf_sh_img.header.get_zooms()[0], atol=1e-03):
+    odf_sh_simg = StatefulImage.load(args.in_odf, is_orientation=True,
+
```
CHrlS98, line 201: remove this empty line

src/scilpy/cli/scil_tracking_local.py
```diff
@@ -223,26 +230,30 @@ def main():
...
-    if np.count_nonzero(seed_img.get_fdata(dtype=np.float32)) == 0:
+    if np.count_nonzero(seed_simg.get_fdata(dtype=np.float32)) == 0:
         raise IOError('The image {} is empty. '
                       'It can\'t be loaded as '
                       'seeding mask.'.format(args.in_seed))
 
-    # Note. Seeds are in voxel world, center origin.
-    # (See the examples in random_seeds_from_mask).
+    # Note. Seeds are in world space (RASMM) for CPU, and voxel space for GPU.
+    # Both use center origin.
```
CHrlS98, line 244: Seeds are always in voxel space, center origin

src/scilpy/cli/scil_tracking_local_dev.py
```diff
@@ -107,10 +108,14 @@ def _build_arg_parser():
...
-                         help="Spherical function relative threshold value "
-                              "for the \ninitial direction. [%(default)s]")
+    track_g.add_argument(
+        '--sfthres_init',
+        metavar='sf_th',
+        type=float,
+        default=0.5,
+        dest='sf_threshold_init',
+        help="Spherical function relative threshold value "
+        "within each voxel for the \ninitial direction. [%(default)s]")
```
CHrlS98, line 118: going a bit crazy on the formatting here

src/scilpy/cli/scil_tracking_local_dev.py
```diff
@@ -382,16 +405,21 @@ def main():
     if propagator is None and propagators:
         propagator = next(iter(propagators.values()))
 
-    mask_res = mask_img.header.get_zooms()[:3]
-    mask = DataVolume(mask_data, mask_res, args.mask_interp)
+    if 'mask' not in locals() or mask is None:
```
CHrlS98, line 408: not very efficient to iterate through all local variables to figure out whether mask is instantiated or not. Just initialize `mask=None` at start of `main`.

src/scilpy/cli/tests/test_gradients_validate_correct.py
```diff
@@ -26,27 +26,8 @@ def test_execution_processing_dti_peaks(script_runner, monkeypatch):
...
-                            'peaks.nii.gz')
-    in_fa = os.path.join(SCILPY_HOME, 'processing',
-                         'fa.nii.gz')
-
-    # test the actual script
-    ret = script_runner.run(['scil_gradients_validate_correct', in_bvec,
-                             in_peaks, in_fa, 'bvec_corr_fodf', '-v'])
+    ret = script_runner.run(['scil_gradients_validate_correct',
+                             in_dwi, in_bval, in_bvec, 'bvec_corr.bvec',
+                             '--fa_thresh', '0.5', '-v'])
```
arnaudbore, line 32: I don't understand these modifications.

src/scilpy/cli/tests/test_tractogram_flip.py
```diff
@@ -23,9 +23,14 @@ def test_execution_surface_vtk_fib(script_runner, monkeypatch):
...
+    script_runner.run(['scil_tractogram_convert', in_trk,
+                       'gyri_fanning.vtk', '--reference', in_fa,
+                       '--no_bbox_check'])
+    ret = script_runner.run(['scil_tractogram_flip',
+                             'gyri_fanning.vtk',
+                             'gyri_fanning_flipped.vtk',
+                             'x',
+                             '--reference',
+                             in_fa,
+                             '--no_bbox_check'])
```
arnaudbore, line 35: yeurk

src/scilpy/cli/scil_tracking_local_dev.py
```diff
@@ -435,27 +468,17 @@ def main():
...
-    # space after tracking is voxmm.
-    # Smallest possible streamline coordinate is (0,0,0), equivalent of
-    # corner origin (TrackVis)
-    sft = StatefulTractogram(streamlines, mask_img,
-                             space=our_space, origin=our_origin,
-                             data_per_streamline=data_per_streamline)
-    save_tractogram(sft, args.out_tractogram)
+        tracker.save_rap_entry_exit_mask(args.rap_save_entry_exit, mask_simg)
+
+    # save streamlines on-the-fly to file
```
CHrlS98, line 476: since `Tracker.track()` is not a generator it is not saving "on-the-fly"

src/scilpy/cli/scil_tracking_pft.py
```diff
@@ -80,47 +81,21 @@ def _build_arg_parser():
...
-                         help='Spherical function relative threshold value '
-                              'for the \ninitial direction. [%(default)s]')
-    add_sh_basis_args(track_g)
+    track_g.add_argument(
+        '--sfthres_init',
+        dest='sf_threshold_init',
+        type=float,
+        default=0.5,
+        help='Spherical function relative threshold value '
+        'within each voxel for the \ninitial direction. [%(default)s]')
```
CHrlS98, line 98: crazy formatting here also

src/scilpy/cli/scil_tracking_pft.py
```diff
@@ -276,64 +262,46 @@ def main():
...
         seeds,
-        np.eye(4),
+        tracking_affine,
         max_cross=1,
         step_size=vox_step_size,
         maxlen=max_steps,
-        pft_back_tracking_dist=args.back_tracking,
-        pft_front_tracking_dist=args.forward_tracking,
+        pft_back_tracking_dist=args.back_tracking / voxel_size,
+        pft_front_tracking_dist=args.forward_tracking / voxel_size,
```
CHrlS98, line 291: ouch good catch... will certainly have an effect on my OCT tractography paper

src/scilpy/cli/scil_tractogram_flip.py
CHrlS98, line 1: is this related to stateful image?

src/scilpy/cli/scil_tractogram_pairwise_comparison.py
```diff
@@ -105,7 +107,23 @@ def main():
...
     sft_2 = load_tractogram_with_reference(parser, args, args.in_tractogram_2)
-    mask = get_data_as_mask(nib.load(args.in_mask)) if args.in_mask else None
+
+    # Force RAS alignment
+    if args.reference is None:
+        parser.error('--reference is required to force RAS alignment.')
+    ref_simg = StatefulImage.load(args.reference)
+
+    sft_1.to_rasmm()
+    sft_1 = StatefulTractogram(sft_1.streamlines, ref_simg, Space.RASMM)
```
CHrlS98, line 117: I raised this comment earlier, but why do we do need to do this?

src/scilpy/cli/scil_tracking_local_dev.py
```diff
@@ -107,10 +108,14 @@ def _build_arg_parser():
                          type=int, default=0,
                          help='Subdivides each face of the sphere into 4^s new'
                               ' faces. [%(default)s]')
-    track_g.add_argument('--sfthres_init', metavar='sf_th', type=float,
-                         default=0.5, dest='sf_threshold_init',
-                         help="Spherical function relative threshold value "
-                              "for the \ninitial direction. [%(default)s]")
+    track_g.add_argument(
```
CHrlS98, line 111: pep8 went a bit crazy with this one

src/scilpy/cli/scil_tracking_local.py
```diff
@@ -223,26 +230,30 @@ def main():
...
-
-    if np.count_nonzero(seed_img.get_fdata(dtype=np.float32)) == 0:
+    if np.count_nonzero(seed_simg.get_fdata(dtype=np.float32)) == 0:
         raise IOError('The image {} is empty. '
                       'It can\'t be loaded as '
                       'seeding mask.'.format(args.in_seed))
 
-    # Note. Seeds are in voxel world, center origin.
-    # (See the examples in random_seeds_from_mask).
+    # Note. Seeds are in world space (RASMM) for CPU, and voxel space for GPU.
```
CHrlS98, line 243: always vox space and center origin

src/scilpy/cli/scil_sh_reorient.py
CHrlS98, line 72: `args.peaks` is already either true or none.

src/scilpy/cli/scil_sh_reorient.py
CHrlS98, line 15: don't think it works on bingham distributions or tensor coefficients. Only SH and peaks.

src/scilpy/cli/scil_search_keywords.py
CHrlS98, line 1: unrelated to current PR

src/scilpy/cli/scil_qball_metrics.py
CHrlS98, line 170: we could have a save_img just like load_img for this. Might be more intuitive even if it's still just a one-liner.

src/scilpy/cli/scil_tracking_local_dev.py
CHrlS98, line 408: seems inefficient to iterate through all local variables to determine if mask is initialized. Initialize it to `mask=None` instead?

src/scilpy/cli/scil_tracking_local_dev.py
CHrlS98, line 476: `tracker.track()` is not a generator, saving won't be done on-the-fly.

src/scilpy/cli/scil_tracking_pft.py
CHrlS98, line 291: good catch, that's a big issue for my S-OCT tractography pipeline...

src/scilpy/cli/scil_tractogram_pairwise_comparison.py
```diff
@@ -105,7 +107,23 @@ def main():
...
+    sft_2.to_rasmm()
```
CHrlS98, line 118: raised this comment earlier, I stll don't understand why we need this?

src/scilpy/cli/scil_volume_modify_voxel_order.py
```diff
@@ -54,6 +55,15 @@ def _build_arg_parser():
     p.add_argument('--new_voxel_order', required=True,
                    help='The new voxel order (e.g., "RAS", "1,2,3").')
 
+    p.add_argument('--in_bvec',
+                   help='Path of the b-vectors file.')
+    p.add_argument('--in_bval',
+                   help='Path of the b-values file.')
+    p.add_argument('--out_bvec',
+                   help='Path of the modified b-vectors file to write.')
+    p.add_argument('--out_bval',
```
CHrlS98, line 64: why would it modify the bval file?

src/scilpy/cli/scil_volume_resample.py
```diff
@@ -100,15 +99,15 @@ def main():
...
 
         # Must not verify that headers are compatible. But can verify that, at
         # least, the first columns of their affines are compatible.
-        img_zoom_invert = [1 / zoom for zoom in ref_img.header.get_zooms()[:3]]
+        img_zoom_invert = [1 / zoom for zoom in simg.header.get_zooms()[:3]]
         ref_zoom_invert = [1 / zoom for zoom in ref_img.header.get_zooms()[:3]]
 
-        img_affine = np.dot(simg.affine[:3, :3], img_zoom_invert)
-        ref_affine = np.dot(ref_img.affine[:3, :3], ref_zoom_invert)
+        img_affine = np.dot(simg.affine[:3, :3], np.diag(img_zoom_invert))
```
CHrlS98, line 109: we sure about that? cause it's not the same at all

src/scilpy/gpuparallel/opencl_utils.py
```diff
@@ -11,6 +11,12 @@
 from dipy.utils.optpkg import optional_package
 cl, have_opencl, _ = optional_package('pyopencl')
 
+if have_opencl:
```
CHrlS98, line 14: unrelated to stateful image

src/scilpy/io/image.py
CHrlS98, line 18: incomplete docstring

src/scilpy/io/image.py
```diff
@@ -95,7 +95,11 @@ def get_data_as_mask(mask_img, dtype=np.uint8):
 
     # Verify that loaded datatype is ok
     curr_type = mask_img.get_data_dtype().type
-    basename = os.path.basename(mask_img.get_filename())
+    if hasattr(mask_img, 'get_filename') and mask_img.get_filename():
+        basename = os.path.basename(mask_img.get_filename())
+    else:
+        basename = "unnamed"
+
```
CHrlS98, line 102: doesn't the method already expect a nibabel image? this feels a bit overkill no?

src/scilpy/io/stateful_image.py
```diff
@@ -66,11 +105,164 @@ def load(cls, filename, to_orientation="RAS"):
...
+
+            data = self.get_fdata(dtype=np.float32)
+            R = self._get_rotation_matrix(self.affine).T
+            rotated_data = self._rotate_direction_data(
+                data, R, sh_basis=sh_basis, is_legacy=is_legacy,
+                nbr_processes=nbr_processes, is_peaks=is_peaks)
+            self._dataobj = rotated_data
+            return rotated_data
+
+        # R_world_to_voxel = R_voxel_to_world.T
```
CHrlS98, line 161: is this commented code or a comment? If it's a comment, why is it true?

src/scilpy/reconst/fodf.py
CHrlS98, line 1: I think I'd prefer we revert these changes for this PR and keep to a distinct PR. It's a possible source of errors that is unrelated to StatefulImage

src/scilpy/reconst/sh.py
```diff
@@ -354,7 +478,8 @@ def peaks_from_sh(shm_coeff, sphere, mask=None, relative_peak_threshold=0.5,
     # Bring back to the original shape
     peak_dirs_array = np.zeros(data_shape[0:3] + (npeaks, 3))
     peak_values_array = np.zeros(data_shape[0:3] + (npeaks,))
-    peak_indices_array = np.zeros(data_shape[0:3] + (npeaks,))
+    peak_indices_array = np.full(
+        data_shape[0:3] + (npeaks,), -1, dtype=np.int32)
```
CHrlS98, line 482: -1 instead of 0, why?

src/scilpy/reconst/sh.py
```diff
@@ -642,19 +771,17 @@ def convert_sh_basis(shm_coeff, sphere, mask=None,
...
         chunk_len = np.cumsum([0] + [len(c) for c in shm_coeff_chunks])
         tmp_shm_coeff_array = np.zeros((np.count_nonzero(mask), data_shape[3]))
-        for i, new_shm_coeff in results:
-            tmp_shm_coeff_array[chunk_len[i]:chunk_len[i+1], :] = new_shm_coeff
+
+        for i, new_shm_coeff in pool.imap_unordered(
+            _convert_sh_basis_parallel, zip(
+                shm_coeff_chunks, itertools.repeat(B_in), itertools.repeat(invB_out), np.arange(
+                len(shm_coeff_chunks)))):
+            tmp_shm_coeff_array[chunk_len[i]                                :chunk_len[i + 1], :] = new_shm_coeff
```
CHrlS98, line 781: too many spaces

src/scilpy/tracking/propagator.py
```diff
@@ -375,11 +375,6 @@ def __init__(self, datavolume, step_size,
         super().__init__(datavolume, step_size, rk_order, dipy_sphere,
                          sub_sphere, space, origin)
 
-        if self.space == Space.RASMM:
-            raise NotImplementedError(
-                "This version of the propagator on ODF is not ready to work "
-                "in RASMM space.")
-
```
CHrlS98, line 382: okay, however we don't test that it now does work because we still do everything in voxel space in `scil_tracking_local_dev`. So I'm really unsure the dev tracking code is actually robust to rasmm data.

src/scilpy/tracking/utils.py
```diff
@@ -248,60 +251,98 @@ def save_tractogram(
...
+                    # Add half-voxel shift to go from scilpy center-origin
+                    # to nibabel corner-origin in voxmm.
+                    strl_to_save = (strl_vox + 0.5) * original_voxel_size
                 else:
-                    # Streamlines are dumped in true world space with
-                    # origin center as expected by .tck files.
-                    strl = np.dot(strl, ref_img.affine[:3, :3]) + \
-                        ref_img.affine[:3, 3]
+                    # TCK expects RASMM
+                    strl_to_save = strl_rasmm
```
CHrlS98, line 320: can we validate that this fixes issue #1340 ?

CHrlS98 (changes requested):
Alright, here is a first pass.

src/scilpy/cli/tests/test_tractogram_convert.py
```diff
@@ -24,7 +24,8 @@ def test_execution_surface_trk_tck(script_runner, monkeypatch):
     in_fa = os.path.join(SCILPY_HOME, 'surface_vtk_fib',
                          'fa.nii.gz')
     ret = script_runner.run(['scil_tractogram_convert', in_fib,
-                            'gyri_fanning.tck', '--reference', in_fa])
+                             'gyri_fanning.tck', '--reference', in_fa,
+                             '--no_bbox_check'])
```
arnaudbore, line 28: why both of them with no bbox ?

src/scilpy/cli/tests/test_volume_apply_transform.py
```diff
@@ -60,3 +60,25 @@ def test_execution_interp_lin(script_runner, monkeypatch):
...
+    # Run the transformation
+    ret = script_runner.run(['scil_volume_apply_transform',
+                             in_model, in_fa, in_aff,
+                             out_filename, '--inverse', '-f'])
+    assert ret.success
+
+    # Check for header compatibility between the output and the reference
+    ret = script_runner.run(['scil_header_validate_compatibility',
+                             out_filename, in_fa])
+    assert ret.success, "Headers are not compatible!"
```
arnaudbore, line 84: If you want to have this test of scil_header_validate_compatibility put it with this test: test_execution_bst

src/scilpy/cli/scil_fodf_msmt.py
arnaudbore, line 142: you are forcing RAS no ?

src/scilpy/cli/scil_tractogram_pairwise_comparison.py
```diff
@@ -105,7 +107,23 @@ def main():
...
+    sft_2 = StatefulTractogram(sft_2.streamlines, ref_simg, Space.RASMM)
+
+    if args.in_mask:
```
arnaudbore, line 121: Set mask = None before like you do everywhere else.

src/scilpy/io/image.py
```diff
@@ -95,7 +95,11 @@ def get_data_as_mask(mask_img, dtype=np.uint8):
 
     # Verify that loaded datatype is ok
     curr_type = mask_img.get_data_dtype().type
-    basename = os.path.basename(mask_img.get_filename())
+    if hasattr(mask_img, 'get_filename') and mask_img.get_filename():
+        basename = os.path.basename(mask_img.get_filename())
+    else:
+        basename = "unnamed"
```
arnaudbore, line 101: To remove

src/scilpy/io/image.py
arnaudbore, line 34: check also if dtype is integer otherwise raise warning every time.

src/scilpy/cli/scil_viz_bingham_fit.py
```diff
@@ -118,18 +118,20 @@ def main():
     assert_inputs_exist(parser, args.in_bingham)
     assert_outputs_exist(parser, args, [], args.output)
 
-    data = _get_data_from_inputs(args)
+    data, affine = _get_data_from_inputs(args)
```
arnaudbore, line 121: you get the affine but never use it?

src/scilpy/io/mti.py
```diff
@@ -228,7 +228,7 @@ def _prepare_B1_map(args, flip_angles, extended_dir, affine):
     """
     B1_map = None
     if args.in_B1_map and args.in_mtoff_t1:
-        B1_img = nib.load(args.in_B1_map)
+        B1_img, _ = load_img(args.in_B1_map)
```
arnaudbore, line 231: why did we switch everything to load_img ?

src/scilpy/tractanalysis/reproducibility_measures.py
arnaudbore, line 39: agent went bazinga...I think

src/scilpy/tracking/tracker.py
```diff
@@ -392,7 +396,7 @@ def _get_streamlines(self, chunk_id, lock=None):
             # on current process ID.
             eps = s + chunk_id / (self.nbr_processes + 1)
             line_generator = np.random.default_rng(
-                np.abs(hash((seed + (eps, eps, eps), self.rng_seed))))
+                np.abs(hash((tuple(seed + (eps, eps, eps)), self.rng_seed))))
```
arnaudbore, line 399: since everything is autopep8 I'm not sure why here it needs to be casted into a tuple.

src/scilpy/reconst/fodf.py
```diff
@@ -104,35 +100,46 @@ def get_ventricles_max_fodf(data, fa, md, zoom, sh_basis,
...
+    logging.info(f'Number of voxels detected: {len(list_of_max)}')
     if len(list_of_max) == 0:
         logging.warning('No voxels found for evaluation! Change your fa '
                         'and/or md thresholds')
         return 0, out_mask
 
-    logging.info('Average max fodf value: {}'.format(np.mean(list_of_max)))
-    logging.info('Median max fodf value: {}'.format(np.median(list_of_max)))
+    logging.info(f'Average max fodf value: {np.mean(list_of_max)}')
+    logging.info(f'Median max fodf value: {np.median(list_of_max)}')
```
arnaudbore, line 142: why now ?

src/scilpy/cli/scil_gradients_validate_correct.py
```diff
@@ -2,23 +2,15 @@
 # -*- coding: utf-8 -*-
 """
 Detect sign flips and/or axes swaps in the gradients table from a fiber
-coherence index [1]. The script takes as input the principal direction(s)
-at each voxel, the b-vectors and the fractional anisotropy map and outputs
-a corrected b-vectors file.
+coherence index [1]. The script takes as input the DWI, b-values and b-vectors
```
arnaudbore, line 5: I know it's not that much but this script alone would be better alone in its on PR.

src/scilpy/cli/scil_fodf_ssst.py
```diff
@@ -107,13 +115,49 @@ def main():
...
+            'Your shells seem to be very far apart (max delta: {}, min non-b0 shell: {}). '
+            'This might cause problems for the estimation of the FRF. '
+            'Consider using scil_frf_msmt.py.'.format(
+                max_non_b0_delta, min_non_b0_shell))
+
+    if len(dwi_shells) > 0 and np.max(dwi_shells) < 900 and sh_order > 4:
+        logging.warning(
+            'Your maximum b-value ({}) is relatively low. '
+            'High SH order ({}) might be unstable. '
+            'Consider using --sh_order 4.'.format(np.max(dwi_shells), sh_order))
```
arnaudbore, line 142: All this verifications can be done somewhere else...it would also help understand what is done.

src/scilpy/reconst/sh.py
arnaudbore, line 781: Need to be cleaned

src/scilpy/cli/scil_volume_math.py
```diff
@@ -76,9 +77,10 @@ def main():
     # Find at least one mask, but prefer a 4D mask if there is any.
     mask = None
     found_ref = False
+    ref_img = None
     for input_arg in args.in_args:
         if not is_float(input_arg):
-            ref_img = nib.load(input_arg)
+            ref_img, _ = load_img(input_arg)
```
arnaudbore, line 83: I know we want a statefulImage but load_img is heavier is doing some float check that we don't need here... I think we could come up with a slightly lighter way to do that.

src/scilpy/reconst/sh.py
arnaudbore, line 482: -1 ?

src/scilpy/cli/scil_bundle_generate_priors.py
arnaudbore, line 125: double rasmm !!

src/scilpy/cli/scil_search_keywords.py
```diff
@@ -98,9 +98,10 @@ def main():
     nbr_cpu = validate_nbr_processes(parser, args)
 
     if args.verbose == "WARNING":
```
arnaudbore, line 100: it should work but could have been done in a different PR.

src/scilpy/cli/scil_tracking_pft.py
```diff
@@ -80,47 +81,21 @@ def _build_arg_parser():
     p.add_argument('out_tractogram',
                    help='Tractogram output file (must be .trk or .tck).')
 
-    track_g = p.add_argument_group('Tracking options')
+    track_g = add_tracking_options(p)
     track_g.add_argument('--algo', default='prob', choices=['det', 'prob'],
                          help='Algorithm to use (must be "det" or "prob"). '
                               '[%(default)s]')
-    track_g.add_argument('--step', dest='step_size', type=float, default=0.2,
```
arnaudbore, line 87: default step size with add_tracking_options is 0.5.

src/scilpy/cli/scil_tracking_local_dev.py
```diff
@@ -414,11 +445,13 @@ def main():
         rap = None
 
     logging.info("Instantiating tracker.")
+    # We must force save_seeds=True so that Tracker returns
+    # (streamlines, seeds) as expected by scilpy.tracking.utils.save_tractogram
     tracker = Tracker(propagator, mask, seed_generator, nbr_seeds, min_nbr_pts,
                       max_nbr_pts, args.max_invalid_nb_points,
-                      compression_th=args.compress_th,
+                      compression_th=None,
```
arnaudbore, line 452: why we need to get ride of compression ?

src/scilpy/io/stateful_image.py
arnaudbore, line 161: cryptic

arnaudbore (commented):
OMG. That was NOT fun. did not check most of the test files.

src/scilpy/cli/tests/test_gradients_validate_correct.py
frheault: The previous script assumed that orientations were already computer (peaks), but this was an approximation of the original algorithms (was not testing possibility of correct strides with wrong bvecs).

This new version is slower (recompute the DTI/evec1 everything combination), but more accurate.

src/scilpy/cli/tests/test_tractogram_convert.py
frheault: I believe at some point the test was failling (probably due to my SFT hack with re-assigning a new reference). It seems like it is non neccesary (probably got fixed along the way and I did not revert the test)

src/scilpy/cli/tests/test_tractogram_flip.py
frheault: Same answer as scil_tractogram_convert, it seems at some point there was a floating point problem that was making the bbox fail. Reverted too.

src/scilpy/cli/scil_NODDI_maps.py
frheault: It should never be load_img except in scil_volume_math, it seems like a lot of script from Manon use it for no reason, I can fix it. But this adds more line to this PR ;)

src/scilpy/cli/scil_bundle_generate_priors.py
frheault: Discussed IRL

But short version: Since SFT is not aware of the existence of SIMG this trick is required because the affine attribute of the image's header (modified to RAS) is not compatible with the tractogram's header.

src/scilpy/cli/scil_fodf_ssst.py
frheault: Done in verify_data_vs_sh_order

src/scilpy/cli/scil_fodf_ssst.py
frheault: Done, moved to src/scilpy/gradients/bvec_bval_tools.py

src/scilpy/cli/scil_gradients_validate_correct.py
frheault: Since this come with the whole gradient and simg stuff to actually make it work (compatible with mrtrix for validation), this would have to be reverted (that script only) and then after this PR is merge this script is re-added to dev_3.0.x.

Would you prefer that ? (fairly easy, but it two-pass)

src/scilpy/cli/scil_qball_metrics.py
frheault: @arnaudbore what do you think?

I can do an approach like save_data(new_data, reference=simg, filename)
(in this case it is not save_img, since we don't have an img)

src/scilpy/tractanalysis/reproducibility_measures.py
frheault: It was already using global variables before in _compute_difference_for_voxel, it is not a good practice to have global variables in functions that are not declared at the top of a module.
If I remember correctly it was a warning I saw when I modified the rotation_matrix parameters and I was unsure if my local modification to orientation (in the function) had an impact outside. So I fixed that warning in case

(I did not introduce the global variable in multiprocessing in this PR)

frheault: It was already using global variables before in _compute_difference_for_voxel, it is not a good practice to have global variables in functions that are not declared at the top of a module.
If I remember correctly it was a warning I saw when I modified the rotation_matrix parameters and I was unsure if my local modification to orientation (in the function) had an impact outside. So I fixed that warning in case

src/scilpy/tracking/utils.py
frheault: Based on my testing, output both as tck and trk are equivalent (I tested with various voxel sizes and strides) and they were the same in MI-Brain and post-reloading using nibabel.

src/scilpy/tracking/tracker.py
frheault: When I was doing testing I remember an warning when I was testing with single voxel and it was simply that an hash function requires a unmodifiable type (tuple) and since the addition operation between a list or numpy array to works, but then it is not hashable, this is safer

frheault: But I will revert this one, it is indeed not very important in this PR

src/scilpy/tracking/propagator.py
frheault: The warning still exist in the tracker.py, my understanding was this was duplicated code. But it is true they could be used separately. I will add it back

src/scilpy/reconst/sh.py
frheault: This should indeed be another PR, this fixes a bug where empty/background voxels were mistakenly counted as having a real fiber direction, because "no peak here" and "peak pointing at direction #0" were both being stored as the same number (0) until this change made "no peak" its own distinct marker (-1).

This was causing me trouble when checking is_peaks after recomputing fodf_metrics

frheault: This should indeed be another PR, this fixes a bug where empty/background voxels were mistakenly counted as having a real fiber direction, because "no peak here" and "peak pointing at direction #0" were both being stored as the same number (0) until this change made "no peak" its own distinct marker (-1).

This was causing me trouble when checking is_peaks after recomputing fodf_metrics

src/scilpy/io/stateful_image.py
frheault: I removed the comment, at the time it was not a internal function, but now the documentation of the function is clearer and the comment useless.

Specifically for orientation (fodf, peaks) it is very important to have no shearing AND no scaling in the affine so that function does garantee that.

frheault: Remove, now the function is internal and the docstring is clearer

src/scilpy/io/stateful_image.py
frheault: Should I say that we "flatten the orientations if they are peaks" instead of talking about lobes if not SH (like a double negation or something)

src/scilpy/cli/scil_volume_modify_voxel_order.py
frheault: It does not. Will remove

src/scilpy/cli/scil_volume_math.py
frheault: Switch to simg

src/scilpy/cli/scil_volume_resample.py
frheault: It was a bug before, but not a problem since we always use this script with good input. I changed the code to use _get_rotation_matrix directly which is even more robust.

src/scilpy/cli/scil_tractogram_pairwise_comparison.py
frheault: Discussed IRL

Because StatefulTractogram is unaware of the existance of scilpy's StatefulImage a TRK with a header (in non-RAS) wouldn't be compatible with a SIMG in RAS (because I manipulate .affine and the original is stored as .original_affine)

src/scilpy/cli/scil_tractogram_pairwise_comparison.py
frheault: Discussed IRL

Because StatefulTractogram is unaware of the existance of scilpy's StatefulImage a TRK with a header (in non-RAS) wouldn't be compatible with a SIMG in RAS (because I manipulate .affine and the original is stored as .original_affine)

src/scilpy/cli/scil_tractogram_flip.py
frheault: It is not, I probably used it during testing with TCK and it broke. Will do a separated PR

src/scilpy/cli/scil_viz_bingham_fit.py
frheault: I will investiguate, if my memory is right it should not be used, and despite the new return of that function I should has used data, _ =

src/scilpy/cli/scil_tracking_pft.py
frheault: I must admit I was on autofix anytime I saw either mm or world space parameters, so I had no idea this was big or not... If you confirm you have a dataset where this was a problem, that's great !

src/scilpy/cli/scil_tracking_local_dev.py
frheault: Thanks ! I taught it was !

src/scilpy/cli/scil_tracking_local_dev.py
frheault: I taught the streamlines were generated using a yield, so that later save_tractogram had that compress_thr. My mistake, I put back the threshold here and put None in the save_tractogram

src/scilpy/cli/scil_sh_reorient.py
frheault: Good catch, this is leftover from old behavior where True means it is peaks, but None meant run the heuristic and warn the user, but with changes in the code it does not make sense anymore

frheault: Ok I actually changed a bit the warning vs error here, because if someone does not say it is peaks but it has absolutely no chance of being SH (10x10x10x5x3) for example the heuristic is still needed.

Here is a schematic of my new code (made by Gemini), leaving it either True or None from the scripts
_rotate_direction_data(data, is_peaks=...)
│
├─ is_peaks = True  ─────────────────────────► is_sh = False
├─ is_peaks = False ─────────────────────────► is_sh = True
│
└─ is_peaks = None (not specified)
   │
   └─ last_dim = data.shape[-1]
      │
      ├─ last_dim == 3
      │     (CERTAIN: no valid SH order has exactly 3 coeffs)
      │     └─► is_peaks_guess = True, silent, is_sh = False
      │
      └─ last_dim != 3
            (AMBIGUOUS: size could be valid SH *or* N*3 peaks,
             e.g. 15 coeffs = order-4 full SH, also divisible by 3)
            └─► is_data_peaks(data) statistical guess
                (norm uniformity / argmax position / exact-zero ratio)
                └─► logging.warning("auto-detected as SH|peaks, "
                                    "pass is_peaks explicitly if wrong")
                └─► is_sh = not guess

is_sh
│
├─ True  ──► rotate_sh(data, R, ...)         # Wigner-D based, correct for SH
│
└─ False ──► reshape(-1,3); data @ R.T        # naive 3-vector rotate,
             (requires last_dim % 3 == 0,       correct only for real peaks
              else: raise ValueError)

src/scilpy/cli/scil_sh_reorient.py
frheault: I extracted all fixels/bingham modifications to another branch that will follow after this is merged, it will be easier to tests and easier to review.

src/scilpy/cli/scil_qball_metrics.py
frheault: @arnaudbore do you want a save_img that takes `save_img(data, ref)` ?

frheault:
@CHrlS98 @arnaudbore I believe I now also hate VSCode Github extension as now half my comments are reply and half are reviews... I don't know how this happened, I am pretty sure I was always clicking on the ''add as a comment'' and not ''start a review''

EDIT : After checking it seems like all the comments were in response to Charles comments that were duplicated (it seems like your first round of review that you lost was still valid and I replied to them in VSCode, but my comments appears as a review instead of as comment... I am not sure I understand, but I apologize !

src/scilpy/cli/scil_bingham_metrics.py
CHrlS98, line 1: we should add a todo or warning for the moment in case you become a father before the next PR is ready

src/scilpy/cli/scil_gradients_validate_correct.py
CHrlS98: Doesn't the current version with StatefulImage work? If we transform peaks to vox space the current method should work no?

src/scilpy/gradients/tests/test_bvec_bval_tools.py
```diff
@@ -126,15 +127,33 @@ def test_round_bvals_to_shell():
...
+    bvals = np.asarray([0, 0, 700, 700, 700, 700])
+    verify_bval_range(bvals, b0_threshold=20, sh_order=4)
+
+    # Test case where max b-value is high enough (no warn)
+    bvals = np.asarray([0, 0, 1000, 1000, 1000, 1000])
+    verify_bval_range(bvals, b0_threshold=20, sh_order=8)
+
+    # Test case with no non-b0 shells
+    bvals = np.asarray([0, 0, 10, 10])
+    verify_bval_range(bvals, b0_threshold=20, sh_order=8)
```
arnaudbore: to be efficient those warning should be catched otherwise it's not very helpful...

src/scilpy/reconst/fiber_coherence.py
```diff
@@ -59,6 +74,66 @@ def compute_coherence_table_for_transforms(directions, values):
...
 
+def find_best_gradient_correction(data, bvals, bvecs, fa, mask,
+                                  b0_threshold, verbose=True):
+    """
+    Refit the DTI model under all 24 axis permutations/flips of bvecs,
+    returning the transform that maximizes fiber coherence.
+
+    Contrary to compute_coherence_table_for_transforms, which rotates
+    already-fitted peaks, this refits the tensor model from scratch for
+    each candidate transform, which is more robust but more expensive.
```
CHrlS98: I agree for more expensive, but not sure about more robust part. Why would it be that much more robust?

src/scilpy/io/stateful_image.py
CHrlS98: I think so, that's what this is right? Not sure this would flatten other fixel representations (e.g. Bingham) correctly, so lobes may be too general

src/scilpy/cli/scil_qball_metrics.py
arnaudbore: I'm in favor of something like `save_img`. Could it be only `StatefulImage.save` since we are using `StatefulImage.load`.

src/scilpy/gpuparallel/opencl_utils.py
arnaudbore: Still interesting. Submit a PR in master directly.

src/scilpy/cli/tests/test_tractogram_convert.py
arnaudbore: change at least one of them without.

src/scilpy/tracking/utils.py
CHrlS98: If .tck also works in mrview I think we can consider the issue solved

src/scilpy/io/mti.py
arnaudbore: Still not modified or not answered

src/scilpy/cli/scil_volume_resample.py
```diff
@@ -103,13 +103,10 @@ def main():
 
         # Must not verify that headers are compatible. But can verify that, at
         # least, the first columns of their affines are compatible.
-        img_zoom_invert = [1 / zoom for zoom in simg.header.get_zooms()[:3]]
-        ref_zoom_invert = [1 / zoom for zoom in ref_img.header.get_zooms()[:3]]
+        img_rot = StatefulImage._get_rotation_matrix(simg.affine)
```
CHrlS98, line 106: Update the comment

src/scilpy/cli/scil_volume_math.py
arnaudbore: but still heavier than just a nib.load... why moving to load_img with simg

arnaudbore (changes requested):
One comment and few answers to previous comments.. we are almost there!

src/scilpy/cli/scil_tracking_pft.py
CHrlS98: I thought I was tracking with 0.040 mm backtracking distance (4 voxels for my 10 microns dataset) but it was actually 0.040 voxels, which is a bit ridiculous.

CHrlS98 (commented):
Here you go, starting to see the end of the road at last

src/scilpy/cli/scil_sh_reorient.py
CHrlS98: got it

src/scilpy/reconst/fiber_coherence.py
frheault: Prefer to keep this version, it mimics mrtrix's approach and DWI as input is way less annoying than requiring precomputed peaks as a prerequisite. 

Also more accurate (tests actual strides+bvecs combos vs the peaks approach which just rotates an existing approximation), it's slower, but worth it. Also removed compute_coherence_table_for_transforms, it was dead code now.

This avoids human error if people are using peaks from other sources or change/loose bvecs between operation. It fixes bvecs at the source.

src/scilpy/gradients/tests/test_bvec_bval_tools.py
frheault: Added caplog assertions, turns out verify_bval_spread had a real bug (>= instead of >), was warning even when shells were fine. Fixed.

src/scilpy/reconst/fiber_coherence.py
CHrlS98: parfait keep that way then

src/scilpy/cli/tests/test_tractogram_convert.py
frheault: Both already run without --no_bbox_check since 6834d317, that flag moved to #1352. You're looking at the old diff

src/scilpy/cli/scil_volume_resample.py
frheault: Updated, comment was stale

arnaudbore (approved):

CHrlS98 (approved):
ALRIGHT LET'S GO! 🚀

